PR #3, opened 2016-07-29
Refactor image uploading & downloading transcoder

typfel:
This PR extracts the image uploading/downloading from the `ZMClientMessageTranscoder`
into two new request strategies. The plan after that is to add a modified upstream
sync to the ZMClientMessageTrancoder for updating a message once a link preview
has been processed.

What is missing:
- ~~~Tests needs to be updated~~~
- ~~~Link preview uploading needs to be added to `ZMClientMessageTranscoder`~~~ - I would like to this as separate PR.


marcoconti83:
@typfel is this PR still valid? Is it outdated?

mikeger:
@typfel you're up to resurrect this zombie PR?

typfel:
yes I am

daehn:
Nice, Danger now comments inline ✅

typfel:
Very cool!

Source/Synchronization/Strategies/ImageDownloadRequestStrategy.swift
```diff
@@ -0,0 +1,121 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+public class ImageDownloadRequestStrategy : ZMObjectSyncStrategy, RequestStrategy {
+    
+    private let authenticationStatus : AuthenticationStatusProvider
+    private var downStreamSync : ZMDownstreamObjectSyncWithWhitelist!
```
mikeger: I think it should be `downstreamSync` as in the class name (`ZMDownstreamObjectSyncWithWhitelist`)

Source/Synchronization/Strategies/ImageDownloadRequestStrategy.swift
```diff
@@ -0,0 +1,121 @@
...
+            if message.imageMessageData != nil {
+                guard let assetId = message.assetId?.transportString() else { return nil }
+                return requestFactory.requestToGetAsset(assetId, inConversation: conversation.remoteIdentifier, isEncrypted: message.isEncrypted)
+            } else if (message.fileMessageData != nil) {
+                guard let assetId = message.fileMessageData?.thumbnailAssetID else { return nil }
+                return requestFactory.requestToGetAsset(assetId, inConversation: conversation.remoteIdentifier, isEncrypted: message.isEncrypted)
+            }
+        }
+        
+        return nil
```
mikeger: Would it crash in that case, since return is force-unwrap?

typfel: Ah don't know but I should change it be optional anyway.

Source/Synchronization/Strategies/ImageUploadRequestStrategy.swift
```diff
@@ -0,0 +1,211 @@
...
+    public func updateUpdatedObject(managedObject: ZMManagedObject, requestUserInfo: [NSObject : AnyObject]?, response: ZMTransportResponse, keysToParse: Set<NSObject>) -> Bool {
+        guard let message = managedObject as? ZMAssetClientMessage else { return false }
+        
+        update(message, withResponse: response, updatedKeys: keysToParse)
+        
+        var needsMoreRequests = false
+        
+        if keysToParse.contains(ZMAssetClientMessageUploadedStateKey) {
+            switch message.uploadState {
+            case .UploadingPlaceholder:
```
mikeger: I think we don't show the polka dots in UI any more, I guess we can drop sending it too (like web)? Just a suggestion.

typfel: Yep, but this PR is just porting existing functionality. I prefer not to change any behaviour.

mikeger:
Good PR! But also the more swift we put in, the more we have to migrate 3️⃣

Source/Synchronization/Strategies/ImageDownloadRequestStrategy.swift
typfel: Ok I can't change it to be optional since the `!` comes from the protocol but it also doesn't crash when you return `nil`.

*Or* we change the protocol so the return type is nullable.

mikeger: Let's not touch it for now and add nullability with swift 3 changes.

marcoconti83: returning `nil` in a `requestForFetchingObject` will cause an assert anyway. See `ZMTrapUnableToGenerateRequest`

Source/Synchronization/Strategies/ImageUploadRequestStrategy.swift
```diff
@@ -0,0 +1,211 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
```
marcoconti83: Do we need UIKit?

Source/Synchronization/Strategies/ImageDownloadRequestStrategy.swift
typfel: DownstreamSyncObjectSync doesn't seem to trap actually, it just deletes the object so maybe it makes sense to change the protocol when porting to swift 3.

Source/Synchronization/Strategies/ImageUploadRequestStrategy.swift
typfel: Nop :)

PR #7, opened 2016-08-02
Add address book upload analytics

daehn:
**What's in this PR?**

* Adds address book upload tracking ([6980](https://wearezeta.atlassian.net/browse/ZIOS-6980))

### DIFF
--- a/Source/AddressBookTracker.swift
+++ b/Source/AddressBookTracker.swift
@@ -1,0 +1,86 @@
+//
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+
+
+private enum Attribute: String {
+    case Outcome = "outcome"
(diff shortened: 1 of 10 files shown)


Source/Registration/ZMAddressBookEncoder.m
```diff
@@ -167,6 +168,7 @@ - (void)createPayloadWithCompletionHandler:(void(^)(ZMEncodedAddressBook *encode
                     index++;
                 }
                 result.otherData = (otherHashesCards.count < 1) ? nil : otherHashesCards;
+                result.addressBookSize = addressBook.numberOfContacts;
```
marcoconti83: We still have that artificial limit of 1000 contacts max, right? In this case, I think `numberOfContacts` will be capped at 1000. What we wanted to know is whether the users has 12.000 contacts, so we can figure out if we are too restrictive with a limit of 1000.

Source/AddressBookTracker.swift
```diff
@@ -0,0 +1,86 @@
...
+import Foundation
+
+
+private enum Attribute: String {
+    case Outcome = "outcome"
+    case Size = "size"
+    case Interval = "interval"
+}
+
+private extension Bool {
```
mikeger: I think that this is a bit weird. I would think of moving it into the scope of `AddressBookTracker`

Source/Registration/ZMAddressBookEncoder.m
daehn: @marcoconti83 `numberOfContacts` was newly added and returns the number of all contacts in the address book

Source/AddressBookTracker.swift
daehn: @mikeger unfortunately extensions can only by on file scope, but as it is only used once I will inline it

mikeger:
**All checks have failed**
Looks like Travis cannot find `ZMSystem` WAT

daehn:
@mikeger travis is not yet fully set up for `wire-iso-sync-engine`

mikeger:
@daehn All right, I see now

daehn:
@marcoconti83 please re-review

PR #15, opened 2016-08-08
Load selfUser for syncMOC on syncMOC queue in UserImageTranscoder

mikeger:
- Another approach to fixing the crashes happening in current release
- Originally self user was created from syncMOC on main thread
- Idea is that the data gets corrupted because of loading happening on the incorrect thread

### DIFF
--- a/Source/Synchronization/Transcoders/ZMUserImageTranscoder.m
+++ b/Source/Synchronization/Transcoders/ZMUserImageTranscoder.m
@@ -75,12 +75,8 @@ static NSString * const RequestUserProfileSmallAssetNotificationName = @"ZMReque
 {
     self = [super initWithManagedObjectContext:moc];
     if (self != nil) {
-        ZMUser *selfUser = [ZMUser selfUserInContext:moc];
-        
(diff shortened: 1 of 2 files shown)


Source/Synchronization/Transcoders/ZMUserImageTranscoder.m
```diff
@@ -101,7 +96,13 @@ - (instancetype)initWithManagedObjectContext:(NSManagedObjectContext *)moc image
                                                                                          entityName:ZMUser.entityName
                                                                       predicateForObjectsToDownload:filterForMediumImage
                                                                                managedObjectContext:self.managedObjectContext];
-        [self.mediumDownstreamSync whiteListObject:selfUser];
+        
+        [moc performGroupedBlock:^{
+            ZMUser *selfUser = [ZMUser selfUserInContext:moc];
```
marcoconti83: Nice fix, this is perfect to perform async as we don't need to whitelist RIGHT NOW. Every time we avoid doing a sync block it is a deadlock we spare ourselves 👍

PR #16, opened 2016-08-08
Remove logic to process plain text messages

daehn:
**What's in this PR?**

* This PR removes the logic to process plain text messages
* In the process a bug related to the `isTyping` state was found, if a user send a message the typing state was not reset for events of type `otr-message-add`.

‼️ This PR depends on a major update to `wire-ios-mocktransport`after https://github.com/wireapp/wire-ios-mocktransport/pull/3 has been merged.


marcoconti83:
Merged wireapp/wire-ios-mocktransport#3

PR #19, opened 2016-08-11
Wait longer for xcodebuild

daehn:


marcoconti83:
On our junkins machine, it might take longer than 30 minutes. Maybe we should bump it to 45?

daehn:
@marcoconti83 travis timed out even with 90 min 😁...

marcoconti83:
@daehn I think the issue might be somewhere else then. Probably the simulator is stuck on the host machine. One idea would be to look at the travis configuration for any other frameworks currently working fine with travis that has a test host app (some of them don't) and see if we spot any difference.

daehn:
`wire-ios-link-preview` also has a test host app, but the install and test commands in the `.travis.yml` are not really different there

PR #20, opened 2016-08-12
Remove `dispatch_async` in favour or `performGroupedBlock` where possible

marcoconti83:
Remove `dispatch_async` in favour or `performGroupedBlock` where possble

Where not possible, make sure to enter/leave groups.
Only using groups gives us the guarantee that tests will wait for the queue to be empty before moving to the next test. If we can't guarantee this, we will have a lot of flaky tests.

### DIFF
--- a/Source/Notifications/ZMNotifications+UserSession.m
+++ b/Source/Notifications/ZMNotifications+UserSession.m
@@ -42,9 +42,7 @@ static NSString *const ZMInitialSyncCompletedNotificationName = @"ZMInitialSyncC
 
 + (void)notifyInitialSyncCompleted
 {
-    dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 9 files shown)


Source/Notifications/ZMNotifications+UserSession.m
```diff
@@ -42,9 +42,7 @@ @implementation ZMUserSession (InternalNotifications)
 
 + (void)notifyInitialSyncCompleted
 {
-    dispatch_async(dispatch_get_main_queue(), ^{
-        [[NSNotificationCenter defaultCenter] postNotificationName:ZMInitialSyncCompletedNotificationName object:nil];
-    });
+    [[NSNotificationCenter defaultCenter] postNotificationName:ZMInitialSyncCompletedNotificationName object:nil];
```
Einh06: We are not switching to the main queue anymore here, are we sure if won't cause any bug? the fact that we were enforcing switching tells me there is something more to this.

PR #21, opened 2016-08-12
Use new address book upload endpoint

marcoconti83:
The new address book upload endpoint does not return suggestions of people you might know. However, we do not display people you might know anywhere in the client anymore, so we can safely set it to an empty set.
The code to handle people you might know is not being removed completely yet because we are still debating whether it will come back soon.

### DIFF
--- a/Source/Synchronization/ZMAddressBookSync.m
+++ b/Source/Synchronization/ZMAddressBookSync.m
@@ -31,7 +31,7 @@
 #import "zmessaging/zmessaging-Swift.h"
 
 static NSString * const ZMAddressBookTranscoderNeedsToBeUploadedKey = @"ZMAddressBookTranscoderNeedsToBeUploaded";
-static NSString * const ZMOnboardingEndpoint = @"/onboarding/v2";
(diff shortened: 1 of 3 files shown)


tiago-loureiro:
> The new address book upload endpoint does not return suggestions of people you might know. 

Just a little correction. We currently _do_ return that, I was mostly wondering if it is really necessary since no UI (except for Web) seems to use it.

marcoconti83:
@tiago-loureiro thanks for clarifying. Then let's say the effort to change to code to parse the new format from the new endpoint is not worth taking yet if we don't know if we will display this information anywhere. Once we have a final decision, we will either adapt it or remove the unused code.

marcoconti83:
As Zbigniew pointed out, this might still be used to display the "connect" button instead of the "invite" button in the list of contacts in the address book. So maybe after all we need to still parse it. Pending investigation.

marcoconti83:
This PR is now obsolete, I'm working on another branch on an overhaul of the address book upload. Will port those changes there, closing this PR.

PR #25, opened 2016-08-19
Fix missed call notification

Einh06:
**What's in this PR**
This PR fixes an issue where the user would not see notification when he/she missed a call.

This was due to the fact that we would update the last read timestamp of the conversation for event like `VoiceChannelActivate`, or any other event that does not generate a message or system message. This would lead to the clearing of notification, leaving us unable to update the call notification to a missed call. 
With this changed, the only kind of event that would update the conversation timestamp are successful call termination.

Additionally, we would wrongfully consider equivalent 2 call state events with the same ID and type, even though they might have a different sequence number. The sequence number is what identify in which order call state event happens, so 2 distinct events can't have the same sequence number. This is corrected by now also checking if the sequence numbers are the same.

### DIFF
--- a/Source/Synchronization/Transcoders/ZMConversationTranscoder.m
+++ b/Source/Synchronization/Transcoders/ZMConversationTranscoder.m
@@ -342,10 +342,10 @@ static NSString *const ConversationInfoArchivedValueKey = @"archived";
     BOOL eventIsCompletedVoiceCall = NO;
     if (event.type == ZMUpdateEventConversationVoiceChannelDeactivate) {
         NSString *reason = [[event.payload optionalDictionaryForKey:@"data"] optionalStringForKey:@"reason"];
-        eventIsCompletedVoiceCall = ! [reason isEqualToString:@"missed"];
+        eventIsCompletedVoiceCall = ![reason isEqualToString:@"missed"];
     }
     
-    if ((! [ZMMessage doesEventTypeGenerateMessage:event.type]) || eventIsCompletedVoiceCall) {
+    if (eventIsCompletedVoiceCall) {
         [self updateLastReadForInvisibleEventInConversation:conversation
                                                   timeStamp:timeStamp
                                            oldLastTimeStamp:oldLastTimeStamp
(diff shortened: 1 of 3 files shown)


marcoconti83:
In the PR description, I am a bit confused by:

> This was due to the fact that we would update the last read timestamp of the conversation for event like call started, or any other event that does not. The only kind of event that would update the conversation timestamp are successful call termination.

"...that does not" what?. Also, where we updating it or not (1st sentence)? From the PR it seems that now we do, we did not before.

Source/Synchronization/Transcoders/ZMConversationTranscoder.m
```diff
@@ -342,10 +342,10 @@ - (void)updatePropertiesOfConversation:(ZMConversation *)conversation fromEvent:
     BOOL eventIsCompletedVoiceCall = NO;
     if (event.type == ZMUpdateEventConversationVoiceChannelDeactivate) {
         NSString *reason = [[event.payload optionalDictionaryForKey:@"data"] optionalStringForKey:@"reason"];
-        eventIsCompletedVoiceCall = ! [reason isEqualToString:@"missed"];
+        eventIsCompletedVoiceCall = ![reason isEqualToString:@"missed"];
     }
     
-    if ((! [ZMMessage doesEventTypeGenerateMessage:event.type]) || eventIsCompletedVoiceCall) {
+    if (eventIsCompletedVoiceCall) {
```
marcoconti83: So now we only update last read only if we get a completed voice call event. What other kind of events are we excluding by removing `! [ZMMessage doesEventTypeGenerateMessage:event.type])`

Einh06: By removing we exclude all message that are not messages(text, asset, knock etc.) or system messages (member leave / join).
That would include for instance `VoiceChannelActivate`, `Typing`.

marcoconti83: Thanks for clarifying!

PR #28, opened 2016-08-24
Port event decoder branch to new repository

daehn:
#### This PR ports the branch `feature/add-event-decoder-ZIOS-5318` from @basine to the new repository, some small adjustments have been made.

___

* Adds a separate database on the sync-engine which is used to store received update events before they have been decrypted.
* The `EventDecoder` is responsible for decrypting and storing the update events and is used by it's public method `processEvents(_:block:)` which takes an array of decrypted `ZMUpdateEvent` and calls the passed in block with the decrypted ones.
* The update events get stored as `StoreUpdateEvents` in the database.
* The separate database acts as an on-disk cache and ensures the decrypted events are not lost in case of a crash before processing them (as they can only be decrypted once).
* The mentioned adjustments include making sure the database uses an in memory persistent store when testing, inserting `performGroupedBlock` calls for the new `eventMOC` and adjusting failing and flaky tests.

___

🚫 This PR depends on: 
- [x]  A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/16 has been merged.
- [x] A new release of  `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/21 has been merged.


daehn:
This has to be a major release as the public API has been changed to inject the shared container url.

Source/Notifications/Push notifications/ZMLocalNotificationDispatcher.m
```diff
@@ -150,18 +150,54 @@ - (void)cancelNotificationForConversation:(ZMConversation *)conversation
...
+        }
+        // if the event is not a message, we want to process it in any case
+        if (event.messageNonce == nil) {
+            return YES;
+        }
+        // if the event is a message, we only want to process it if it does not have a preexisting message
+        // TODO Sabine: Can this lead to multipart messages not being shown at all?
+        // e.g. when we receive an asset message, we receive a preview and medium image message
+        // However, we only process the preview. If under bad network, the medium image part arrives before the preview,
+        // we would not display any notification at all, would we?
```
Einh06: is this comment still valid?

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -0,0 +1,225 @@
...
+            
+            strongSelf.eventMOC.performGroupedBlockAndWait {
+                // decryptEvents and insert counting upwards from highest index in DB
+                for (idx, event) in newEvents.enumerate() {
+                    if let storedEvent = StoredUpdateEvent.create(event, managedObjectContext: strongSelf.eventMOC, index: idx + lastIndex + 1) {
+                        newStoredEvents.append(storedEvent)
+                    }
+                }
+                
+                strongSelf.eventMOC.saveOrRollback()
```
Einh06: Are we sure this wont cause retain cycle? (strongSelf capturing itself triggers a red flag in my mind)

marcoconti83: `performGroupedBlockAndWait` will not cause retain cycle because it's sync (other than if there is a deadlock and the block is never executed, but then the deadlock is the biggest problem).
But `...andWait` are dangerous because they can easily generate deadlocks, and as a general rule we do not use them but for a few very specific exceptions.

Source/Notifications/Push notifications/ZMLocalNotificationDispatcher.m
basine: Yes :) But I guess we ignored the fact earlier as well. We could add an extra check for asset messages to see if the prefetchedMessage has the preview.

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -0,0 +1,225 @@
...
+//
+
+import Foundation
+import Cryptobox
+import ZMCDataModel
+
+private let zmLog = ZMSLog(tag: "EventDecoder")
+
+
+@objc public class EventDataController: NSObject {
```
marcoconti83: What does this class do? `Controller` is a bit of a generic term. Can you please add a comment to the class with a list of its responsibilities? Bonus cookie points for finding a way to remove `Controller` from the name

Source/Synchronization/Decoding/EventDecoder.swift
marcoconti83: Is there a way to remove the `andWait`?

Source/Synchronization/Decoding/EventDecoder.swift
daehn: I agree, It sets up the event core data stack, so maybe `EventCoreDataStack` or something like this would make more sense?

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -0,0 +1,225 @@
...
+        if !fileManager.fileExistsAtPath(_storeURL.path!) {
+            do {
+                try fileManager.createDirectoryAtURL(_storeURL, withIntermediateDirectories: true, attributes: nil)
+            } catch {
+                assertionFailure("Failed to get or create directory \(error)")
+            }
+        }
+        
+        do {
+            try _storeURL.setResourceValue(1, forKey:NSURLIsExcludedFromBackupKey)
```
marcoconti83: We should also set file protection level to protected until unlocked, as we do for the other resources

Source/Synchronization/Decoding/EventDecoder.swift
Einh06: We could go as far as to ask wether or not it make sense to have a class altogether. It's just a `NSManagedObjectContext` that has convenience initialization, it does not provide additional functionality.

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -0,0 +1,225 @@
...
+    private func consumeStoredEvents(someStoredEvents: [StoredUpdateEvent], someEvents: [ZMUpdateEvent], lastIndexToFetch: Int64, consumeBlock: ConsumeBlock) {
+        var storedEvents = someStoredEvents
+        var events = someEvents
+        
+        var hasMoreEvents = true
+        while hasMoreEvents {
+            // if the index of the last fetched object is not the lastIndexToFetch, we have more objects to fetch
+            hasMoreEvents = storedEvents.last?.sortIndex != lastIndexToFetch
+            
+            if events.count > 0 {
```
marcoconti83: I can't figure out the logic of this method (specifically, the while loop and the `hasMoreEvents`), could you please clarify how it works? I think it would be good to explain that in code comments so our future self in 6 months won't have to ask the same question :)

Source/Synchronization/Strategies/UserClientRequestFactory.swift
```diff
@@ -30,12 +30,12 @@ enum UserClientRequestError: ErrorType {
 
 public class UserClientRequestFactory {
     
-    public init(keysCount: UInt = 100, missingClientsUserPageSize pageSize: Int = 128) {
+    public init(keysCount: UInt16 = 100, missingClientsUserPageSize pageSize: Int = 128) {
```
marcoconti83: Why the change to UInt16?

basine: Because cryptobox generateMoreKeys expects a UInt16

Source/Synchronization/Decoding/EventDecoder.swift
daehn: We need to make sure that we only process the stored events after they have been created, if we do this async this might not be the case. Before we did not switch to the event MOC queue to create the events.

daehn: We can try wrapping this and the processing part in a `performGroupedBlock`

Source/Synchronization/Strategies/UserClientRequestFactory.swift
marcoconti83: Good point :) We can generate at maximum 65k keys, hence the UInt16 restriction.

Source/Synchronization/Decoding/EventDecoder.swift
marcoconti83: @Einh06 has a point

daehn: We already have a lot of convenience functions on `NSManagedObjectContext` to create contexts etc, I agree that we should keep this consistent, but if we do move this in an extension then we need to make sure that the naming and comments make the difference between this core data stack and the main data base clear.

Source/Synchronization/Decoding/EventDecoder.swift
Einh06: The fact that there is no retain cycle is not self evident. This tells me we should still make it clear to the reader, to keep the reasoning of what the code is doing local, and not have to jump somewhere else.

marcoconti83: We will add to the documentation of `performGroupedBlockAndWait`, as part of this method contract, that the block will not be stored internally, just passed to the queue and retained only until the queue executes it, then discarded.

Source/Synchronization/Decoding/EventDecoder.swift
daehn: I checked how we do it for the other resources and core data already uses `NSFileProtectionCompleteUntilFirstUserAuthentication` as the default

marcoconti83:
New version of data-model is out: 13.1.0

daehn:
@marcoconti83 @Einh06 please re-review

daehn:
Does not build with the new data-model version `13.1.0` because there have been some changes to the delivery state field

marcoconti83:
This PR does not build properly because of data-model, but @basine upcoming PR will fix it.

PR #29, opened 2016-08-25
New address book upload specs

marcoconti83:
# Reason for this PR
"Uploading" the address book is now in fact a search. The backend does not store address book, so if I upload only once, I will never get any match for users that join in the future. For this reason, I need to constantly search using the contacts I have in the address book.
Previously, we were only uploading if the address book changed. We were also only considering the first 1000 contacts.

# Changes
- Ported most of the code to Swift
- Removed upload of "self" card. This is not required anymore from the backend
- Do not check if the address book changed since last upload. Now we need to upload even if the list of contacts did not change.
- Upload all contacts, even if they are more than 1000. They will be uploaded in batches.
- Accept phone numbers that start with +0 (this is needed for test automation by QA)
- (EDIT) unified a few mocks used in tests

# Known issues
Still TODO (will be in a future PR):
   - Tracking
   - Parsing suggestions from address book upload


Source/Data Model/ZMUser+Contact.swift
```diff
@@ -1,26 +1,26 @@
...
-@protocol ZMSingleRequestTranscoder;
-
-@interface ZMEmptyAddressBookSync : ZMAddressBookSync 
-@end
+extension ZMUser {
+    
+    func matchingContact(userSession: ZMUserSession) -> ZMAddressBookContact? {
+        return AddressBookSearch().contactForUser(self)
+    }
+}
```
typfel: missing new line

Source/Registration/AddressBook.swift
```diff
@@ -0,0 +1,298 @@
...
+    /// Closure used to generate iterator. Used in testing
+    private let allPeopleClosure : AllPeopleClosure
+    typealias AllPeopleClosure = (ref: ABAddressBook) -> (AnyGenerator<ABRecordRef>)
+    
+    /// Closure to get number of people. Used in testing
+    private let numberOfPeopleClosure : NumberOfPeopleClosure
+    typealias NumberOfPeopleClosure = (ref: ABAddressBook) -> (Int)
+
+    /// Address book. Will fail if it has no authorization to access AB
+    /// - parameter generatingClosure: custom function to return an iterator (used for testing)
```
daehn: I looks like `generatingClosure` has been renamed to `allPeopleClosure`

Source/Registration/AddressBook.swift
```diff
@@ -0,0 +1,298 @@
...
+}
+
+// MARK: - Default behaviour (test injection)
+
+extension AddressBook {
+    
+    /// Uses the passed in closure, or the standard method it the closure is nil, to
+    /// check if the AB access was granted. Returns whether it was granted.
+    static private func checkAccessToAB(checkClosure: (()->(Bool))?) -> Bool {
+        if let checkClosure = checkClosure { // for some reason, using ?? won't compile here
```
daehn: `return checkClosure?() ?? AddressBook.userHasAuthorizedAccess` should work, but might also lose some readability then

Source/UserSession/Search/ZMSearchResult+addressBook.swift
```diff
@@ -0,0 +1,57 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// This is used for testing only
+var debug_searchResultAddressBookOverride : AddressBookAccessor? = nil
+
+extension ZMSearchResult {
```
typfel: To be consistent i would CamelCase the file name to `ZMSearchResult+AddressBook.swift`

Source/Registration/AddressBook+encoding.swift
```diff
@@ -0,0 +1,118 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import ZMUtilities
+
+/// Mark: - Encoding
+extension AddressBook {
```
typfel: To be consistent i would CamelCase the file name to `AddressBook+Encoding.swift`

Source/UserSession/ZMUserSession+AddressBook.swift
```diff
@@ -0,0 +1,57 @@
...
+    public static func removeAddressBookUploadObserverToken(token: NSObjectProtocol) {
+        NSNotificationCenter.defaultCenter().removeObserver(token)
+    }
+}
+
+// MARK: - Address book upload
+extension ZMUserSession {
+    
+    /// Asynchronously uploads the next chunk of the address book
+    public func uploadAddressBook() {
```
typfel: Have you checked how often we call this in UI? Maybe were relying on the fact that AB was only uploaded when changed.

Source/Registration/AddressBook.swift
```diff
@@ -0,0 +1,298 @@
...
+}
+
+extension String {
+    
+    /// Returns a normalized phone number or nil
+    var validatedPhoneNumber : String? {
+        
+        // allow +0 numbers
+        if self.hasPrefix("+0") {
+            return "+" + (self
```
julesmummdry: the part I am happy about 👍🏼🍉

marcoconti83:
@daehn @typfel  fixed

PR #32, opened 2016-08-25
Fix searching with address book when some contacts do not have valid …

marcoconti83:
Uploading the address book in batches would get stuck trying to upload the last batch

We get the information on how many AB contacts are available directly from the AB (because we don't want to parse/normalize ALL of them just to know how many there are, it might take a long time). However some of these contacts might actually have invalid email or phone number, so the actually number of contacts that we can upload might be lower. This was causing the last batch to be empty, but we thought it was not the last batch, so we kept trying to upload the next (also empty).

# Changes
Before this PR, we were relying on the number of contacts in the AB, and not the number of "uploadable" contacts, to know if we reached the last batch of the search and we should restart from the first. But in case of non valid contacts, we would think that we never reach the end.
Now we correctly check if we reached the last "upload able" contact by checking if we have less contacts in the batch that we could have, and if that's the case, reset the batch counter so that we reupload the first one the next time.

Also, when trying to upload a new batch, it could be that the computed batch is empty - because the previous batch stopped exactly at the last "uploadable" contact. If this is the case, and we are trying to upload this (empty) batch, immediately reupload the first one instead.

### DIFF
--- a/Source/Synchronization/Strategies/AddressBookUploadRequestStrategy.swift
+++ b/Source/Synchronization/Strategies/AddressBookUploadRequestStrategy.swift
@@ -183,7 +183,7 @@ extension AddressBookUploadRequestStrategy : RequestStrategy, ZMSingleRequestTra
 
 extension EncodedAddressBookChunk {
     
-    /// Whether another address book chunk needs to be uploaded after this one
+    /// Whether this is the last chunk and no following chunk needs to be uploaded after this one
     var isLast : Bool {
         return UInt(self.otherContactsHashes.count) < maxEntriesInAddressBookChunk
     }
--- a/Source/Synchronization/Strategies/AddressBookUploadRequestStrategy.swift
+++ b/Source/Synchronization/Strategies/AddressBookUploadRequestStrategy.swift
@@ -49,7 +49,7 @@ private let addressBookLastUploadedIndex = "ZMAddressBookTranscoderLastIndexUplo
     private var requestSync : ZMSingleRequestSync!
     
     /// Encoded address book chunk
-    private var encodedAddressBookChunk : EncodedAddressBookChunk? = nil
+    private var encodedAddressBookChunkToUpload : EncodedAddressBookChunk? = nil
     
     /// Is the payload being generated? This is an async operation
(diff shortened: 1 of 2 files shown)


Tests/Source/Synchronization/Strategies/AddressBookUploadRequestStrategyTest.swift
```diff
@@ -288,6 +290,51 @@ extension AddressBookUploadRequestStrategyTest {
...
+        let cardsNumber = Int(Double(maxEntriesInAddressBookChunk) * 1.5)
+        self.addressBook.fillWithContacts(UInt(cardsNumber))
+        self.addressBook.numberOfContactsOverride = UInt(maxEntriesInAddressBookChunk * 5)
+        
+        // when
+        let request1 = self.getNextUploadingRequest()
+        let request2 = self.getNextUploadingRequest()
+        let request3 = self.getNextUploadingRequest()
+        
+        // then
```
typfel: duplicate `then`

marcoconti83:
The TODO comment that the bot reported was fixed on the other PR: https://github.com/wireapp/wire-ios-sync-engine/pull/30

Source/Synchronization/Strategies/AddressBookUploadRequestStrategy.swift
```diff
@@ -148,16 +148,55 @@ extension AddressBookUploadRequestStrategy : RequestStrategy, ZMSingleRequestTra
...
+    private func startUpload(encodedChunk: EncodedAddressBookChunk) {
+        self.encodedAddressBookChunkToUpload = encodedChunk
+        self.requestSync.readyForNextRequest()
+        ZMOperationLoop.notifyNewRequestsAvailable(self)
+    }
+}
+
+extension EncodedAddressBookChunk {
+    
+    /// Whether another address book chunk needs to be uploaded after this one
```
daehn: The comment seems misleading to me, I assume no other address book chunk needs to be uploaded if this returns `true`

marcoconti83: Fixed!

PR #39, opened 2016-09-01
Enable sending confirmation messages in background

basine:
**What's in this PR**
Commit 1: Dissect UserClientRequestStrategy and created MissingClientRequestStrategy which only cares about missing clients
Commit 2: Add `BackgroundAPNSConfirmationStatus` that keeps track of sending status of confirmation messages, starts a background activity for each message to be sent and ends the activity when the request finishes
Commit 3 & 4: Force requests to VoIP session to make sure requests are sent out quickly

**Needs**
- [x] Release of wire-ios-data-model with https://github.com/wireapp/wire-ios-data-model/pull/28


marcoconti83:
Conflict!

Source/Synchronization/Strategies/MissingClientsRequestFactory.swift
```diff
@@ -0,0 +1,53 @@
+//
+//  MissingClientsRequestFactory.swift
+//  zmessaging-cocoa
+//
```
daehn: Missing copyright header

marcoconti83:
This PR moves some code from one transcoder to two separate ones. As such, there are a lot of code changes (1600+ lines). Can you please highlight with PR comments which parts should we focus on reviewing, and which parts are just copy-paste?

Source/UserSession/BackgroundAPNSConfirmationStatus.swift
```diff
@@ -0,0 +1,86 @@
...
+        tornDown = true
+    }
+    
+    deinit {
+        assert(tornDown, "Needs to tear down BackgroundAPNSConfirmationStatus")
+    }
+    
+    // Called after a confirmation message has been created from an event received via APNS
+    public func needsToConfirmMessage(messageNonce: NSUUID) {
+        let backgroundTask = backgroundActivityFactory.backgroundActivity(withName: messageNonce.transportString()) { [weak self] in
```
marcoconti83: Instead of using `messageNonce.transportString` as the label, I would add some text, e.g. `Waiting for message \(messageNonce.transportString)`. This could be more useful when debugging activities, the nonce is quite cryptic as a name.

Source/UserSession/BackgroundAPNSConfirmationStatus.swift
```diff
@@ -0,0 +1,86 @@
...
+            guard let strongSelf = self else { return }
+            // The message failed to send in time. We won't continue trying.
+            strongSelf.managedObjectContext.performGroupedBlock{
```
marcoconti83: How do we know that the message failed to send? It it because the activity times out?

PR #41, opened 2016-09-02
Push notifications for reactions

basine:
**What's in this PR**

We are now creating push notifications for reactions for messages the selfUser sent.
We are also cancelling push notifications for "unlikes" from the same user.

Note: Creating pushes from message updateEvents needs refactoring. We should create notifications from messages instead of events to avoid parsing the event payload twice or more times.


Source/Notifications/Push notifications/Notification Types/ZMLocalNotificationForMessageEvent.swift
```diff
@@ -216,3 +214,61 @@ public class ZMLocalNotificationForKnockMessage : ZMLocalNotificationForPostInCo
...
+        {
+            cancelNotifications()
+            shouldBeDiscarded = true
+            return nil
+        }
+        return nil
+    }
+    
+    override func configureAlertBody() -> String {
+        // TODO: use correct string
```
marcoconti83: Todo?

marcoconti83:
Looks good!

marcoconti83:
About the comment ("we should refactor..."), please create a ticket to track it.

Source/Notifications/Push notifications/Notification Types/ZMLocalNotificationForMessageEvent.swift
basine: Right, that's Florian's comment. Will check if there is anything to do

PR #43, opened 2016-09-02
handle unarchive with message events separately

basine:
**In this PR**

Just one tiny change: We are currently unarchiving conversations by checking the event type. We should not do this for message events since they might be invisible.

Depends on:
- [x] release with: https://github.com/wireapp/wire-ios-data-model/pull/34


marcoconti83:
Data model merged

PR #45, opened 2016-09-06
Do not send confirmation messages

basine:
**In this PR**

We will not send confirmation messages for now.
Note: Since I couldn't run tests, I am not 100% sure I removed all related tests.

### DIFF
--- a/Source/Synchronization/Transcoders/ZMClientMessageTranscoder.m
+++ b/Source/Synchronization/Transcoders/ZMClientMessageTranscoder.m
@@ -33,7 +33,8 @@
 #import <zmessaging/zmessaging-Swift.h>
 #import "ZMOperationLoop.h"
 
-
+// TODO Confirmation enable
+static const BOOL SendDeliveryReceipts = NO;
 
 @interface ZMAssetClientMessage (ImagePredicates)
 
@@ -327,15 +328,16 @@
         case ZMUpdateEventConversationOtrMessageAdd:
         case ZMUpdateEventConversationOtrAssetAdd:
             updateResult = [ZMOTRMessage messageUpdateResultFromUpdateEvent:event
-                                                  inManagedObjectContext:self.managedObjectContext
-                                                          prefetchResult:prefetchResult];
-            // TODO Confirmation: comment back in 
-//            if (updateResult.needsConfirmation) {
-//                ZMClientMessage *confirmation = [updateResult.message confirmReception];
-//                if (event.source == ZMUpdateEventSourcePushNotification) {
-//                    [self.apnsConfirmationStatus needsToConfirmMessage:confirmation.nonce];
-//                }
-//            }
+                                                     inManagedObjectContext:self.managedObjectContext
+                                                             prefetchResult:prefetchResult];
+            if (SendDeliveryReceipts) {
+                if (updateResult.needsConfirmation) {
+                    ZMClientMessage *confirmation = [updateResult.message confirmReception];
+                    if (event.source == ZMUpdateEventSourcePushNotification) {
+                        [self.apnsConfirmationStatus needsToConfirmMessage:confirmation.nonce];
+                    }
+                }
+            }
             break;
         default:
             return nil;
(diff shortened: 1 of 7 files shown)


Source/Synchronization/Transcoders/ZMClientMessageTranscoder.m
```diff
@@ -329,12 +329,13 @@ - (ZMMessage *)messageFromUpdateEvent:(ZMUpdateEvent *)event
             updateResult = [ZMOTRMessage messageUpdateResultFromUpdateEvent:event
                                                   inManagedObjectContext:self.managedObjectContext
                                                           prefetchResult:prefetchResult];
-            if (updateResult.needsConfirmation) {
-                ZMClientMessage *confirmation = [updateResult.message confirmReception];
-                if (event.source == ZMUpdateEventSourcePushNotification) {
-                    [self.apnsConfirmationStatus needsToConfirmMessage:confirmation.nonce];
-                }
-            }
+            // TODO Confirmation: comment back in
```
marcoconti83: Instead of commenting out the code, with the risk that it will become not compilable in the future without us noticing, we should have a `if(false) { ... }` so we are sure to update it if the API changes.

Tests/Source/Integration/APNSTests.m
```diff
@@ -490,62 +490,64 @@ - (void)testThatItFetchesTheNotificationAndPingsBackToTheBackendWhenReceivingAVo
...
-    };
-    
-    // when
-    [self.userSession receivedPushNotificationWithPayload:[self APNSPayloadForNotificationPayload:payload] completionHandler:nil source:ZMPushNotficationTypeVoIP];
-    WaitForAllGroupsToBeEmpty(0.2);
-    
-    
-    XCTAssert([self waitForCustomExpectationsWithTimeout:0.5]);
-}
+// TODO Confirmation : Comment back in
```
marcoconti83: See comment above.
It would be even better to have a constant `SendDeliveryReceipts` and add a `if(!SendDeliveryReceipts) { return }` in all tests.

Tests/Source/Integration/ConversationTests+Confirmation.swift
```diff
@@ -18,118 +18,119 @@
...
-        if convObserver.notifications.count > 0 {
-            return XCTFail()
-        }
-        guard let messageChangeInfo = messageObserver.notifications.firstObject  as? MessageChangeInfo else {
-            return XCTFail()
-        }
-        XCTAssertTrue(messageChangeInfo.deliveryStateChanged)
-    }
-}
+// TODO Confirmation: comment back in
```
marcoconti83: Same as above

Tests/Source/Synchronization/Transcoders/ZMClientMessageTranscoderTests.m
```diff
@@ -1373,153 +1373,154 @@ - (void)testThatThePreviewGenericMessageDataHasTheOriginalSizeOfTheMediumGeneric
...
-    ZMTransportRequest *request = [self.sut.upstreamObjectSync nextRequest];
-    [request completeWithResponse:[ZMTransportResponse responseWithPayload:@{} HTTPstatus:200 transportSessionError:nil]];
-    WaitForAllGroupsToBeEmpty(0.5);
-    
-    // then
-    XCTAssertTrue(confirmationMessage.isZombieObject);
-}
-
-@end
+// TODO Confirmation: comment back in
```
marcoconti83: Same as above

marcoconti83:
Thanks for addressing this @basine! 

One more improvement: if we can share one single constant between the code and the tests, then we will make sure that the tests are re-enabled as soon as the code is.

E.g. adding `external const BOOL SendDeliveryReceipts;` to the header, so we can use it in tests too. 

Then instead of using an `#if #endif` in the tests, we can just use 
`if(SendDeliveryReceipts) { return }`
in each test.
 
This also has the advantage of compiling all test code. With the `#if`, it is actually not even trying to compile it, so any API error won't be detected, in the same fashion as if it was commented out.

basine:
Yeah, but then I have to add this to every single test?!

marcoconti83:
As far as I know there's no other way :(

PR #49, opened 2016-09-06
Fix restart address book search from first batch

marcoconti83:
Address book search was stopping when it reaches the end of the AB. It should have restarted from the beginning.

# Changes
Return empty chunk when we tried to generate a chunk past the 0 starting position and there were no more contacts. This will be interpreted as "I reached the end" and trigger the restart.


marcoconti83:
There is still something fishy going on here. If I have an AB big enough to require 3 batches, this is the sequence of batches that will be uploaded when I test on simulator:
`Batch 1 -> Batch 2 -> Batch 3 
Batch 2 -> Batch 1 -> Batch 3`
At the end of the day, it's still uploading all three batches but this warrants more investigation.

PR #53, opened 2016-09-09
Fix multithreading violation in EventDecoder

daehn:
# What's in this PR?

There was a Core Data multithreading violation introduced in https://github.com/wireapp/wire-ios-sync-engine/pull/28 in which we fetched `StoredEvents` while still on the `syncMOC`'s queue in `consumeStoredEvents:someEvents:lastIndexToFetch:consumeBlock:`.

* This PR changes the event processing to be asynchronous and dispatches all processing onto the `eventMOC`'s queue. The `consumeBlock` is called on the `syncMOC` with the decrypted update events.
* `processBatch:storedEvents:block:` now takes an additional optional completion closure which is used in the `consumeStoredEvents` method to ensure the previously processed events have been deleted when the next batch is fetched.

### DIFF
--- a/Source/Synchronization/Decoding/EventDecoder.swift
+++ b/Source/Synchronization/Decoding/EventDecoder.swift
@@ -119,6 +119,8 @@ extension NSManagedObjectContext {
     let syncMOC: NSManagedObjectContext
     weak var encryptionContext : EncryptionContext?
     
+    private typealias EventsWithStoredEvents = (storedEvents: [StoredUpdateEvent], updateEvents: [ZMUpdateEvent])
+    
     public init(eventMOC: NSManagedObjectContext, syncMOC: NSManagedObjectContext) {
         self.eventMOC = eventMOC
         self.syncMOC = syncMOC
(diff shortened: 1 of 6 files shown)


Source/UserSession/BackgroundAPNSConfirmationStatus.swift
```diff
@@ -28,9 +28,9 @@ extension UIApplication : ApplicationStateOwner {}
 
 @objc
 public class BackgroundAPNSConfirmationStatus : NSObject {
-    
-    /// Switch for sending delivery receipts
-    public static let sendDeliveryReceipts : Bool = true
+
+    /// TODO: Switch sending delivery receipts back on
```
marcoconti83: Let's change this in a separate PR, it's a different topic

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -131,11 +133,11 @@ extension NSManagedObjectContext {
     /// If the app crashes while processing the events, they can be recovered from the database
     /// Recovered events are processed before the passed in events to reflect event history
     public func processEvents(events: [ZMUpdateEvent], block: ConsumeBlock) {
-
+        
         // fetch first batch of old events
         let batchSize = EventDecoder.BatchSize
         let oldStoredEvents = StoredUpdateEvent.nextEvents(eventMOC, batchSize: batchSize, stopAtIndex: nil)
```
marcoconti83: Another threading violation

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -168,56 +170,72 @@ extension NSManagedObjectContext {
...
-        // process old events
-        consumeStoredEvents(oldStoredEvents, someEvents: oldEvents, lastIndexToFetch: lastIndex, consumeBlock: block)
-        
-        // process new events
-        if newEvents.count > 0 {
-            processBatch(newEvents, storedEvents: newStoredEvents, block: block)
-        } else {
-            block([])
+        process(
+            oldEvents: (storedEvents: oldStoredEvents, updateEvents: oldUpdateEvents),
```
marcoconti83: As discussed, we don't need to pass around old events and new events. Once the new events have been inserted in CD, we just call a function that will fetch all CD entries (in batches) and process them, it does not matter which ones are new and which ones are old.
As an additional optimization, when we fetch a batch, we should make sure to enable the flag in the fetch request that will materialize all results immediately instead of returning faults. This will save us a lot of round trips to the disk for old events.

daehn:
@marcoconti83 please re-review.

marcoconti83:
There are still a lot of perform inside perform inside perform inside perform

images

Not sure if they are all needed, but it looks ok. I'd like at least another person to review this.

PR #54, opened 2016-09-09
Refactor notification messages

basine:
**In this PR**

I cleaned up the notification creation by
(1) For events that create messages: Process the messages instead of the raw events to avoid double parsing of the event payload
(2) For call events: Move the responsibility of call state tracking (who initiated the call, what kind of call is it, did the selfUser join etc. into a separate object)
(3) For all events that don't create a message: Fetch and pass in the conversation before creating the event to avoid multiple fetches when copying events.

It's probably best reviewed by opening the project since files have been moved / deleted etc.


marcoconti83:
Closed in favor of https://github.com/wireapp/wire-ios-sync-engine/pull/60 because of crazy merge conflicts (I had to rebase)

PR #61, opened 2016-09-12
Feature/failing integration tests

typfel:
## What was failing
Since the event decoder now operates asynchronously when processing events, the call to `consumeUpdateEvents:` is no longer blocking. This has the side affect that the calls to `enqueueDelayedSave` in `processUpdateEvents:` are executed before the any update events has been processed and there by causing the sync context not be saved in some cases.

## The fix
Move `enqueueDelayedSave` calls into the event coder's processing block.

### DIFF
--- a/Source/Synchronization/ZMSyncStrategy.m
+++ b/Source/Synchronization/ZMSyncStrategy.m
@@ -622,6 +622,7 @@ ZM_EMPTY_ASSERTING_INIT()
         }
         [self.localNotificationDispatcher processEvents:decryptedEvents liveEvents:YES prefetchResult:nil];
         [self.syncMOC saveIfTooManyChanges];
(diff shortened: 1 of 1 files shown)


Source/Synchronization/ZMSyncStrategy.m
```diff
@@ -622,6 +620,7 @@ - (void)consumeUpdateEvents:(NSArray<ZMUpdateEvent *>*)events
         }
         [self.localNotificationDispatcher processEvents:decryptedEvents liveEvents:YES prefetchResult:nil];
         [self.syncMOC saveIfTooManyChanges];
```
daehn: The `saveIfTooManyChanges` call can go, `enqueueDelayedSave` already checks if there a too many changes and saves right away in that case.

PR #62, opened 2016-09-13
Add test to verify each event is only processed once in `EventDecoder`

daehn:
# What's in this PR?

* Adds a test to verify each event is only processed once in `EventDecoder`
* This test is currently **failing intentionally** and was added to track the progress on this issue.
____

#### Explanation of the underlying Issue ([7274](https://wearezeta.atlassian.net/browse/ZIOS-7274)): 

The `EventDecoder` API was converted to be asynchronous to fix the multithreading violations and avoid too many -performGroupedBlockAndWait: calls in #53, this current implementation however has the limitation that if called successively, the same events can be processed multiple times.

_Consider the following example:_

While processing the notification stream the event decoder creates lots of `StoreUpdateEvents` and stores them in the event database (accessed through the `eventMOC`), as clients need to be added on the `syncMOC` and the actual forwarding to the transcoders is done on the `syncMOC` a couple of context switches need to be done. Each contexts queue (`syncMOC` and `eventMOC`) is a serial queue, but as we are switching between them other block can be enqueued.

Forwarding the events to all the transcoders can take a bit, and the events are only deleted from the `eventMOC` after the processing has been done, this means that in case of an incoming push notification it can happen that all events are processed again as they have not yet been deleted.

_Possible Solutions:_
* We wanted to avoid lot's of nested blocking calls to avoid potential deadlocks but one solution would be to make the method synchronous by using `..andWait:` calls...
* Another way would be to have an in memory counter that keeps track of the last fetched `StoredUpdateEvent`s `sortIndex`, subsequent calls would only fetch events after that index, in case of a crash the index would be reset.

### DIFF
--- a/Tests/Source/Synchronization/EventDecoderTests.swift
+++ b/Tests/Source/Synchronization/EventDecoderTests.swift
@@ -123,5 +123,36 @@ class EventDecoderTest: MessagingTest {
         // then
         XCTAssertEqual(callCount, 2)
     }
+    
+    func testThatItDoesNotProcessTheSameEventsTwiceWhenCalledSuccessively() {
+        // given
+        EventDecoder.testingBatchSize = 2
+        
+        let event1 = dummyEvent()
+        let event2 = dummyEvent()
+        let event3 = dummyEvent()
+        let event4 = dummyEvent()
+        
+        let _ = StoredUpdateEvent.create(event1, managedObjectContext: eventMOC, index: 0)
+        eventMOC.saveOrRollback()
 
+        sut.processEvents([event2]) { (events) in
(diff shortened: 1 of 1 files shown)


marcoconti83:
The first proposed solution (`...andWait:`) is dangerous. 
However, let's consider this: if the sync thread is the only user of the event storage, then we need to switch on the event decoder thread only because Core Data would consider it a violation, not because we actually want to do any async operation or there is any threading issue. 
Therefore I think using `andWait` could be OK: logically it would be as if the sync thread wraps the event storage thread. 
We need to make sure though that code executed on the event storage thread **never** uses any `andWait`

PR #63, opened 2016-09-13
Fix crashing/failing tests

typfel:
## What is fixed?
Some tests where failing since they where crashing.

## What changed
- the `ZMClientMessageTranscoderTests` were failing since they didn't create requests since the self client was missing.
- the `ZMAssetTranscoderTests` were crashing since we were mocking the `ZMImageMessage` class and  creating  an instance of it. This led to a crash when the instances where being deallocated. To avoid the crash we create our instance before we start mocking the class.


marcoconti83:
👍 less failing tests!!

PR #66, opened 2016-09-14
Change event decoder implementation to be synchronous

daehn:
# What's in this PR?

* This PR fixes the issue discussed in https://github.com/wireapp/wire-ios-sync-engine/pull/62
* The `-processEvents:block:` method of the `EventDecoder` is now synchronous which avoids processing the same events multiple times before they have been deleted.


marcoconti83:
Can you please try to run the tests on this with the CoreData concurrency debug flag switched on? So we are sure that we do everything on the right thread.

marcoconti83:
👍 It looks so much easier to follow now!

daehn:
@marcoconti83 I tried to run it with the CD debug flag, but this currently asserts in the creation of the `syncMOC`, which will be fixed when https://github.com/wireapp/wire-ios-data-model/pull/43 has been merged.

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -134,90 +134,76 @@ extension NSManagedObjectContext {
...
-                // encryption context only if we stored all events in the database
-                strongSelf.eventMOC.performGroupedBlockAndWait {
-                    
-                    // Decrypt the events and insert them counting upwards from the highest index in the DB
-                    for (idx, event) in newUpdateEvents.enumerate() {
-                        _ = StoredUpdateEvent.create(event, managedObjectContext: strongSelf.eventMOC, index: idx + startIndex + 1)
-                    }
-                    
-                    strongSelf.eventMOC.saveOrRollback()
+                // Decrypt the events and insert them counting upwards from the highest index in the DB
```
typfel: Comment is a bit misleading since the events are already decrypted here.

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -134,90 +134,76 @@ extension NSManagedObjectContext {
...
-            guard events.storedEvents.count > 0 else { return }
+        let events = fetchNextEventsBatch()
+        guard events.storedEvents.count > 0 else { return }
 
-            self.processBatch(events.updateEvents, storedEvents: events.storedEvents, block: consumeBlock) {
-                self.process(consumeBlock)
-            }
-        }
+        processBatch(events.updateEvents, storedEvents: events.storedEvents, block: consumeBlock)
+        process(consumeBlock)
```
typfel: Why the recursive call to process?

daehn: It's only fetching one batch (500 update events) and processes them, then it calls itself again to process the next batch until there are no more events.

typfel: Ah ok

marcoconti83:
Data model released as 17.0.4

daehn:
@marcoconti83 @typfel I bumped data model and run the event decoder test with the assertions enabled ✅. I made some changes in the test setup in order for the assertions not to fire there.

PR #70, opened 2016-09-14
Fix failing tests in ConversationTests+OTR

daehn:
# What's in this PR?

Fixes multiple failing tests in `ConversationTests+OTR`, which all where related to inserting a failed to decrypt system message.

### _Why were these tests failing?_

There were multiple issues with the tests, the first one being that a 💣 has been inserted as (plain) text, which is normally done when the sender can not establish a session with the receiver but still needs to send the message. On the receiving side messages with this string are discarded and no system message is inserted (which is another issue that can be discussed). 

Another issue was a missing save after the system message has been inserted. If only one update event is processed and that one can not be decrypted, the system message is appended from a call made inside the `EventDecoder`, the event decoder then does not have any events to propagate to the `ZMSyncStrategy` and thus will not call the consume block, which lead to no save being enqueued. ~~There are multiple solution to this issue, we now manually enqueue a save when inserting the can not decrypt system message.~~

**The behavior of the `EventDecoder` has been changed to make sure the passed in consume block is always called at least one, in case there was only one event and that event could not be decrypted it will now be called with an empty array.**

### DIFF
--- a/Source/E2EE/CBCryptoBox+UpdateEvents.m
+++ b/Source/E2EE/CBCryptoBox+UpdateEvents.m
@@ -85,15 +85,17 @@ NSString * CBErrorCodeToString(CryptoboxError cryptoBoxError);
     && (error.code == CryptoboxErrorCryptoboxOutdatedMessage || error.code == CryptoboxErrorCryptoboxDuplicateMessage);
     
     // do not notify user if it's just a duplicated one
-    if(didFailBecauseDuplicated) {
+    if (didFailBecauseDuplicated) {
         return;
     }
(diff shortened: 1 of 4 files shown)


typfel:
I can't find the manual save on decryption error in the diff.

daehn:
You're right, I forgot that change was in `wire-ios-data-model`, will open a PR there as well.

daehn:
@typfel I updated the behavior of the `EventDecoder` to make the behavior more transparent from the caller site, the passed in block is now always called at least once, no need to enqueue an additional save.

PR #72, opened 2016-09-14
Restore transient flag from stored event, which fixes push channel test

typfel:
## Why was the test failing
The failing push channel test was testing that we ignore transient events when deciding where to resume from the notification queue. But we were not restoring the transient flag after re-creating an event from a stored event.

## What changed
We now pass in the transient flag when constructing an event from a stored event.


marcoconti83:
@typfel please also port the transport change to Swift 3

PR #75, opened 2016-09-14
Use correct name background activity name

basine:
We changed the name for the background activity in the project, but not in the tests.


Source/UserSession/BackgroundAPNSConfirmationStatus.swift
```diff
@@ -23,7 +23,8 @@ public class BackgroundAPNSConfirmationStatus : NSObject {
     
     /// Switch for sending delivery receipts
     public static let sendDeliveryReceipts : Bool = true
-
+    static let backgroundNameBase : String = "Sending confirmation message with nonce "
```
typfel: This will lead to double spaces 😱

PR #77, opened 2016-09-14
Retry Travis for wire-ios-sync-engine

daehn:
# Whats in this PR?

This is a test to see if Travis can handle `wire-ios-sync-engine` now that the tests are faster.


daehn:
Let's try this again using the Xcode 8 image after `wire-iso-sync-engine` has been ported to Swift 3.

daehn:
Travis now runs the `wire-iso-sync-engine` tests (at least most of the time). There are still 7 tests failing on Travis which need to be fixed before this can be merged.

PR #88, opened 2016-09-30
Add support for canceling proxy requests

typfel:
## Why these changes?
For the updated giphy feature we allow searching for gifs as you type. For this this work correctly we need to be able to cancel previous searches.

## What Changed
- `ZMUserSession.proxiedRequestWithPath:` now returns a `ProxyRequest` object.
- added `ZMUserSession.cancelProxiedRequest:`

### DIFF
--- a/Source/UserSession/ProxiedRequestsStatus.swift
+++ b/Source/UserSession/ProxiedRequestsStatus.swift
@@ -46,8 +46,8 @@ public final class ProxiedRequestsStatus: NSObject {
     private let requestCancellation : ZMRequestCancellation
 
     /// List of requests to be sent to backend
-    public var pendingRequests : [ProxyRequest] = []
-    public var executedRequests : [(ProxyRequest, ZMTaskIdentifier)] = []
+    public var pendingRequests : Set<ProxyRequest> = Set()
+    public var executedRequests : [ProxyRequest : ZMTaskIdentifier] = [:]
     
     public init(requestCancellation: ZMRequestCancellation) {
         self.requestCancellation = requestCancellation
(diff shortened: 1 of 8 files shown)


daehn:
Is the `DO NOT MERGE` still valid?

typfel:
Yes, unless we want to make hotfixes for the crashes in RC

Source/UserSession/ProxiedRequestsStatus.swift
```diff
@@ -19,15 +19,54 @@
...
+    @objc(addRequest:)
+    public func add(request: ProxyRequest) {
+        pendingRequests.append(request)
+    }
     
-    public func addRequest(_ type:ProxiedRequestType, path: String, method: ZMTransportRequestMethod = .methodGET, callback: ((Data?, HTTPURLResponse?, NSError?) -> Void)?) {
-        pendingRequests.append(Request(type, path, method, callback))
+    @objc(cancelRequest:)
+    public func cancel(request: ProxyRequest) {
+        if let index = pendingRequests.index(of: request) {
```
marcoconti83: Is there a reason why we use an `Array` instead of a `Set` for `pendingRequests`? `indexOf`+`remove` is `O(n)` with an `Array`, with a `Set` this would be `O(1)`

Source/UserSession/ProxiedRequestsStatus.swift
```diff
@@ -19,15 +19,54 @@
...
+            pendingRequests.remove(at: index)
+        }
+        
+        if let index = executedRequests.index(where: {$0.0 == request }) {
```
marcoconti83: Same as above, `Array` vs. `Set`

Source/UserSession/ProxiedRequestsStatus.swift
typfel: We want execute the requests in order. We could use an ordered set I guess, but also I don't expect these arrays to be very large.

marcoconti83: The order is important if the requests are serialized (we wait for the previous one to finish before starting a new one). If they all go out at the same time, it does not matter as they will be completed at random different times anyway depending on the network. I'm not sure which one is the case here.

Source/UserSession/ProxiedRequestsStatus.swift
```diff
@@ -46,26 +46,23 @@ public final class ProxiedRequestsStatus: NSObject {
...
     @objc(cancelRequest:)
     public func cancel(request: ProxyRequest) {
-        if let index = pendingRequests.index(of: request) {
-            pendingRequests.remove(at: index)
-        }
+        pendingRequests.remove(request)
         
-        if let index = executedRequests.index(where: {$0.0 == request }) {
-            let (_, taskIdentifier) = executedRequests.remove(at: index)
+        if let taskIdentifier = executedRequests[request] {
```
marcoconti83: In the code with arrays, we were removing the request from `executedRequests`. Now we don't anymore, it will grow forever.

PR #92, opened 2016-10-12
Remove ZMBadge class

marcoconti83:
`ZMBadge` had become just a container for `ZMApplication` that would bloat a lot of `init` parameter lists, while not doing anything significant. If anything it could be replaced with a protocol that `ZMApplications` implement, but since the badge is already part of the `ZMApplication` protocol, I just removed `ZMBadge` altogether

# Changes
Remove `ZMBadge` in favor of `ZMApplication`


mikeger:
@marcoconti83 it was a mac compatibility wrapper most likely. Not needed now anyways.

PR #93, opened 2016-10-13
Fix failed to decrypt when receiving messages in background

marcoconti83:
When in background, APNS enabled, and receiving more than ~10 messages from the same user without opening the app, upon opening the app later some error messages ("failed to decrypt") will be inserted in the conversation, even if no message was missing.

This was causing by trying to decrypt the message twice (once, successfully, from APNS and once later, when coming back to foreground and fetching the notification stream). It is not expected to be able to decrypt the same message twice, but normally cryptobox will return a "duplicated message" error, which we expect and ignore. 

However, if the session with the sender had advanced the key enough times (and it does, because APNS cause a confirmation message to be sent back to the sender, which advance the session), the original key used for the first message is already gone outside of the history of the session and instead of "duplicated message" cryptobox will return "invalid message".

The solution is not to rely on cryptobox to tell us that the message is duplicated (i.e. previously received) but to keep track ourselves.

# Changes
Added a list of notification IDs that we already received through APNS. The list is used to filter out events from the notification stream, and then discarded (we will never fetch the same event from the notification stream twice, so we don't need to filter those events anymore).

# Known issues
This fixes the most common cause of event not being decrypted, but there are still some edge cases that were introduced with the delivery confirmation behavior.

- If APNS delivery is failing and some messages are not pushed, then it recovers later (causing delivery confirmations to be sent), when coming back online we might not be able to decrypt the messages that were missed for the same reason (the session key moved ahead too much).
- if there is a muted conversation (which does not cause APNS) with user B and there is a one-to-one conversation between A and B that is not silenced, if B user writes enough messages in the silenced conversation while A is in the background, A might not be able to decrypt messages from the muted conversation when fetching the notification stream

A better solution would be to have APNS cause a fetch of the entire notification stream. However we should investigate how this will affect APNS reliability as the app will use more data and battery and most likely be more penalized by the OS.

### DIFF
--- a/Source/Synchronization/Decoding/EventDecoder.swift
+++ b/Source/Synchronization/Decoding/EventDecoder.swift
@@ -23,8 +23,18 @@ import WireMessageStrategy
 
 private let zmLog = ZMSLog(tag: "EventDecoder")
 
+/// Key used in persistent store metadata
 private let previouslyReceivedEventIDsKey = "zm_previouslyReceivedEventIDsKey"
 
+/// Holds a list of received event IDs
+@objc public protocol PreviouslyReceivedEventIDsCollection : NSObjectProtocol {
+    
+    func discardListOfAlreadyReceivedPushEventIDs()
+}
+
+
+
+/// Decodes and stores events from various sources to be processed later
 @objc public final class EventDecoder: NSObject {
     
     public typealias ConsumeBlock = (([ZMUpdateEvent]) -> Void)
@@ -50,7 +60,6 @@ private let previouslyReceivedEventIDsKey = "zm_previouslyReceivedEventIDsKey"
         super.init()
         self.createReceivedPushEventIDsStore()
     }
-
 }
 
 // MARK: - Process events
@@ -165,11 +174,6 @@ extension EventDecoder {
         return Set(self.eventMOC.persistentStoreMetadata(forKey: previouslyReceivedEventIDsKey) as! [String])
     }
     
-    /// Discards the list of already received events
-    public func discardListOfAlreadyReceivedPushEventIDs() {
-        self.eventMOC.setPersistentStoreMetadata(NSArray(), forKey: previouslyReceivedEventIDsKey)
-    }
-    
     /// Store received event IDs 
     fileprivate func storeReceivedPushEventIDs(from: [ZMUpdateEvent]) {
         let uuidToAdd = from
@@ -191,5 +195,13 @@ extension EventDecoder {
                 return event
             }
         }
+    }
+}
+
+extension EventDecoder : PreviouslyReceivedEventIDsCollection {
+    
+    /// Discards the list of already received events
+    public func discardListOfAlreadyReceivedPushEventIDs() {
+        self.eventMOC.setPersistentStoreMetadata(NSArray(), forKey: previouslyReceivedEventIDsKey)
     }
 }
--- a/Source/Synchronization/Decoding/EventDecoder.swift
+++ b/Source/Synchronization/Decoding/EventDecoder.swift
@@ -58,7 +58,7 @@ private let previouslyReceivedEventIDsKey = "zm_previouslyReceivedEventIDsKey"
         self.eventMOC = eventMOC
         self.syncMOC = syncMOC
         super.init()
-        self.createReceivedPushEventIDsStore()
+        self.createReceivedPushEventIDsStoreIfNecessary()
     }
 }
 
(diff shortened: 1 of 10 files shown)


Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -228,16 +140,68 @@ extension NSManagedObjectContext {
...
+    fileprivate func createReceivedPushEventIDsStore() {
+        if self.eventMOC.persistentStoreMetadata(forKey: previouslyReceivedEventIDsKey) as? [String] == nil {
+            self.eventMOC.setPersistentStoreMetadata(NSArray(), forKey: previouslyReceivedEventIDsKey)
+        }
+    }
+    
+    
+    /// List of already received event IDs
+    fileprivate var alreadyReceivedPushEventIDs : Set<UUID> {
+        let array = self.eventMOC.persistentStoreMetadata(forKey: previouslyReceivedEventIDsKey) as! [String]
```
mikeger: Would the loading code be executed every time `alreadyReceivedPushEventIDs` is called?

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -153,13 +58,20 @@ extension NSManagedObjectContext {
         self.eventMOC = eventMOC
         self.syncMOC = syncMOC
         super.init()
+        self.createReceivedPushEventIDsStore()
```
mikeger: Is the list of event id's is not persisted between app runs?

zmessaging-cocoa.xcodeproj/xcshareddata/xcschemes/zmessaging-ios.xcscheme
```diff
@@ -102,6 +102,11 @@
             isEnabled = "NO">
          </EnvironmentVariable>
          <EnvironmentVariable
+            key = "OS_ACTIVITY_MODE"
+            value = "disable"
+            isEnabled = "YES">
+         </EnvironmentVariable>
+         <EnvironmentVariable
```
mikeger: Some new debug key?

Source/Synchronization/Decoding/EventDecoder.swift
typfel: Maybe it's more clear what's happening if this method is called something like `createReceivedPushEventIDsStoreIfNecessary`

zmessaging-cocoa.xcodeproj/xcshareddata/xcschemes/zmessaging-ios.xcscheme
marcoconti83: This is new in Xcode 8

Source/Synchronization/Decoding/EventDecoder.swift
marcoconti83: Yes, we do load from disk every time (which is called once per APNS or per notification stream fetch). We could cache it in memory, but caching has a lot of edge cases that makes it dangerous. For example, the metadata is saved when the context is saved. However, we use `saveOrRollback`, which means that the context will be rolled back in case of error. So if we keep a cached version in memory, we would have to listen for rollbacks to invalidate the cache and reload from disk.

Source/Synchronization/Decoding/EventDecoder.swift
marcoconti83: It is persisted between runs. But the first time you install this new version, the list is not there. I changed the name as suggested by Jacob to make it more clear.

zmessaging-cocoa.xcodeproj/xcshareddata/xcschemes/zmessaging-ios.xcscheme
mikeger: Nice!

Source/Synchronization/Decoding/EventDecoder.swift
mikeger: Wouldn't it be a performance hit then, if we persist the list between runs, and read it out from disk for every notification?

marcoconti83: Consider that the list is a list of one string per message and I expect it to be small enough.E.g. if you get 50 messages before you open the app, it's 50 strings. Once you open the app, all those 50 strings are discarded and now the list has size 0.

mikeger: Ok, I was confused, I thought we keep the list forever & never discard it.

PR #94, opened 2016-10-14
Fix non passing tests due to change in DataModel

Einh06:
**What's in this PR**
This PR fixes tests that were broke changing the way contexts are initialized.

### DIFF
--- a/Tests/Source/MessagingTest.m
+++ b/Tests/Source/MessagingTest.m
@@ -659,6 +659,20 @@ static int32_t eventIdCounter;
     XCTAssertNil(error, @"Error establishing session: %@", error);
 }
 
+- (UserClient *)setupSelfClientInMoc:(NSManagedObjectContext *)moc;
+{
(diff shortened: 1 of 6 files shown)


Tests/Source/MessagingTest.m
```diff
@@ -659,6 +659,20 @@ - (void)establishSessionWithClient:(UserClient *)userClient
...
+    client.remoteIdentifier = @"identifier";
+    client.user = selfUser;
+    
+    [moc setPersistentStoreMetadata:client.remoteIdentifier forKey:ZMPersistedClientIdKey];
+    [moc saveOrRollback];
+    
+    return client;
+}
+
 - (UserClient *)createSelfClient
```
marcoconti83: I think it would be better to have `createSelfClient` internally call `setupSelfClientInMoc` so that we don't have two implementations of the same

marcoconti83:
@Einh06 What's the status of this?

Einh06:
@marcoconti83 This is pushed now.

PR #95, opened 2016-10-14
Ephemeral messages

basine:
**In this PR**

1) Obfuscating push notifications for ephemeral messages.
@jenskasemets There is a new string that needs translations.

2) Adding some integration tests for ephemeral messages.

Needs releases for:

- [x] ZMCDataModel
- [x] WireMessageStrategy
- [x] Rerun tests after merging https://github.com/wireapp/wire-ios-sync-engine/pull/94


Source/Notifications/Push notifications/Notification Types/MessageNotifications/ZMLocalNotificationForMessage.swift
```diff
@@ -41,12 +41,18 @@ extension NotificationForMessage {
         }
     }
     
-    public func configureNotification(_ message: MessageType) -> UILocalNotification {
+    public func configureNotification(_ message: MessageType, isEphemeral: Bool = false) -> UILocalNotification {
         let notification = UILocalNotification()
-        let shouldHideContent = message.managedObjectContext!.value(forKey: ZMShouldHideNotificationContentKey)
-        if let shouldHideContent = shouldHideContent as? NSNumber , shouldHideContent.boolValue == true {
-            notification.alertBody = ZMPushStringDefault.localized()
+        var shouldHideContent = isEphemeral
```
marcoconti83: Just sharing my experience: if you want to avoid having this as a `var`, you can declare it `let` and initialize it like this:
```
let shouldHideContent : Bool
if (...) { 
    shouldHideContent = true
} else {
    shouldHideContent = isEphemeral
}
```
The compiler will not complain because every branch will initialize the `let` before it's ever used. I find this style very cool :)

PR #99, opened 2016-10-19
CallKit

mikeger:
# Purpose

Adds support for CallKit calling screens and Phone app integration.
# Details

ZMCallKitDelegate is the main logical part of CallKit integration of wire-ios-sync-engine. CallKit is the iOS VoIP calling integration framework that allows to place VoIP calls with the same priority level as other (GSM) calls.

CallKit is supported from iOS 10. On the devices that do not support it we still show the notifications.

It is possible to disable CallKit and use old-style notifications logic: `+[ZMUserSession setUseCallKit:]`.

CallKit integration consists of 2 parts:

 I.  Showing iOS calling UI when call happens
     I.a Receive calls
     I.b Place calls
II. Interaction with Phone app and redialing existing calls from there

Using CallKit means asking iOS to start and end the call, so it can show the proper calling UI and play the ringtone.

 Flow for receieving the call (I.a):
img_4952
1. BE sends the new call state in conversation via push
2. ZMCallStateTranscoder decode the payload and update the conversation/ZMVoiceChannel fields
3. ZMVoiceChannel observer sends the update to ZMCallKitDelegate -[ZMCallKitDelegate voiceChannelStateDidChange:]
4. ZMCallKitDelegate indicates the call to CallKit's CXProvider in -[ZMCallKitDelegate indicateIncomingCallInConversation:]
5. CXProvider approves the call and informs ZMCallKitDelegate that call is possible in -[ZMCallKitDelegate provider:performStartCallAction:]
6. ZMCallKitDelegate joins the call with -[ZMVoiceChannel join]

Flow for sending the call (I.b):

img_4953
1. API consumer (UI app) calls -[ZMVoiceChannel joinInUserSession:]. This call is forwarded to -[ZMCallKitDelegate requestStartCallInConversation:]
2. ZMCallKitDelegate indicates the call to CallKit's CXCallController, that asks CXProvider if call is possible. 
3. CXProvider is indicating the call is possible with the callback -[ZMCallKitDelegate provider:performStartCallAction:]
4. ZMCallKitDelegate joins the call with -[ZMVoiceChannel join]

Flow for interaction with last calls / Phone app:
1. The app is launched or brought to foreground
2. UIApplicationDelegate receives the call for continuing the User Activity that is forwarded to ZMUserSession
3. ZMUserSession forwards the call to -[ZMCallKitDelegate continueUserActivity:]
4. ZMCallKitDelegate looks up the caller from the NSUserActivity and starts the call
# Requires

[x] https://github.com/wireapp/wire-ios-data-model/pull/63


Source/Notifications/Push notifications/Helpers/NotificationSounds.swift
```diff
@@ -23,24 +23,26 @@ let ZMLocalNotificationRingingDefaultSoundName = "ringing_from_them_long.caf"
...
+    public static func notificationNewMessageSoundName() -> String {
+        return ZMCustomSoundName("ZMMessageSoundName") ?? ZMLocalNotificationNewMessageDefaultSoundName
+    }
 }
 
-func ZMLocalNotificationNewMessageSoundName() -> String {
-    return ZMCustomSoundName("ZMMessageSoundName") ?? ZMLocalNotificationNewMessageDefaultSoundName
+func ZMCustomSoundName(_ key: String) -> String? {
+    guard let soundName = UserDefaults.standard.object(forKey: key) as? String else { return nil }
+    print("soundName: \(soundName)")
```
basine: print

Source/Calling/ZMCallKitDelegate.m
```diff
@@ -0,0 +1,657 @@
...
+    
+    if (1 == contacts.count) {
+        ZMConversation *callConversation = [ZMConversation resolveConversationForPersons:contacts
+                                                                               inContext:self.userSession.managedObjectContext];
+        if (nil != callConversation) {
+            if (isVideo) {
+                NSError *joinError = nil;
+                [callConversation.voiceChannel joinVideoCall:&joinError inUserSession:self.userSession];
+                if (nil != joinError) {
+                    ZMLogError(@"Cannot start the video call: %@", joinError);
```
mikeger: I think it mostly to reassure that we handle the error case, right now the method does not really throw the meaningful errors.
Regarding return the method should return NO if the activity type is not known, it's not too clear so I updated the method description.

Source/Calling/ZMCallKitDelegate.m
```diff
@@ -0,0 +1,657 @@
...
++ (NSSet *)activeStates {
+    return [NSSet setWithObjects:@(ZMVoiceChannelStateOutgoingCall),
+            @(ZMVoiceChannelStateOutgoingCallInactive),
+            @(ZMVoiceChannelStateSelfIsJoiningActiveChannel),
+            @(ZMVoiceChannelStateSelfConnectedToActiveChannel),
+            @(ZMVoiceChannelStateIncomingCall),
+            @(ZMVoiceChannelStateIncomingCallInactive), nil];
+}
+
+- (BOOL)inActiveState
```
mikeger: Thanks for the feedback, I was also not soo sure, renamed it to `inNonIdleState`

Source/Calling/ZMFlowSync.m
```diff
@@ -162,10 +162,10 @@ - (NSArray *)requestGenerators;
 
 - (ZMTransportRequest *)nextRequest
 {
-    if (!self.pushChannelIsOpen) {
+    if (!self.pushChannelIsOpen && ![ZMUserSession useCallKit]) {
```
mikeger: I will need to investigate on it more, since beta testers are reporting that sometimes the call cannot be established, might be something connected to push channel not being open.

Source/Calling/ZMCallKitDelegate.m
```diff
@@ -0,0 +1,657 @@
...
+{
+    CXEndCallAction *endCallAction = [[CXEndCallAction alloc] initWithCallUUID:conversation.remoteIdentifier];
+    
+    CXTransaction *endCallTransaction = [[CXTransaction alloc] initWithAction:endCallAction];
+    
+    [self.callController requestTransaction:endCallTransaction completion:^(NSError * _Nullable error) {
+        if (nil != error) {
+            ZMLogError(@"Cannot end call: %@", error);
+        }
+        [conversation.voiceChannel leave];
```
basine: I mean the CallKit state - just a bit worried that while we end the call internally the user is left in some unconsistent state

Source/Calling/ZMCallKitDelegate.m
```diff
@@ -0,0 +1,653 @@
...
+    case ZMVoiceChannelStateSelfConnectedToActiveChannel:
+            [self.provider reportOutgoingCallWithUUID:conversation.remoteIdentifier
+                                      connectedAtDate:[NSDate date]];
+            conversation.voiceChannel.callStartDate = [NSDate date];
+        break;
+    case ZMVoiceChannelStateNoActiveUsers:
+        {
+            CXEndCallAction *endCallAction = [[CXEndCallAction alloc] initWithCallUUID:conversation.remoteIdentifier];
+            
+            CXTransaction *endCallTransaction = [[CXTransaction alloc] initWithAction:endCallAction];
```
basine: Careful: voiceChannel is a transient property which will be recreated when the conversation is faulted and unfaulted. This might have unexpected side-effects for variables stored along with it.

Source/Calling/ZMFlowSync.m
basine: I think inside useCallKit you could check if the pushChannel is not open and if that's the case, just call  [self.transportSession restartPushChannel]; That might do the trick.

Source/Calling/ZMCallKitDelegate.m
mikeger: You most likely wanted to comment on `conversation.voiceChannel.callStartDate`. Yep I realise the danger, but in case the call start date is not set I do the best effort on UI: start counting the call timer from when UI appeared. Otherwise it has to be persisted in some way. May be you have a better idea where to put it.

Source/Calling/ZMFlowSync.m
mikeger: I will try to call `[self.transportSession restartPushChannel];` in `-join`.

Source/Calling/ZMCallKitDelegate.m
basine: yep, that was my intention :D 
I wasn't sure what you are using start date for. If it's nothing crucial, I guess it doesn't matter where you store it.

Source/Calling/ZMVoiceChannel+CallFlow.m
```diff
@@ -69,6 +69,8 @@ - (void)startOrCancelTimerForState:(ZMVoiceChannelState)state
 - (void)joinInUserSession:(ZMUserSession *)userSession
 {
     if ([ZMUserSession useCallKit]) {
+        [userSession.transportSession restartPushChannel];
```
basine: did you check if it's safe to call this when the push channel is already open? we wouldn't need to send another request then.

mikeger: Currently if the push channel is open, then it would close it and reopen. Looks like a sub-optimal behavior. It looks tricky to check if the push channel is open now, but I will try to make a property and check if it is necessary to call `restartPushChannel`.

mikeger:
@basine please re-review. Current version should work in the old way until UI says to use CallKit, so it should be "safe" to merge even not releasing the feature immediately. Thanks!

Source/Calling/ZMVoiceChannel+CallFlow.m
```diff
@@ -69,8 +69,10 @@ - (void)startOrCancelTimerForState:(ZMVoiceChannelState)state
 - (void)joinInUserSession:(ZMUserSession *)userSession
 {
     if ([ZMUserSession useCallKit]) {
-        [userSession.transportSession restartPushChannel];
-        
+        // Push channel must be open in order to process the call signalling
+        if (!userSession.pushChannelIsOpen) {
+            [userSession.transportSession restartPushChannel];
```
basine: If we receive a call in the background, we are using a background NSURLSession for the push channel request which might not be immediately executed. We might want to switch to the foreground or voip URLSession here.

Source/Calling/ZMVoiceChannel+VideoCalling.m
```diff
@@ -40,6 +40,10 @@ @implementation ZMVoiceChannel (VideoCalling)
 - (BOOL)joinVideoCall:(NSError **)error inUserSession:(ZMUserSession *)userSession
 {
     if ([ZMUserSession useCallKit]) {
+        // Push channel must be open in order to process the call signalling
+        if (!userSession.pushChannelIsOpen) {
+            [userSession.transportSession restartPushChannel];
```
basine: see above - this might use the background NSURLSession

Source/Calling/ZMVoiceChannel+CallFlow.m
mikeger: I'm not sure; do we use `NSURLSession` for push channel? I thought it's a socket, nothing to do with URL session?

mikeger:
@basine now can be merged

PR #104, opened 2016-10-26
Fetch the notification stream when receiving a push notification

daehn:
# What's in this PR?

**TL;DR: This PR adds the logic to fetch the notification stream (`/notifications`) after receiving a push notification. The purpose is to reduce possible decryption errors due to out of sync clients as well as reduce the amount of processing that is necessary when opening the app after a while.**
### Behavior before the changes:

When receiving a push notification with content we were performing a pingback to the backend to notify it that we received the notification, if we failed to perform this pingback request in 30 seconds the backend issues an alert notification. If the received notification was a notice (only containing the notification ID and no other content), we were fetching the content and cancelling the fallback notification at the same time. This was done by the `PingBackRequestStrategy` and  `PushNoticeRequestStrategy`.
### Behavior after the changes:

If we receive a notification we do not care if it was a notice or if it contained any content, the `BackgroundAPNSPingBackStatus` still enqueues the notification as we need to cancel the fallback notification.  The `PingBackRequestStrategy` and  `PushNoticeRequestStrategy` have been removed, but the `ZMMissingUpdateEventsTranscoder`, will pick up notifications from this queue if the application is running in the background. The transcoder then performs a notification stream fetch starting at the last notificationID as usual, but specifies an additional query parameter `cancel_fallback` in order to cancel the fallback notification. After the event has been fetched it is forwarded to the ping back status which will remove the ID from it's queue and itself forward the events to be decrypted and processed by the sync strategy like before.

A couple of things changed as it can happen that we missed some notifications in the meantime, in this case we will receive multiple events in the notification stream and forward all of them, but we only want to generate local notifications for new events (as the user will have received fallbacks for the missed ones and we don't want to have multiple notifications shown for the same event). As the notification dispatcher will only generate notifications for events with a source of type `.pushNotification` we set this source only for events that have a UUID greater than or equal to the last notification ID used to fetch the stream (see https://github.com/wireapp/wire-ios-transport/pull/17).

In order to retry a request in case we received a try again later error from the backend we now only remove the notifications from the pingback stack if we either failed or received a first batch of events or the first request was successful.
---
- [x] This PR depends on a new `wire-ios-request-strategy` release after https://github.com/wireapp/wire-ios-request-strategy/pull/5 has been merged.
- [x] This PR depends on a new `wire-ios-mock-transport` release after https://github.com/wireapp/wire-ios-mocktransport/pull/11 has been merged.


basine:
Did you test that on the client?

Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
```diff
@@ -123,17 +130,20 @@ - (void)appendPotentialGapSystemMessageIfNeededWithResponse:(ZMTransportResponse
...
+    ZMSTimePoint *tp = [ZMSTimePoint timePointWithInterval:10 label:NSStringFromClass(self.class)];
+    NSArray *eventsDictionaries = [self.class eventDictionariesFromPayload:payload];
     
     NSMutableArray *parsedEvents = [NSMutableArray array];
     NSMutableDictionary *lastCallStateEvents = [NSMutableDictionary dictionary];
     NSUUID *latestEventId = nil;
     
     for(NSDictionary *eventDict in eventsDictionaries) {
-        NSArray *events = [ZMUpdateEvent eventsArrayFromPushChannelData:eventDict];
+        NSUUID *sourceThreshold = self.notificationEventsToCancel.identifier;
```
daehn: Good point, I don't know how we could avoid this though, we could add a leeway which then could lead to a fallback and the full text notification being shown, but that might be better than no notification being shown for this event.

daehn:
Yes I tested it for the last couple of days and it worked pretty well so far, I think I run into the issue you mentioned, I once noticed missing one of two notifications that came right after each other where I did not get a fallback though.

daehn:
After merging in develop a couple of tests were failing as the logging assertions now fails tests.
This needs a new release of `wire-ios-mock-transport` after https://github.com/wireapp/wire-ios-mocktransport/pull/11 has been merged.

marcoconti83:
New mocktransport has been released. Is this ready to merge in?

marcoconti83:
I already bumped mocktransport on sync-engine develop

daehn:
Yes, can be merged

PR #108, opened 2016-11-01
Use correct user collection for member join / leave notifications

basine:
We were using the wrong property (removedUsers / addedUsers instead of users) on ZMSystemMessage to create the push notification.

### DIFF
--- a/Source/Notifications/Push notifications/Notification Types/MessageNotifications/ZMLocalNotificationForSystemMessage.swift
+++ b/Source/Notifications/Push notifications/Notification Types/MessageNotifications/ZMLocalNotificationForSystemMessage.swift
@@ -42,7 +42,7 @@ final public class ZMLocalNotificationForSystemMessage : ZMLocalNotification, No
         else {return nil}
         
         // We don't want to create notifications when a user leaves the conversation
-        if message.systemMessageType == .participantsRemoved, let removedUser = message.removedUsers.first, message.sender == removedUser {
+        if message.systemMessageType == .participantsRemoved, let removedUser = message.users.first, message.sender == removedUser {
             return nil
         }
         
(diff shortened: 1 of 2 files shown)


Source/Notifications/Push notifications/Notification Types/MessageNotifications/ZMLocalNotificationForSystemMessage.swift
```diff
@@ -81,12 +80,8 @@ final public class ZMLocalNotificationForSystemMessage : ZMLocalNotification, No
         var user: ZMUser?
         var key : NSString = (isLeaveEvent ? ZMPushStringMemberLeaveMany : ZMPushStringMemberJoinMany) as NSString
         if userCount == 1 {
-            if (users.first == message.sender) {
-                key = ZMPushStringMemberLeaveSender as NSString
-            } else {
-                user = users.first
-                key = (isLeaveEvent ? ZMPushStringMemberLeave : ZMPushStringMemberJoin) as NSString
-            }
+            user = message.users.first
```
marcoconti83: What happened to `ZMPushStringMemberLeaveSender`?

Source/Notifications/Push notifications/Notification Types/MessageNotifications/ZMLocalNotificationForSystemMessage.swift
```diff
@@ -70,9 +70,8 @@ final public class ZMLocalNotificationForSystemMessage : ZMLocalNotification, No
     func alertBodyForParticipantEvents(_ message: ZMSystemMessage) -> String {
         let isLeaveEvent = (message.systemMessageType == .participantsRemoved)
         let isCopy = (userCount != 0)
-        let users = isLeaveEvent ? message.removedUsers : message.addedUsers
         
-        userCount = userCount + users.count
+        userCount = userCount + message.users.count
```
marcoconti83: So if we are not using `message.removedUsers` or `message.AddedUsers` here, what are those fields for? Is it for the "You did not use this device in a long time" message?

basine: Exactly - it's for the potential gap system message

PR #112, opened 2016-11-07
Fix/failing tests

Einh06:
**What's in this PR**
This PR fixes tests broke by the new wire-ios-data-model

- [ ] wire-ios-data-model release


daehn:
@Einh06 Can you bump the dependency, I just released a new message strategy

PR #120, opened 2016-11-10
Respect call timeout; Hide the call that is picked up remotely

mikeger:
# Issue

1. Call used to continue to ring indefinitely
2. If the call is picked up on the other device, CallKit used to continue ringing

# Fixes

Correctly observe the call state transient properties and respect additional call states.

### DIFF
--- a/Source/Calling/ZMCallKitDelegate.m
+++ b/Source/Calling/ZMCallKitDelegate.m
@@ -97,6 +97,7 @@ NS_ASSUME_NONNULL_BEGIN
 @end
 
 @interface ZMCallKitDelegate (VoiceChannelObserver)
+- (void)managedObjectContextDidSave:(NSNotification *)notification;
 - (void)conversationVoiceChannelStateDidChange:(ZMConversation *)conversation;
 @end
 
@@ -267,6 +268,11 @@ NS_ASSUME_NONNULL_END
         [[NSNotificationCenter defaultCenter] addObserver:self
                                                  selector:@selector(managedObjectsDidChange:)
                                                      name:NSManagedObjectContextObjectsDidChangeNotification
+                                                   object:userSession.managedObjectContext];
+        
+        [[NSNotificationCenter defaultCenter] addObserver:self
+                                                 selector:@selector(managedObjectContextDidSave:)
+                                                     name:NSManagedObjectContextDidSaveNotification
                                                    object:userSession.managedObjectContext];
         
         [[NSNotificationCenter defaultCenter] addObserver:self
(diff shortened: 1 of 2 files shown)


Source/Calling/ZMCallKitDelegate.m
```diff
@@ -485,6 +491,32 @@ - (void)managedObjectsDidChange:(NSNotification *)notification
     handleChangeset(insertedSet);
 }
 
+- (void)managedObjectContextDidSave:(NSNotification *)notification
+{
+    NOT_USED(notification);
+    
+    for (NSString *conversationID in self.lastConversationsState.allKeys.copy) {
```
marcoconti83: What's in `lastConversationsState`? All conversations, or only those that have an ongoing call?

typfel:
Why do you need observer when the context save instead of using the VoiceChannelObservers?

Source/Calling/ZMCallKitDelegate.m
mikeger: @marcoconti83 @typfel  This is the map of conversation ids to last known state; the conversations appear there, as we observe the changes in the non-transient fields of `ZMConversation` via context save notification. This method is needed to catch the changes on transient fields, like `callTimedOut` that also affect `-`-[ZMVoiceChannel state]`.

mikeger:
@typfel we cannot use voice channel observer because it should run when the app is not active, therefore not that much memory / computation power available. Conversation voice channel observer creates snapshots of all conversations.

Source/Calling/ZMCallKitDelegate.m
marcoconti83: I'm a bit concerned about discarding the notification and going through all conversations every time there is a save notification, which is very often (also it's fired 2-3 times for every single `save` call). I think it would be more efficient to just go through the conversations that are mentioned in the notification instead of all of them, otherwise for people with hundreds of conversations, it would be quite slow (you will have to fetch them from the fault state just to check that they are still the same as before).

mikeger: @marcoconti83 As I mentioned, I need to observe the transient properties that are not making the object included in the changeset. 

`lastConversationsState` does not contain all conversations, it only contain ones that we seen once changing the state in `managedObjectsDidChange:`.

Unfortunately it just does not work another way or I don't see another way of achieving observation of the transient properties like `callTimedOut`.

You can check the test `testThatItRequestsEndCall_Timeout`, if I would look into didSaveNotification changeset, there would be nothing because CoreData does not keep track of the values of the transient properties.

marcoconti83: Sorry, I missed the part about transient properties. Then I agree this is probably the only way to do this.

PR #130, opened 2016-11-21
User profile update refactoring

marcoconti83:
# Part 1 of multiple PRs

`ZMUserProfileUpdateStatus` was based on a single enumeration that was keeping track of the status of multiple operations. As we are adding even more operations (that could, in theory, be performed in parallel), it's better to have separate tracking for each operation.

This PR has nothing to do with usernames yet but they will be added in the same status/transcoder combo.

# Changes
Track status of individual operations in `UserProfileUpdateStatus`
Also rewrote in Swift the status and the transcoder.


Source/Synchronization/Strategies/UserProfileUpdateRequestStrategy.swift
```diff
@@ -0,0 +1,158 @@
+//
+//  UserProfileUpdateRequestStrategy.swift
+//  zmessaging-cocoa
+//
+//  Created by Marco Conti on 15/11/16.
```
daehn: Missing copyright header

Source/UserSession/UserProfileUpdateStatus.swift
```diff
@@ -0,0 +1,279 @@
...
+    /// phone number to validate
+    fileprivate var synchingPhoneNumberForValidationCode : SyncToBackendPhase<String> = .idle
+    
+    /// last set password and email
+    fileprivate var lastEmailAndPassword : ZMEmailCredentials?
+    
+    let managedObjectContext : NSManagedObjectContext
+    
+    /// Callback invoked when there is a new request to send
+    let newRequestCallback : ()->()
```
typfel: Requiring the user of this class pass in this block on `init` makes it a bit harder to use since he/she must know it should contain a `RequestAvailableNotification.notifyNewRequestsAvailable(nil)`

marcoconti83: Good point

marcoconti83:
@typfel please re-review

PR #133, opened 2016-11-22
Replace TypingTranscoder with TypingStrategy

basine:
Swift for the win!


Source/Synchronization/Strategies/TypingStrategy.swift
```diff
@@ -0,0 +1,192 @@
...
+            }
+        }
+        
+    }
+}
+
+extension TypingStrategy : ZMRequestGenerator {
+    
+    public func nextRequest() -> ZMTransportRequest? {
+        guard clientRegistrationDelegate.clientIsReadyForRequests,
```
marcoconti83: Does `clientIsReadyForRequest` also check if we are logged in (normally done from the authentication status)?

marcoconti83:
image

Source/Synchronization/Strategies/TypingStrategy.swift
```diff
@@ -0,0 +1,192 @@
...
+              let user = ZMUser(remoteID: userID, createIfNeeded: true, in: syncContext),
+              let conversation = conversationsByID?[conversationID] ?? ZMConversation(remoteID: conversationID, createIfNeeded: true, in: syncContext)
+        else { return }
+        
+        if event.type == .conversationTyping {
+            guard let payloadData = event.payload["data"] as? [String: String],
+                  let status = payloadData[StatusKey]
+            else { return }
+            processIsTypingUpdateEvent(for: user, in: conversation, with: status)
+        } else if event.type == .conversationOtrMessageAdd {
```
daehn: This will also be triggered for delivery receipts which might be undesired. In order to know we would need to create a generic message from the update event (which is already decrypted so it should not be that much overhead to create it here as well I guess)

Source/Synchronization/Strategies/TypingStrategy.swift
basine: Yes, basically it checks if the clientRegistrationStatus is in ZMClientRegistrationPhaseRegistered which is only the case when we are already logged in.

Source/Synchronization/Strategies/TypingStrategy.swift
marcoconti83: @daehn @basine  so is this an issue that we want to fix? If yes, Silvan you should "request changes" in your review, I was about to merge this.

daehn: This is not a regression and did behave in the same way before, so I don't think we necessarily need to fix this in this PR, if we don't then I will create a ticket to track it.

marcoconti83: Cool, please create ticket then, thanks!

PR #140, opened 2016-11-24
Feature/usernames set handle

marcoconti83:
We want to set the handle from the client side

# Changes
- Add ability to set handle, and callbacks for when it fails or succeeds
- Remove some redundant enumerations from tests


Tests/Source/Integration/UserHandleTests.swift
```diff
@@ -86,4 +86,59 @@ class UserHandleTests : IntegrationTestBase {
...
+        
+        let selfUser = ZMUser.selfUser(inUserSession: self.userSession)!
+        XCTAssertEqual(selfUser.handle, handle)
+        
+        self.mockTransportSession.performRemoteChanges { _ in
+            XCTAssertEqual(self.selfUser.handle, handle)
+        }
+    }
+    
+    func testThatItIsNotifiedWhenFailsToSetTheHandle() {
```
daehn: The name suggests that the invoked callback should be `.didFailToSetHandle`, it looks like the test checks the same as `testThatItCanSetTheHandle`

zmessaging-cocoa.xcodeproj/xcshareddata/xcschemes/zmessaging-ios.xcscheme
```diff
@@ -114,7 +114,7 @@
          <EnvironmentVariable
             key = "ZMLOG_TAGS"
             value = "MockTransportRequests"
-            isEnabled = "NO">
+            isEnabled = "YES">
```
daehn: Can be removed

marcoconti83:
Fixed!

PR #143, opened 2016-11-25
Feature/usernames random generator

marcoconti83:
When the user does not have a handle defined, we want to suggest one that is derived from her user name. If this is not possible, we will use a randomly generated word.

# Known issues
The list of words is not finalized


Source/Registration/RandomHandleGenerator.swift
```diff
@@ -0,0 +1,170 @@
+//
+//  RandomHandleGenerator.swift
+//  zmessaging-cocoa
+//
```
daehn: Missing copyright header, I will check why travis didn't pick this up.

Source/Registration/RandomHandleGenerator.swift
```diff
@@ -0,0 +1,170 @@
...
+        let mutableString = NSMutableString(string: self) as CFMutableString
+        for transform in [kCFStringTransformToLatin, kCFStringTransformStripDiacritics, kCFStringTransformStripCombiningMarks] {
+            CFStringTransform(mutableString, nil, transform, false)
+        }
+        return String(mutableString)
+    }
+    
+    /// returns self only with alphanumeric and underscore
+    fileprivate var onlyAlphanumericWithUnderscore : String {
+        let allowedCharacters = CharacterSet(charactersIn: "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890_")
```
daehn: AFAIK only lowercase characters should be allowed

Tests/Source/Registration/RandomHandleGeneratorTests.swift
```diff
@@ -0,0 +1,92 @@
+//
+//  RandomHandleGeneratorTests.swift
+//  zmessaging-cocoa
+//
+//  Created by Marco Conti on 24/11/16.
```
daehn: Missing copyright headers

Source/Registration/RandomHandleGenerator.swift
daehn: Somehow travis did not run for this particular PR, maybe because it's not a PR to develop but to another feature branch

marcoconti83: Yeah it's because it's not to develop, I've seen it before

Source/Registration/RandomHandleGenerator.swift
marcoconti83: ...but we want to turn them lowercase, not delete them. So I leave them in at this stage, and then call lowercase later.

PR #146, opened 2016-11-28
Update random1.txt

Yorgos-V:


Resources/random1.txt
```diff
@@ -53,7 +50,7 @@ lemon
 light
 lilac
 lime
-lime
+lord
```
daehn: I would argue that's too religious

marcoconti83: I think he wanted to auto generate `lordtilapia`

Yorgos-V: That was my intention :)
I checked the list with @infotexture , he found it ok

PR #152, opened 2016-12-02
[Incremental PR] Address book matching

marcoconti83:
We want to handle address book access with `AddressBook` framework (iOS 8) and `Contacts` framework (iOS 9+) where available. 

`Contacts` has some advantages (other than not being deprecated: `AddressBook` is deprecated by Apple). This allows us to run a search query using the appropriate API which is hopefully better performing that manually iterating through the entire AB. For `Addressbook` this functionality is not present (and the method just returns an empty list).

We considered removing support for `AddressBook` completely but this would mean that iOS 8 users will not be able to match with anyone using AB upload, which is undesirable. 

# Changes
- Create separate `AddressBookIOS8` and `AddressBookIOS9` implementations of address book access, with a common superclass.
- Use the `AddressBookEntry` property of `ZMUser` as source of link between a Wire user and its address book contact

# Testing
With the current testing infrastructure it's not possible to test actual address book access (can't automatically accept "allow access" system popup)

# Known issues
- Big chunks of testing are still missing, pending `ZMSearchDirectory` refactoring
- Could not find a way to implement`numberOfContacts` with the `Contacts` framework (other than iterating through all of them)
- The `AddressBookEntry` is still not populated during AB upload.


daehn:
@marcoconti83 if we want to try testing this and need address book permissions we could give https://github.com/plu/JPSimulatorHacks another try.

marcoconti83:
@daehn worth a try. Will look at it once we are done with the feature.

Source/UserSession/Search/ZMSearchResult+AddressBook.swift
```diff
@@ -17,41 +17,74 @@
...
     
     /// Creates a new search result with the same results and additional
-    /// results obtained by searching through the address book
+    /// results obtained by searching through the address book with the same query
     public func extendWithContactsFromAddressBook(_ query: String,
-                                                  usersToMatch: [ZMUser],
                                                   userSession: ZMUserSession) -> ZMSearchResult {
-        
+        /*
+         When I have a search result obtained (either with a local search of from the BE) by matching on Wire
```
daehn: Typo `of from the BE` → `or from the BE`

Source/Registration/AddressBookIOS8.swift
```diff
@@ -0,0 +1,148 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import AddressBook
+import libPhoneNumber
+
+/// iOS AddressBook-based address book
+class AddressBookIOS8 : AddressBook {
```
mikeger: So it was decided to keep iOS 8 support for this feature?

Tests/Source/Synchronization/Strategies/AddressBookUploadRequestStrategyTest.swift
```diff
@@ -209,16 +207,17 @@ extension AddressBookUploadRequestStrategyTest {
         XCTAssertNotNil(request1)
         XCTAssertNotNil(request2)
         if let cards1 = (request1?.payload as? [String:AnyObject])?["cards"] as? [[String:AnyObject]] {
+            print(cards1)
```
daehn: `print` should be removed

Source/Registration/AddressBookIOS8.swift
marcoconti83: Yes, see PR description:

> We considered removing support for AddressBook completely but this would mean that iOS 8 users will not be able to match with anyone using AB upload, which is undesirable.

We will not support searching in AB on iOS 8, but we still need to use AB for automatching even on iOS 8

PR #154, opened 2016-12-05
Extract top people search to separate class

marcoconti83:
`ZMSearchDirectory` does too many things

# Changes
Extract top conversations retrieval to a separate class

# Known issues
- The parameters list for some `init` such as `ZMSyncStrategy` is getting ridiculously long. We discussed it and we will refactor it soon.
- The observation is gone. Will be done in a separate PR.


Source/UserSession/Search/TopConversationsDirectory.swift
```diff
@@ -0,0 +1,78 @@
...
+        self.managedObjectContext.setPersistentStoreMetadata(valueToSave, forKey: topConversationsObjectIDKey)
+    }
+
+    /// Load list from persistent store
+    fileprivate func loadList() {
+        guard let ids = self.managedObjectContext.persistentStoreMetadata(forKey: topConversationsObjectIDKey) as? [URL] else {
+            return
+        }
+        let managedObjectIDs = ids.flatMap { self.managedObjectContext.persistentStoreCoordinator?.managedObjectID(forURIRepresentation: $0) }
+        self.topConversationsCache = managedObjectIDs.flatMap { self.managedObjectContext.object(with: $0) as? ZMConversation }
```
daehn: Are we sure the object IDs are guaranteed to still exist?  As far as I can recall `objectWithID` will always return an object even if it does not exist and then in turn throw an execution when accessing a property of such object. In other places we use `existingObjectWithID` to avoid this.

daehn: From the documentation about `object(with:)`:

> If the object is not registered in the context, it may be fetched or returned as a fault. This method always returns an object. The data in the persistent store represented by objectID is assumed to exist—if it does not, the returned object throws an exception when you access any property (that is, when the fault is fired). The benefit of this behavior is that it allows you to create and use faults, then create the underlying data later or in a separate context.

marcoconti83: Nice catch, but yeah we should be good because of that method that fetches the object ID

PR #155, opened 2016-12-06
Add tracking to handle suggestion including number of requests

daehn:
# What's in this PR?

* This adds tracking to the handle suggestion logic, specifically we are interested in the number of attempts that have been needed until an available handle has been found.
* The tracking logic mostly lives in a separate struct in order to keep the `UserProfileUpdateStatus` clean from too much analytics related state and code.


Source/UserSession/UserProfileUpdateStatus.swift
```diff
@@ -51,15 +51,27 @@ private let alternativeAutogeneratedNames = 4
     
     /// best handle suggestion found so far
     fileprivate(set) public var bestHandleSuggestion : String?
+
+    /// Analytics tracker used to track th enumber of attempts until a handle suggestion has been found
```
daehn: Typo 😱

PR #158, opened 2016-12-07
Remove sync states

basine:
**Motivation for this PR**

We currently have a state machine consisting of states that reflect the state of the app which goes from unauthenticated > authenticated > syncing > event processing (normal) state ( > background states). 
Each state manages a subset of all (remaining) transcoders and defines when each transcoder is allowed to make requests and what happens after a request of a specific transcoder finishes, as well as when a state transitions to the next. The "knowledge" of when a transcoder is active therefore does not lie in the transcoder itself, but in the state machine.

In the last months we have been moving from the concept of transcoders - the old solution for syncing object changes  in which request activity is steered externally by state machines - to 
syncStrategies where request activity is steered internally, by several status reflecting the app state.

This PR is removing the following states and combines them in one `SyncStatus`
`ZMLastUpdateEventIDPhase`
`ZMEventCatchUpPhaseOne`
`ZMEventCatchUpPhaseTwo`
`ZMSlowSyncPhaseOne`
`ZMSlowSyncPhaseTwo`

By doing that the contained transcoders have turned into strategies as well.

**ToDo in comming PRs**
- Rename the transcoders to strategies (and add more tests)
- Other strategies should not make requests during the sync
- By removing the slow and quick sync states, the state machine immediately enters the event processing state after authentication. Requests however need to be blocked until the sync completes. We should move these transcoders to strategies as well and get rid of the eventProcessing state.


Source/Synchronization/Sync States/ZMSyncStateMachine.m
```diff
@@ -179,15 +158,10 @@ - (void)prepareForSuspendedState;
...
-{
-    [self goToState:self.updateEventsCatchUpPhaseOneState];
-}
-
-- (void)startSlowSync
-{
-    [self goToState:self.downloadLastUpdateEventIDState];
-}
+//- (void)startQuickSync
+//{
```
daehn: Can this be removed?

Source/Synchronization/Transcoders/ZMConversationTranscoder.m
```diff
@@ -160,6 +168,33 @@ - (BOOL)isSlowSyncDone
...
+{
+    if (!self.clientRegistrationDelegate.clientIsReadyForRequests) {
+        return nil;
+    }
+    
+    SyncStatus *status = self.syncStatus;
+    if (status.currentSyncPhase == self.expectedSyncPhase) {
+        if (!self.didStartSlowSync) {
+            [self setNeedsSlowSync];
+            self.didStartSlowSync = YES;
```
basine: In this case unfortunately not: `setNeedsSlowSync` resets the listPaginator which fetches all remoteIDs of all conversations and it also resets the remoteIDSync which is "filled" with IDs of conversations to fetch when the listPaginator finishes a page. While the listPaginator uses a ZMSingleRequestSync which has an internal status, the remoteIDSync does not have any state information (it would return isDone == true when it hasn't received any IDs to fetch yet as well as when it has finished fetching IDs). I'd prefer leaving it like that for this PR and review it later.

Source/Synchronization/ZMSyncStrategy.m
```diff
@@ -304,12 +320,18 @@ - (NSManagedObjectContext *)moc
...
 
 - (void)didInterruptUpdateEventsStream
 {
-    [self.stateMachine didInterruptUpdateEventsStream];
+//    if (self.stateMachine.currentState != self.stateMachine.backgroundFetchState
+//        && self.stateMachine.currentState != self.stateMachine.backgroundState
+//        && self.stateMachine.currentState != self.stateMachine.preBackgroundState)
+//    {
+//        // TODO Sabine how should the background states be handled?
+        [self.syncStatus pushChannelDidClose];
```
basine: That's already handled in the SyncStatus class - this comment was removed in later PRs

Source/Synchronization/Sync States/ZMSyncStateMachine.m
basine: A lot of these comments are deleted in later PRs

Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
```diff
@@ -270,17 +297,30 @@ @implementation ZMMissingUpdateEventsTranscoder (Pagination)
...
-    
+    SyncStatus *syncStatus = self.syncStatus;
     NSUUID *latestEventId = [self processUpdateEventsAndReturnLastNotificationIDFromPayload:response.payload syncStrategy:self.syncStrategy];
     if (latestEventId != nil) {
-        self.lastUpdateEventID = latestEventId;
+        if (response.HTTPStatus == 404 && syncStatus.currentSyncPhase == SyncPhaseFetchingMissedEvents) {
+            // If we fail during quick sync we need to re-enter slow sync and should not store the lastUpdateEventID until after the slowSync has been completed
+            // Otherwise, if the device crashes or is restarted during slow sync, we lose the information that we need to perform a slow sync
+            [syncStatus updateLastUpdateEventIDWithEventID:latestEventId];
+            // TODO Sabine: What happens when we receive a 404 when we are fetching the notification for a push notification? In theory we would have to enter slow sync as well or at least not store the lastUpdateEventID until the next proper sync in the foreground
```
basine: This would be a separate PR, noted as a TODO

PR #159, opened 2016-12-07
Add observer to top conversations

daehn:
# What's in this PR?

* Add the possibility the subscribe to changes to the top conversations using a `TopConversationsDirectoryObserver`.


Source/UserSession/Search/TopConversationsDirectory.swift
```diff
@@ -65,6 +65,7 @@ extension TopConversationsDirectory {
     private func persistList() {
         let valueToSave = self.topConversations.map { $0.objectID.uriRepresentation() }
         self.managedObjectContext.setPersistentStoreMetadata(valueToSave, forKey: topConversationsObjectIDKey)
+        TopConversationsDirectoryNotification.post()
```
daehn: The notification is not fired on the main thread, but `topConversationsDidChange` on observers is being called on the main thread.

PR #162, opened 2016-12-07
Use new backend endpoint for checking handles availability

marcoconti83:
There is now a better endpoint on the backend to test for handles availability


Source/Synchronization/Strategies/UserProfileUpdateRequestStrategy.swift
```diff
@@ -146,8 +146,11 @@ extension UserProfileRequestStrategy : ZMSingleRequestTranscoder {
             return ZMTransportRequest(path: "/self/handle", method: .methodPUT, payload: payload)
             
         case self.handleSuggestionSearchSync:
-            let handles = (self.userProfileUpdateStatus.suggestedHandlesToCheck ?? []).joined(separator: ",")
-            return ZMTransportRequest(path: "/users?handles=\(handles)", method: .methodGET, payload: nil)
+            let payload = [
+                    "handles" : self.userProfileUpdateStatus.suggestedHandlesToCheck ?? [],
```
daehn: Should we actually do a request if the array of suggestions is not there, I assume if this happens that something went wrong.

marcoconti83: True, will assert

marcoconti83: It's too late not to send a request at this point

marcoconti83:
@daehn  Fixed

PR #168, opened 2016-12-08
Filter search results when searching by handle

marcoconti83:
Filter user and group conversations from search result, when the query is a query for a username (`@query`)


Source/UserSession/Search/ZMSearchDirectory.m
```diff
@@ -207,7 +207,7 @@ - (ZMSearchToken)searchForUsersAndConversationsMatchingQueryString:(NSString *)q
     request.includeDirectory = YES;
     request.includeGroupConversations = YES;
     request.includeRemoteResults = YES;
-    request.includeAddressBookContacts = YES;
+    request.includeAddressBookContacts = NO;
```
marcoconti83: This reverts the previous change that include the AB results in the query, which we will do later (lacking UI support for this at the moment)

PR #170, opened 2016-12-09
Fetch *some* contacts when searching the AB for an empty query

marcoconti83:
When searching with an empty query in the AB, we would get an empty result. We should have gotten the entire AB.

# Changes
Returns a part of the AB (max 2000 contacts)

# Testing
We don't yet have a system in place to test address book searches

### DIFF
--- a/Source/Registration/AddressBookIOS8.swift
+++ b/Source/Registration/AddressBookIOS8.swift
@@ -73,7 +73,7 @@ extension AddressBookIOS8 : AddressBookAccessor {
             guard !matchingQuery.isEmpty else {
                 return true
             }
-            return $0.displayName.lowercased().contains(matchingQuery)
+            return $0.displayName.range(of: matchingQuery, options: .caseInsensitive) != .none
         }
     }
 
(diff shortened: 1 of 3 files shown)


Source/Registration/AddressBookIOS8.swift
```diff
@@ -68,7 +69,12 @@ extension AddressBookIOS8 : AddressBookAccessor {
     
     /// Returns contacts matching search query
     func rawContacts(matchingQuery: String) -> [ContactRecord] {
-        return []
+        return self.firstRawContacts(number: addressBookContactsSearchLimit).filter {
+            guard !matchingQuery.isEmpty else {
+                return true
+            }
+            return $0.displayName.lowercased().contains(matchingQuery)
```
mikeger: Can be done with `.range(of: filterString, options: .caseInsensitive) != .none`, I guess it would not create a new lowercased instance.

marcoconti83: Done

PR #171, opened 2016-12-09
Do not filtering out conversation with @ in name

marcoconti83:
When filtering out conversation from a search result when the query was a username search (with starting with `@`), conversations that have `@` in the name should still be included

### DIFF
--- a/Source/UserSession/Search/Internal/ZMSearch.m
+++ b/Source/UserSession/Search/Internal/ZMSearch.m
@@ -155,15 +155,17 @@
 
 - (NSArray *)conversationsMatchingSearchString:(NSString *)searchString
 {
-    if([searchString hasPrefix:@"@"]) { // ignore group conversations if query is for usernames
-        return @[];
-    }
     NSFetchRequest *conversationFetchRequest = [ZMConversation sortedFetchRequestWithPredicate:[ZMConversation predicateForSearchString:searchString]];
-    
     NSSortDescriptor *sortDescriptor = [NSSortDescriptor sortDescriptorWithKey:ZMNormalizedUserDefinedNameKey ascending:YES];
     conversationFetchRequest.sortDescriptors = @[sortDescriptor];
     
     NSArray *conversationResults = [self.searchContext executeFetchRequestOrAssert:conversationFetchRequest];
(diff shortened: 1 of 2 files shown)


Source/UserSession/Search/Internal/ZMSearch.m
```diff
@@ -155,15 +155,17 @@ - (NSArray *)connectedUsersMatchingSearchString:(NSString *)searchString
 
 - (NSArray *)conversationsMatchingSearchString:(NSString *)searchString
 {
-    if([searchString hasPrefix:@"@"]) { // ignore group conversations if query is for usernames
-        return @[];
-    }
     NSFetchRequest *conversationFetchRequest = [ZMConversation sortedFetchRequestWithPredicate:[ZMConversation predicateForSearchString:searchString]];
```
mikeger: I think we can use different predicate in case search contains @: do not search in participant's names then. Would that eliminate the need in re-filtering the results?

daehn: We still want to include names that start with an @ / match the query

mikeger: @daehn right now we have a predicate `predicateForSearchString` to match either conversation or participant name. If query string contains '@' then if we match with this predicate, one or another (conv name or participant name) should have @. Then we don't need to extra filter the query result, as I understand?

marcoconti83: True but this predicate is set in data-model, and I did not want to put the logic of "if the user searched with a "@", then ..." there.

mikeger: Right, may be `predicateForSearchString` can have a parameter like `matchUserNames:Bool`, because when it matches usernames it does a subquery, would be cool to avoid doing the subquery and then filtering on top to remove the subquery effect.

marcoconti83: I would merge this in, then we can do that change in another PR as it requires changes in data model and the in-between frameworks

marcoconti83: Turns out we already have such a predicate, so no changes needed in data model. Will do another PR on SE.

PR #172, opened 2016-12-09
Do not abort AB processing on invalid entry

marcoconti83:
The AB iteration was aborting in case of invalid entry (no valid email or password). It should continue to the next contact instead.

### DIFF
--- a/Source/Registration/AddressBook.swift
+++ b/Source/Registration/AddressBook.swift
@@ -47,7 +47,7 @@ extension AddressBookAccessor {
     
     /// Enumerates the contacts, normalized and validated, invoking the block for each contact.
     /// Non valid contacts (no email nor phone) will be excluded from the enumeration.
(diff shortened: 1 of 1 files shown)


Source/Registration/AddressBook.swift
```diff
@@ -51,7 +51,7 @@ extension AddressBookAccessor {
     func enumerateValidContacts(block: @escaping (ZMAddressBookContact)->(Bool)) {
         self.enumerateRawContacts {
             guard let parsed = ZMAddressBookContact(contact: $0, phoneNumberNormalizer: self.phoneNumberNormalizer) else {
-                return false
+                return true
```
typfel: ` If the block returns true, it will stop enumerating them.`  this should be:  ` If the block returns false, it will stop enumerating them.` ?

marcoconti83:
@typfel fixed

PR #174, opened 2016-12-12
Fix test crashes / bugs

basine:
**Motivation for this PR**
1) I was having issues with tests crashing in tearDown. After further investigation it was most probably due to an object being allocated with an managedObjectContext that was being released at the same time. This in turn was due to an improper tearDown method that first created the object to be tornDown if the object didn't exist. Since most objects we store in the userInfo of the MOC respond to the `tearDown` method, we are now looping through the values of the dictionary instead.

2) Another crash happened for me when running the GSMHandler tests due to the fact the we never tore down the SUT which was registered as an observer with NSNotificationCenter.

3) I was having issues with very slow TypingStrategy tests which I suppose was due to a missing call the the OperationLoop to ask sync objects for next requests.


Tests/Source/UserSession/ZMUserSessionTestsBase.m
```diff
@@ -106,15 +108,40 @@ - (void)setUp
...
 {
+    [self tearDownUserInfoObjectsOfMOC:self.syncMOC];
+    [self.syncMOC.userInfo removeAllObjects];
+    
+    [self tearDownUserInfoObjectsOfMOC:self.uiMOC];
+    [self.uiMOC.userInfo removeAllObjects];
+    
     [super cleanUpAndVerify];
     [self.sut.authenticationStatus removeAuthenticationCenterObserver:self];
+    self.authFailHandler = nil;
```
typfel: Why do we need to nil out every property? If things are not getting deallocated we should make sure we don't have reference cycle.

basine: That's from trying to find the root cause of the crash (which was not that one).

daehn:
🔥 Conflicts

daehn:
@basine @typfel I resolved the conflict, can this be merged?

marcoconti83:
Fixed @typfel's comment on separate commit

Tests/Source/UserSession/ZMUserSessionTestsBase.m
marcoconti83: Actually this is a `tearDown`. The class instance will be reused for all tests. If a test sets all the properties again in the `setUp`, then all is good. But if one property is not set, we will keep the old one from previous test around, which might (e.g. if register as observer) still create some interference with current test. 
I don't like the `setUp`-`tearDown` model (or at least, how we use it) for this reason. But anyway, to be on the safe side, nullifying all properties is not a bad idea.

PR #178, opened 2016-12-13
Add hotfix marking all connected users to be updated from the backend

daehn:
# What's in this PR?

* Users updating from a non-username version might have missed the updates from users that previously set a username, which is why we now refetch all connected users in a hotfix.
* A refetch of unconnected users should already be initiated by the UI with a call to `refreshData`.


typfel:
😮

PR #183, opened 2016-12-14
Abstract pending hot fix changes into variable on user session

daehn:
# What's in this PR?

* As mentioned in https://github.com/wireapp/wire-ios/pull/482 this hides the underlying conditions that might cause a hot fix to be pending from the UI.


daehn:
@marcoconti83 I added tests, please re-review

PR #188, opened 2016-12-19
Move out fetching client strategy

Einh06:
**What's in this PR**
This PR removes fetching of client from `UserClientRequestStrategy`

- [x] Requires release of `wire-ios-message-strategy`


Source/Synchronization/Strategies/UserClientRequestStrategy.swift
```diff
@@ -134,7 +115,7 @@ public final class UserClientRequestStrategy: ZMObjectSyncStrategy, ZMContextCha
             return request
```
marcoconti83: This could be even simplified by removing the last if, and just having
`return modifiedSync.nextRequest()`

PR #191, opened 2016-12-21
Fix/core data concurrency

marcoconti83:
Fix core data concurrency issues and adopt new ZMTransport API (also modified for concurrency issues)

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/113


Cartfile.resolved
```diff
@@ -9,7 +9,7 @@ github "wireapp/wire-ios-link-preview" "4.0.2"
 github "wireapp/wire-ios-utilities" "13.0.1"
 github "wireapp/wire-ios-images" "7.0.0"
 github "wireapp/wire-ios-testing" "8.0.1"
-github "wireapp/wire-ios-transport" "16.0.0"
+github "wireapp/wire-ios-transport" "fea6ddb437dd8da10c34e6fdbb6f94ab28040164"
```
typfel: Need new version

marcoconti83: Need to bump everything actually, starting with this: 
https://github.com/wireapp/wire-ios-data-model/pull/113

Source/Calling/ZMGSMCallHandler.m
```diff
@@ -179,22 +179,22 @@ - (void)rejoinVoiceChannelAfterGSMInterruption
 
 - (BOOL)hasStoredInterruptedCallConversation
 {
-    return ([self.uiManagedObjectContext persistentStoreMetadataForKey:ZMInterruptedCallConversationObjectIDKey] != nil);
+    return ([self.syncManagedObjectContext persistentStoreMetadataForKey:ZMInterruptedCallConversationObjectIDKey] != nil);
```
typfel: So this should only be called from the sync context then? Which it's not I think.

marcoconti83: I tried calling (outgoing/incoming) and I never hit the crash for CD concurrency here, so I guess it works.

marcoconti83: ... as opposed to the other places where I had to make changes, those were making CD crash

marcoconti83: Good point. Did not try that.

marcoconti83:
@typfel please re-review

marcoconti83:
I tested it and found issues in the code you were suspicious about :)
Had to add a lot of context switches but now it works

typfel:
Nice

PR #193, opened 2016-12-23
Fix concurrency issues

marcoconti83:
Concurrency issues on login and GSM call

### DIFF
--- a/Source/Calling/ZMGSMCallHandler.m
+++ b/Source/Calling/ZMGSMCallHandler.m
@@ -36,7 +36,6 @@ NSString *const ZMInterruptedCallConversationObjectIDKey = @"InterruptedCallConv
 @property (nonatomic) CTCallCenter *callCenter;
 @property (nonatomic) ZMCallStateLogger *callStateLogger;
 
-@property (nonatomic) ZMConversation *storedConversation;
 @property (nonatomic, readonly) ZMConversation *activeCallUIConversation;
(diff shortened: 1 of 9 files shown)


Source/Calling/ZMGSMCallHandler.m
```diff
@@ -36,7 +36,6 @@ @interface ZMGSMCallHandler ()
 @property (nonatomic) CTCallCenter *callCenter;
 @property (nonatomic) ZMCallStateLogger *callStateLogger;
 
-@property (nonatomic) ZMConversation *storedConversation;
```
typfel: Why was this a concurrency issue? You can't query an object from another context?

marcoconti83: No, one context one thread.
Actually this was a bug introduced with the previous PR. This used to be a computed property, I changed the method name without changing the property name, so it was actually storing something there in those code paths where I missed replacing the old name with the new name.

Source/UserSession/ZMAuthenticationStatus.m
```diff
@@ -420,5 +420,11 @@ - (ZMEmailCredentials *)emailCredentials
     return nil;
 }
 
+- (BOOL)registeredOnThisDeviceOnContext:(NSManagedObjectContext *)moc;
+
```
typfel: Blank line

PR #195, opened 2016-12-28
Fix/persistent metadata

marcoconti83:
New API for persistent store metadata in Data Model

# Changes
- adopt new API for persistent store metadata
- fix all failing tests

# Testing
This PR also fix failing tests on DEV


Source/Notifications/Push notifications/Helpers/SessionTracker.swift
```diff
@@ -204,7 +204,7 @@ public final class Session : NSObject, NSCoding, NSCopying {
     
     /// unarchives previous calls that haven't been cancelled yet
     func unarchiveOldSessions(){
-        guard let archive = managedObjectContext.storedValue(key: SessionTracker.ArchivingKey) as? Data,
+        guard let archive = managedObjectContext.persistentStoreMetadata(key: SessionTracker.ArchivingKey) as? Data,
```
Einh06: Why aren't we using the new API here?

Source/Notifications/Push notifications/Helpers/SessionTracker.swift
```diff
@@ -213,7 +213,7 @@ public final class Session : NSObject, NSCoding, NSCopying {
     /// Archives sessions
     func updateArchive(){
         let data = NSKeyedArchiver.archivedData(withRootObject: sessions)
-        managedObjectContext.store(value: data, key: SessionTracker.ArchivingKey)
+        managedObjectContext.setPersistentStoreMetadata(data, key: SessionTracker.ArchivingKey)
```
Einh06: Same as above

Source/Notifications/Push notifications/Helpers/SessionTracker.swift
marcoconti83: The new API is `persistentStoreMetadata` (used to be the same name even before, but the getter would accept `Any`, now it only accepts specific types that are guaranteed to be safe to persist). Somewhere else we use `store` because `store` is part of the `KeyValueStore` protocol, that we use to hide away the fact that this is a `NSManagedObjectContext` in those places where we don't want to pass a managed object context but something more abstract.
In all the other places where we know that this is a `NSManagedObjectContext`, we could use either. I decided to use `persistentStoreMetadata` everywhere we know that this is a managed object context, and `store` where we only know that it is a `KeyValueStore`.

Einh06: Ok, sounds good to me then

marcoconti83:
There are 2 APIs, one is defined at the level of `NSManagedObjectContext` (`persistentMetadata:`), and one is an abstract protocol `KeyValueStore` (`store:`). They both use the same level of safety (strongly typed), and `KeyValueStore` is just a wrapper around `NSManagedObjectContext` API.

PR #197, opened 2017-01-03
Reset audio device when ending a callkit call

typfel:
As requested by AVS we need to call `resetAudioDevice` when a Callkit call is over in order to restore the audio session. This should fix a bug where a ringtone would not play for the second call.

https://wearezeta.atlassian.net/browse/ZIOS-7823


typfel:
@basine yep I agree.  But need to re-factor the class to support testing this, which I would prefer to do on the calling v3 branch.

PR #198, opened 2017-01-03
Move statuses into ZMSyncStrategy

basine:
**In this PR**
ZMClientregistrationStatus, ZMAuthenticationStatus et al used to be initialized by ZMUserSession. We are now intializing them in ZMSyncStrategy instead.

Plus some clean-up of remaining previous PRs.

rebased copy of https://github.com/wireapp/wire-ios-sync-engine/pull/179

### DIFF
--- a/Tests/Source/Synchronization/ZMSyncStrategyTests.m
+++ b/Tests/Source/Synchronization/ZMSyncStrategyTests.m
@@ -26,39 +26,49 @@
 
 #import "MessagingTest.h"
 #import "ZMUserSession+Internal.h"
-#import "ZMConnectionTranscoder.h"
 #import "ZMSyncStrategy+Internal.h"
-#import "ZMUserTranscoder.h"
+#import "ZMSyncStrategy+EventProcessing.h"
+#import "ZMSyncStrategy+ManagedObjectChanges.h"
+#import "ZMUpdateEventsBuffer.h"
+#import "ZMOperationLoop.h"
+#import "AVSMediaManager.h"
+#import "AVSFlowManager.h"
+#import "MessagingTest+EventFactory.h"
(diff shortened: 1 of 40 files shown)


Tests/Source/Synchronization/ZMSyncStrategyTests.m
```diff
@@ -89,10 +99,30 @@ - (void)setUp
     [self verifyMockLater:self.mockUpstreamSync1];
     [self verifyMockLater:self.mockUpstreamSync2];
     
-    self.authenticationStatus = [[ZMAuthenticationStatus alloc] initWithManagedObjectContext:self.syncMOC cookie:nil];
-    self.userProfileUpdateStatus = [[UserProfileUpdateStatus alloc] initWithManagedObjectContext:self.syncMOC];
-    self.clientRegistrationStatus = [[ZMClientRegistrationStatus alloc] initWithManagedObjectContext:self.syncMOC loginCredentialProvider:self.authenticationStatus updateCredentialProvider:self.userProfileUpdateStatus cookie:nil registrationStatusDelegate:nil];
-    self.clientUpdateStatus = [[ClientUpdateStatus alloc] initWithSyncManagedObjectContext:self.syncMOC];
+    self.syncStateDelegate = [OCMockObject niceMockForProtocol:@protocol(ZMSyncStateDelegate)];
+    
+    self.slowSynStatus = [OCMockObject mockForClass:SyncStatus.class];
```
marcoconti83: Why are we moving from using real objects to mocks? Is there any advantage in doing this?
Everything else being equal, using real objects is better as we don't risk breaking random tests later if we forget to stop mocking `alloc` (which will also happens if a test raise an exception for example, the `tearDown` is not invoked).

PR #201, opened 2017-01-11
Add calling related classes from data model

typfel:
## What changed
The calling code has been moved from the data model and the calling observers
has been re-written and placed in `WireCallCenterV2`. `WireCallCenter` has been 
renamed to `WireCallCenterV3`.

To provide a single interface to the UI a new `WireCallCenter` has also been
created which wraps the observers in V2 and V3.

All tests have been updated to the new API except for the `CallKitDelegate`
which still need needs some testing to see if the current approach works.

## What's missing
- tests for `WireCallCenterV2` and `WireCallCenter`
- tests for V3 integration with `ZMCallkitDelegate`,
- tests for V2 integration with `ZMCallkitDelegate` are broken


Source/Calling/VoiceChannelV3.swift
```diff
@@ -0,0 +1,95 @@
...
+    
+    var connectionState : ZMVoiceChannelConnectionState {
+        switch self {
+        case .unknown:
+            fallthrough
+        case .terminating:
+            fallthrough
+        case .incoming:
+            fallthrough
+        case .none:
```
vytis: I guess you could do `case .unknown, .terminating, .incoming, .none` here

Source/Calling/VoiceChannelRouter.swift
```diff
@@ -0,0 +1,64 @@
+//
+//  VoiceChannelRouter.swift
+//  ZMCDataModel
+//
+//  Created by Jacob on 23/11/16.
```
daehn: Missing copyright header

Source/Calling/VoiceChannelRouter.swift
```diff
@@ -0,0 +1,64 @@
...
+public class VoiceChannelRouter : NSObject, VoiceChannel {
+    
+    public static var isCallingV3Enabled : Bool = false
+    
+    public let v3 : VoiceChannelV3
+    public let v2 : ZMVoiceChannel
+    
+    public init(conversation: ZMConversation) {
+        v3 = VoiceChannelV3(conversation: conversation)
+        v2 = ZMVoiceChannel(conversation: conversation)
```
daehn: I assume `ZMVoiceChannel` has been there before, but maybe it would make sense to rename it to `VoiceChannelV2` to avoid confusion.

Source/Calling/VoiceChannelV3.swift
```diff
@@ -0,0 +1,95 @@
+//
+//  VoiceChannelV3.swift
+//  ZMCDataModel
+//
+//  Created by Jacob on 22/11/16.
```
daehn: Missing copyright header

Source/Calling/VoiceChannelRouter.swift
typfel: Be prepared for a lot of noise

PR #203, opened 2017-01-12
Unify V2 & V3 receiving video observering

typfel:
Both V2 & V3 call centers now implement the `ReceivedVideoObserver` protocol. This simplifies the UI code since will now only need to hook up one observer.

### DIFF
--- a/Source/Calling/WireCallCenter.swift
+++ b/Source/Calling/WireCallCenter.swift
@@ -197,22 +197,27 @@ class ReceivedVideoObserverToken : NSObject {
 @objc
 public class WireCallCenter : NSObject {
     
+    /// Add observer of the state of a converations's voice channel. Returns a token which needs to be retained as long as the observer should be active.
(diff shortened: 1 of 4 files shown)


Source/Calling/WireCallCenter.swift
```diff
@@ -157,6 +175,24 @@ class VoiceChannelStateObserverFilter : NSObject,  VoiceChannelStateObserver {
...
 }
 
+class ReceivedVideoObserverToken : NSObject {
+    
+    var tokenV2 : WireCallCenterObserverToken?
+    var tokenV3 : WireCallCenterObserverToken?
+    
+    deinit {
+        if let token = tokenV3 {
+            WireCallCenterV3.removeObserver(token: token)
```
daehn: What about the `tokenV2`?

typfel: I should document it but it gets de-registred when the token `deinits`

typfel: @daehn documented the observer functions

PR #208, opened 2017-01-16
Add answered case to CallState

typfel:
When you received an incoming call we didn't have state for when
the user had accepted the call. This PR introduces a new case for this called "answered".

The userId has also been marked as optional since AVS does not
always send it in their callback handlers.

AVS also introduced a new function called `wcall_iterate_state` which we now use for
listing all conversations with a non idle call state. Previously we were calling
`wcall_get_state` for each conversation, which can be expensive.


Source/Calling/WireCallCenter.swift
```diff
@@ -60,6 +60,10 @@ extension CallClosedReason {
             return VoiceChannelV2CallEndReason.requested
         case .normalSelf:
             return VoiceChannelV2CallEndReason.requestedSelf
+        case .anweredElsewhere:
```
daehn: Could be like this:

```swift
case .answeredElsewhere, .cancelled:
    return VoiceChannelV2CallEndReason.requested
```

PR #215, opened 2017-01-20
Use new ZMBackendEnvironment API

marcoconti83:


daehn:
✅

daehn:
Still weird that it's possible to call `+new` (which I assume internally calls `alloc` and `init`) even though `init` was marked `NS_UNAVAILABLE` without an error.

PR #219, opened 2017-01-23
Insert missed call system messages for V3 calls

typfel:


Tests/Source/Calling/CallStateObserverTests.swift
```diff
@@ -0,0 +1,150 @@
...
+    }
+    
+    override func tearDown() {
+        localNotificationDispatcher.tearDown()
+        
+        super.tearDown()
+    }
+    
+    func testThatInstanceDoesntHaveRetainCycles() {
+        weak var instance = CallStateObserver(localNotificationDispatcher: localNotificationDispatcher, managedObjectContext: syncMOC)
```
marcoconti83: Is this the new "let's weak self every test" strategy?

typfel: Yes

PR #221, opened 2017-01-24
Read server time delta from notification stream response

typfel:
### Why this change
To handle the case when client's clock has the wrong time a `time` field has been added to the notifications stream response. This time is used to calculate a delta between the client and server time, which is used to correct the current time when passing calling messages to AVS.

### DIFF
--- a/Source/Data Model/NSManagedObjectContext+ServerTimeDelta.swift
+++ b/Source/Data Model/NSManagedObjectContext+ServerTimeDelta.swift
@@ -1,6 +1,6 @@
 //
 // Wire
-// Copyright (C) 2016 Wire Swiss GmbH
+// Copyright (C) 2017 Wire Swiss GmbH
 //
 // This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


Source/Data Model/NSManagedObjectContext+ServerTimeDelta.swift
```diff
@@ -0,0 +1,39 @@
...
+
+public extension NSManagedObjectContext {
+    
+    private static let ServerTimeDeltaKey = "ServerTimeDeltaKey"
+    
+    @objc
+    var serverTimeDelta : TimeInterval {
+        
+        get {
+            return userInfo[NSManagedObjectContext.ServerTimeDeltaKey] as? TimeInterval ?? 0
```
daehn: I would add a `precondition` in case this is only supposed to be read and set from one context.

typfel: @daehn updated

PR #223, opened 2017-01-25
Add file message degradation test

daehn:
# What's in this PR?

* See https://github.com/wireapp/wire-ios-data-model/pull/147

### This PR depends on

- [x] `wire-ios-sync-engine` needs to be compatible with `wire-ios-data-model` version `40` before this PR can get merged.


daehn:
Need to update this one for the new client creation API

PR #224, opened 2017-01-25
Add possibility to change email

vytis:
## What's in this PR

We had possibility to set the email, but not to change it afterwards, so added a new method for this. 

Also fixed a threading issue which means `requestSettingEmailAndPassword` method returns an `emailAlreadySet` error in the delegate instead of throwing.

### DIFF
--- a/Tests/Source/Synchronization/Transcoders/UserProfileUpdateRequestStrategyTests.swift
+++ b/Tests/Source/Synchronization/Transcoders/UserProfileUpdateRequestStrategyTests.swift
@@ -121,6 +121,26 @@ extension UserProfileUpdateRequestStrategyTests {
         XCTAssertEqual(request, expected)
     }
     
+    func testThatItCreatesARequestToChangeEmail() {
+        
+        // GIVEN
+        let selfUser = ZMUser.selfUser(in: self.uiMOC)
+        selfUser.emailAddress = "foo@email.com"
+
+        let newEmail = "mario@example.com"
+        try! self.userProfileUpdateStatus.requestEmailChange(email: newEmail)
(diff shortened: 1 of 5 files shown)


Tests/Source/Synchronization/Transcoders/UserProfileUpdateRequestStrategyTests.swift
```diff
@@ -358,7 +358,24 @@ extension UserProfileUpdateRequestStrategyTests {
...
+        
+        // GIVEN
+        let selfUser = ZMUser.selfUser(in: self.uiMOC)
+        selfUser.emailAddress = "foo@email.com"
+        try! self.userProfileUpdateStatus.requestEmailChange(email: "mario@example.com")
+        XCTAssertTrue(self.waitForAllGroupsToBeEmpty(withTimeout: 0.2))
+        
+        // WHEN
+        let request = self.sut.nextRequest()
+        request?.complete(with: self.successResponse())
```
daehn: I think it would be nice to verify that the request includes the correct payload and method (which we are not doing for most of the `UserProfileUpdateRequestStrategy` at the moment).

vytis: Added few tests for this - only email change requests for now, will add some more when I do phone number change.

vytis: Actually I found that there are tests for most of the requests - will only add those that are missing

vytis: So only email change request payload test was missing - added that

PR #230, opened 2017-01-26
Added method to remove email from profile

vytis:
## What's in this PR?

Changes needed to support email removal from UI. Added a method that sends a request to delete the email and some basic tests for it.


Source/Synchronization/Strategies/UserProfileUpdateRequestStrategy.swift
```diff
@@ -205,6 +216,16 @@ extension UserProfileRequestStrategy : ZMSingleRequestTranscoder {
                 self.userProfileUpdateStatus.didFailEmailUpdate(error: error)
             }
             
+        case self.emailDeleteSync:
+            if response.result == .success {
+                ZMUser.selfUser(in: managedObjectContext).needsToBeUpdatedFromBackend = true
```
marcoconti83: Why do we need to redownload the user here? Are you afraid that just deleting the email from the Core Data user would be an issue if someone had already set a new email on another device?

vytis: I guess it's not really needed in this case. The initial problem was that email deletion event doesn't come from the backend in the same way as editing email does. So I ended up removing email in the UI project from `ZMUser` after request is finished.

marcoconti83: I would make that change at the sync-engine level, to abstract away from the UI the fact that you need to do certain operations on the local data once you get a certain callback.

vytis:
@marcoconti83 removing email in the sync engine

vytis:
It has some UX issues so we will not be implementing email removal for now, closing this PR.

PR #231, opened 2017-01-26
Fix outgoing callkit calls displaying the wrong callie

typfel:
Outgoing Callkit calls where displayed with the e-mail of the self user. Also fixed and added callkit tests for v3.

### DIFF
--- a/Source/Calling/ZMCallKitDelegate.m
+++ b/Source/Calling/ZMCallKitDelegate.m
@@ -106,6 +106,9 @@ NS_ASSUME_NONNULL_END
     else if (0 != self.emailAddress.length) {
         return [[CXHandle alloc] initWithType:CXHandleTypeEmailAddress value:self.emailAddress];
     }
+    else if (nil != self.oneToOneConversation.remoteIdentifier) {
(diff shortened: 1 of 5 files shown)


Source/Calling/ZMCallKitDelegate.m
```diff
@@ -348,6 +358,12 @@ - (void)requestStartCallInConversation:(ZMConversation *)conversation videoCall:
                 [self logErrorForConversation:conversation.remoteIdentifier.transportString line:__LINE__ format:@"Cannot start call: %@", error];
             }
         }];
+        
+        CXCallUpdate *update = [[CXCallUpdate alloc] init];
```
daehn: So now the call is started and updated with the name afterwards?

typfel: yes

typfel: Possible refinement would be not to always use the remoteIdentifier as id, but would then need to differentiate between by adding some kind of prefix.

@username
#remoteidentifier

PR #234, opened 2017-01-27
Do not notify pingback status if notificationEventsToCancel is nil

mikeger:
# Issue

We detected the crash issue happening in `pingbackStatus`'es `didReceiveEncryptedEvents:originalEvents:...` when referencing `originalEvents`. Current hypothesis is that for the parameter `originalEvents` the nil value is passed from Obj-C caller in `ZMMissingUpdateEventsTranscoder`. This situation is theoretically possible (there is no concrete steps to reproduce).

# Solution

There is no reason to call `didReceiveEncryptedEvents:originalEvents:...` if `originalEvents` is nil, therefore the call is guarded by if on the caller side.


mikeger:
@basine please review

mikeger:
@basine please review this one, should be a fix for the crash, we discussed it earlier.

PR #236, opened 2017-01-31
Fix/longer session identifiers

marcoconti83:
Adopt the new APIs of data model, mock transport and cryptobox


Tests/Source/E2EE/UserClientRequestFactoryTests.swift
```diff
@@ -28,25 +28,24 @@ class UserClientRequestFactoryTests: MessagingTest {
     
     var sut: UserClientRequestFactory!
     var authenticationStatus: ZMAuthenticationStatus!
+    var spyKeyStore: SpyUserClientKeyStore!
```
typfel: :)

typfel:
The project file changes are necessary?

Tests/Source/E2EE/UserClientRequestFactoryTests.swift
marcoconti83: It's actually a technical term :)

marcoconti83: for example, see: http://stackoverflow.com/a/8758060/412927

PR #237, opened 2017-01-31
Was missing to save when adding missed call system message

typfel:


typfel:
Hmm do normally have tests checking that we save?

ping @marcoconti83

marcoconti83:
@typfel we don't have an infrastructure to test saving

PR #239, opened 2017-01-31
Fix/migration on startup

marcoconti83:
Perform migration of session identifiers

# Changes
We need to apply the patch that performs the migration BEFORE the notification stream or any APNS is processed. So I distinguished between patches applied after the initial sync, and patches applied at startup.

# Known issues
The same hotfix should be applied when running share extension. So we should later move this code to data model, otherwise `wire-ios-share-engine` won't be able to access it. In fact, we should move the entire hotfix structure to data model.

### DIFF
--- a/Source/Synchronization/ZMHotFix.m
+++ b/Source/Synchronization/ZMHotFix.m
@@ -27,7 +27,6 @@
 #import "ZMUserSession.h"
 
 static NSString* ZMLogTag ZM_UNUSED = @"HotFix";
-static NSString * const LastSavedVersionKey = @"lastSavedVersion";
 NSString * const ZMSkipHotfix = @"ZMSkipHotfix";
 
 @interface ZMHotFix ()
@@ -37,7 +36,6 @@ NSString * const ZMSkipHotfix = @"ZMSkipHotfix";
 
 
 @implementation ZMHotFix
-
 
 - (instancetype)initWithSyncMOC:(NSManagedObjectContext *)syncMOC
 {
@@ -54,7 +52,7 @@ NSString * const ZMSkipHotfix = @"ZMSkipHotfix";
     return self;
 }
 
-- (void)applyPatches
+- (void)applyPatchesAfterSyncCompleted
 {
     if ([[self.syncMOC persistentStoreMetadataForKey:ZMSkipHotfix] boolValue]) {
         ZMLogDebug(@"Skipping applying HotFix");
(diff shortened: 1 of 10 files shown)


Source/Synchronization/ZMHotFix.m
```diff
@@ -54,24 +52,35 @@ - (instancetype)initWithHotFixDirectory:(ZMHotFixDirectory *)hotFixDirectory syn
...
         return;
     }
     
     NSString * currentVersionString = [[[NSBundle bundleForClass:ZMUserSession.class] infoDictionary] objectForKey:@"CFBundleShortVersionString"];
-    [self applyPatchesForCurrentVersion:currentVersionString];
+    [self applyPatchesForCurrentVersion:currentVersionString afterSync:YES];
 }
 
-- (void)applyPatchesForCurrentVersion:(NSString *)currentVersionString;
+- (void)applyPatchesAtStartup
```
typfel: This code is duplicated in `applyPatchesAfterSyncCompleted`. Could be extracted into a shared method.

Source/Synchronization/ZMHotFix.m
```diff
@@ -54,24 +52,35 @@ - (instancetype)initWithHotFixDirectory:(ZMHotFixDirectory *)hotFixDirectory syn
...
+    if ([[self.syncMOC persistentStoreMetadataForKey:ZMSkipHotfix] boolValue]) {
+        ZMLogDebug(@"Skipping applying HotFix");
+        return;
+    }
+    
+    NSString * currentVersionString = [[[NSBundle bundleForClass:ZMUserSession.class] infoDictionary] objectForKey:@"CFBundleShortVersionString"];
+    [self applyPatchesForCurrentVersion:currentVersionString afterSync:NO];
+}
+
+- (void)applyPatchesForCurrentVersion:(NSString *)currentVersionString afterSync:(BOOL)afterSync;
```
typfel: There's many methods passing the `afterSync` boolean around. An alternative to that would be to inject the dependences and change the signature of this method to:
`applyPatches:(NSArray *)patches currentVersion:(NSString *) currentVersion  forKey:(NSString *)key`

marcoconti83:
@typfel you raised some good points. Given that we need to re-do this on the share extension too (hence we need to move the patching to the data model layer), I will close this PR.

marcoconti83:
Moved to https://github.com/wireapp/wire-ios-data-model/pull/153

PR #240, opened 2017-02-01
Remove references to static AVS library in xcconfig

typfel:


Resources/Configurations/libs.xcconfig
```diff
@@ -17,14 +17,11 @@
 // 
 
 
-AVS_ROOT = $(PROJECT_DIR)/Source/AVS
```
mikeger: @typfel you should also delete the folder from `Source/AVS` since it's checked in (if it is not needed)

typfel: Right, will do

PR #242, opened 2017-02-02
Fix group calls not hanging up properly

typfel:
We should not leave the voice channel immediately when we request to leave


marcoconti83:
So when do we leave it?

typfel:
In Callkit  you request to leave by creating a "leave" action. Later we will be asked to fulfil this "leave" action and that's where we should leave the voice channel.

PR #248, opened 2017-02-06
perform changes from flowManagerDidChange notification on correct queue

basine:
Only line 254 is a fix: FlowManagerVideoReceiveStateNotification are probably not guaranteed to arrive on the correct thread.

The other changes are just improvements: The VoiceGainNotification can set the conversationID as object so you don't need to filter notifications when receiving them. Note that this only works if you cast the object to NSUUID otherwise it will fail comparing Any to UUID ¯\_(ツ)_/¯


typfel:
Thanks!

PR #255, opened 2017-02-07
New observer system

basine:
**Note:** This needs a new release of ZMCDataModel, WireRequestStrategy and WireMessageStrategy

**Changes in this PR**
1) Refactoring of the WireCallCenterV2
2) Adding and tearing down SearchUserObserverSnapshots when searchDirectory changes or is torndown
3) Adjusting tests using new observer system

**TODO**
- [x] merge & release with https://github.com/wireapp/wire-ios-data-model/pull/138
- [x] release WireRequestStrategy with new data model
- [x] merge & release with https://github.com/wireapp/wire-ios-message-strategy/pull/34
- [x] rerun all tests


typfel:
How does the new observer system work when in the background, is it always reporting changes? `WireCallCenterV2` and everything build on top of it expects observers to also fire when in the background.

Source/Public/ZMUserSession.h
```diff
@@ -32,7 +32,7 @@
 @class ZMConversation;
 @class UserClient;
 @class ZMProxyRequest;
-@class WireCallCenterV2;
+//@class WireCallCenterV2;
```
typfel: Delete

Source/Public/ZMUserSession.h
```diff
@@ -133,7 +133,7 @@ extern NSString * const ZMTransportRequestLoopNotificationName;
 @property (nonatomic, readonly) TopConversationsDirectory *topConversationsDirectory;
 
 /// CallCenter for V2 calls
-@property (nonatomic, readonly) WireCallCenterV2 *wireCallCenterV2;
+//@property (nonatomic, readonly) WireCallCenterV2 *wireCallCenterV2;
```
typfel: Delete

Source/Synchronization/ZMSyncStrategy.m
```diff
@@ -421,9 +426,7 @@ - (void)managedObjectContextDidSave:(NSNotification *)note;
         
         NSSet *conversationsWithCallChanges = [callStateChanges allContainedConversationsInContext:strongUiMoc];
         if (conversationsWithCallChanges != nil) {
-            [NSNotificationCenter.defaultCenter postNotificationName:WireCallCenterV2.CallStateDidChangeNotification
-                                                              object:nil
-                                                            userInfo:@{ @"updated": conversationsWithCallChanges }];
+            [strongUiMoc.wireCallCenterV2 callStateDidChangeWithConversations:conversationsWithCallChanges];
```
typfel: I think the plan was to move this into the generic userInfo merge.

`NSManagedObjectContext+UserInfoMerge.swift`

basine:
@typfel It's currently not processing or forwarding any changes while in the background. I did a few tests with the app and it seemed to work fine nonetheless, but we might have to review this behaviour.

Source/Synchronization/ZMSyncStrategy.m
basine: Maybe this should be a different PR

typfel:
@basine I think we need to review it, you wouldn't be able to receive a call in the background.

basine:
@typfel I tried this and it worked. Nevertheless this would require another ZMCDataModel release and should be a different PR in that case.

typfel:
@basine well then

"It's currently not processing or forwarding any changes while in the background"

is not working as it should or you didn't try it Callkit  or with calling3.

basine:
@typfel as I said, I tried calling while the app was in the background and it worked. We should merge this PR anyway and then fix the background behaviour, because this PR currently blocks the work for everyone else.

typfel:
@basine right maybe we can do it an a follow up PR but we should be aware of what breaks in dev when we merge this. Need to disable internal and so on.

basine:
@typfel I understand your concerns, but this is not going to be released right away anyway - I expect a few more bugs to be uncovered since the refactoring was rather major. Therefore there will be follow up PRs for sure.

PR #257, opened 2017-02-09
Don’t create snapshots for conversations with NoActiveUser state

basine:


typfel:
Can you explain why this fixes the crash?

basine:
Yes, the bug was the following: When you insert a new conversation, the app crashes explicitly unwrapping the conversations remoteIdentifier as a result of a callStateChange notification.

The reason: We only create snapshots "on demand" when the conversation changes and the new voiceChannelState is not invalid (and now also not noActiveUsers). If a new snapshot is created, a notification is sent about the initial state change (which it should not send at all when the callState is noActiveUsers). When you insert a new conversation locally, it does not have a remoteIdentifier yet, ergo when we create a VoiceChannelStateSnapshot for this object, the app will crash.

PR #262, opened 2017-02-14
We need to manually notify about the call state when rejecting a call

typfel:
- [x] need fix from AVS


daehn:
@typfel conflicting 🔥

typfel:
@daehn fixed

PR #265, opened 2017-02-15
return genericMessageStrategy as contextChangeTracker

basine:
- [x] Needs WireMessageStrategy release with https://github.com/wireapp/wire-ios-message-strategy/pull/39


Tests/Source/Synchronization/Strategies/CallingRequestStrategyTests.swift
```diff
@@ -0,0 +1,47 @@
...
+        super.tearDown()
+    }
+    
+    func testThatItReturnsItselfAndTheGenericMessageStrategyAsContextChangeTracker(){
+        // when
+        let trackers = sut.contextChangeTrackers
+        
+        // then
+        XCTAssertTrue(trackers.first is CallingRequestStrategy)
+        XCTAssertTrue(trackers.last is GenericMessageRequestStrategy)
```
marcoconti83: I'm not sure this test is that useful. With this test we are testing internal implementation (e.g. what kind of class is it?) when we should instead test external behavior (does it return the right request if the state is X Y Z?).

PR #269, opened 2017-02-16
Don't ignore calls in newly created groups

typfel:
We were previously ignoring calls in newly created groups because they would
show up during sync. But now the CallState transcoder is only listening to live
events so this should not be necessary anymore.

see line: https://github.com/wireapp/wire-ios-sync-engine/blob/226997e1f3db0d2d04f0d01e2946d2bb1d9dc357/Source/Synchronization/ZMCallStateTranscoder.m#L247

https://wearezeta.atlassian.net/browse/ZIOS-5969

### DIFF
--- a/Source/Synchronization/ZMCallStateTranscoder.m
+++ b/Source/Synchronization/ZMCallStateTranscoder.m
@@ -275,7 +275,6 @@ _Pragma("clang diagnostic pop")
         }
         else if (event.type == ZMUpdateEventConversationCreate) {
             conversation.callStateNeedsToBeUpdatedFromBackend = YES;
-            conversation.isIgnoringCall = YES;
         }
     }
 }


Source/Synchronization/ZMCallStateTranscoder.m
```diff
@@ -275,7 +275,6 @@ - (void)processEvents:(NSArray<ZMUpdateEvent *> *)events
         }
         else if (event.type == ZMUpdateEventConversationCreate) {
```
basine: Didn't we say that we only wanted to set this for non-live events? Now if you are starting your phone and its syncing, it's possible it starts the ring and you can't do anything because the app is still syncing?

basine: Ah sorry, did not read the description of the PR

basine: It's quite possible that we forward call events directly while we are still syncing so we can setup calls faster. Check `- (void)processUpdateEvents:(NSArray *)events ignoreBuffer:(BOOL)ignoreBuffer` in wire-sync-engine ZMSyncStrategy. We filter calling related events and forward them. Have you tested your changes in the app?

daehn: I think if we do this it could be an issue as calling now sends setup messages, if we process them before done syncing (e.g. in the wrong order) we might fail to decrypt them or not?

typfel: @daehn this is only for v2 calls which doesn't signal with encrypted messages

typfel: @basine this change only affects `ZMUpdateEventConversationCreate` which is not a `flowEvent` so it should not be forwarded as a live event during the sync.

PR #272, opened 2017-02-17
Removed tracking identifier

mikeger:
- Removing is easy.
- Some tests are failing now, checking if I broke something or it was broken.


vytis:
Few tests are still failing from the previous changes - https://github.com/wireapp/wire-ios-sync-engine/pull/273

daehn:
@mikeger can this one be merged? I think the classic GitHub race condition re-added the label?

mikeger:
@daehn I am running the tests now

PR #278, opened 2017-02-23
Feature/add strategy superclass

typfel:


Source/Synchronization/Strategies/AddressBookUploadRequestStrategy.swift
```diff
@@ -58,48 +49,40 @@ private let addressBookLastUploadedIndex = "ZMAddressBookTranscoderLastIndexUplo
...
-        self.init(authenticationStatus: authenticationStatus,
-            clientRegistrationStatus: clientRegistrationStatus,
-            managedObjectContext: moc
-        )
+    /// Default AddressBook Generator
+    static let defaultAddressBookGenerator = { return AddressBook.factory() }
+
+    public override var configuration: ZMStrategyConfigurationOption { return .allowsRequestsDuringEventProcessing }
+    
+    public override init(managedObjectContext moc: NSManagedObjectContext, appStateDelegate: ZMAppStateDelegate) {
```
marcoconti83: Why was the addressBookGenerator injection switched from ctor injection to static variable injection? If possible, I would keep it as ctor injection. 
Static variable injection is bad. E.g. if you forget to reset it in a test, future tests will fail because of this.

Source/Synchronization/Strategies/TopConversationsRequestStrategy.swift
```diff
@@ -20,25 +20,25 @@ import Foundation
...
+@objc public class TopConversationsRequestStrategy : ZMAbstractRequestStrategy {
     
     let conversationDirectory : TopConversationsDirectory
-    
     fileprivate var topPeopleSync : ZMSingleRequestSync! = nil
     
+    /// Defines when requests for this Strategy are allowed to be sent
+    public override var configuration: ZMStrategyConfigurationOption { return .allowsRequestsDuringEventProcessing }
+
+    @available (*, unavailable, message: "use `init(managedObjectContext:appStateDelegate:conversationDirectory)`instead")
```
marcoconti83: 👍

Source/Synchronization/Strategies/UserImageStrategy.swift
```diff
@@ -116,10 +119,12 @@ public class UserImageStrategy : NSObject, ZMDownstreamTranscoder, ZMUpstreamTra
         self.recoverFromInconsistentUserImageStatus()
     }
     
-    public func tearDown() {
+    public override func tearDown() {
+        super.tearDown()
         tornDown = true
         assetPreprocessingTracker.tearDown()
         NotificationCenter.default.removeObserver(self)
+        super.tearDown()
```
marcoconti83: Why is `tearDown` called twice? And do we need to set `tearDown` to `true` or is it something the superclass can do?

Source/Synchronization/Transcoders/ZMConnectionTranscoder.m
```diff
@@ -54,12 +54,20 @@ @interface ZMConnectionTranscoder (Pagination) <ZMSimpleListRequestPaginatorSync
...
 @implementation ZMConnectionTranscoder
 
-- (instancetype)initWithManagedObjectContext:(NSManagedObjectContext *)moc syncStatus:(SyncStatus *)syncStatus clientRegistrationDelegate:(id<ClientRegistrationDelegate>)clientRegistrationDelegate;
+- (instancetype)initWithManagedObjectContext:(NSManagedObjectContext *)moc appStateDelegate:(id<ZMAppStateDelegate>)appStateDelegate;
 {
-    self = [super initWithManagedObjectContext:moc];
+    Require(NO);
+    self = [super initWithManagedObjectContext:moc appStateDelegate:appStateDelegate];
+    NOT_USED(self);
+    self = nil;
```
marcoconti83: What exactly is happening here? Why do we set self to nil? Maybe we want to assert instead?

Source/Synchronization/Transcoders/ZMLastUpdateEventIDTranscoder.m
```diff
@@ -31,35 +31,40 @@ @interface ZMLastUpdateEventIDTranscoder ()
...
-@property (nonatomic, weak) id<ClientRegistrationDelegate> clientRegistrationDelegate;
 @property (nonatomic) BOOL didStartSlowSync;
 
 @end
 
 
 @implementation ZMLastUpdateEventIDTranscoder
 
-- (instancetype)initWithManagedObjectContext:(NSManagedObjectContext *)moc {
+- (instancetype)initWithManagedObjectContext:(NSManagedObjectContext *)moc appStateDelegate:(id<ZMAppStateDelegate>)appStateDelegate
```
marcoconti83: I think there's a way to mark an `init` as not available in Obj-c too (see Swift classes above). Maybe `NS_UNAVAILABLE`?

Source/Synchronization/ZMSyncStateManager.h
```diff
@@ -0,0 +1,62 @@
...
+@class SyncStatus;
+@class ClientUpdateStatus;
+
+@protocol ZMSyncStateDelegate;
+@protocol ZMRequestCancellation;
+@protocol ZMApplication;
+@protocol ClientDeletionDelegate;
+
+
+@interface ZMSyncStateManager : NSObject <ZMAppStateDelegate>
```
marcoconti83: I would vote for a different name. `Manager` doesn't really say what it does. If this is just a list of statuses, let's call it `list` or `directory`.
Also `SyncState` is already used for something else (states in the old state machine). Maybe `AppStatus`?

Source/Synchronization/ZMSyncStateManager.h
```diff
@@ -0,0 +1,62 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+@import WireMessageStrategy;
+
+@class ZMCookie;
```
marcoconti83: If this had been written in Swift we would have avoided a lot of typing (here and when redefining the variables internally as readwrite). Maybe it has to be Obj-c otherwise we have to expose too much Obj-c headers as `public` for Swift to see it? Otherwise I think it would be good to write new classes in Swift.

Source/Synchronization/ZMSyncStrategy+EventProcessing.m
```diff
@@ -1,15 +1,27 @@
 //
-//  ZMSyncStrategy+EventProcessing.m
-//  zmessaging-cocoa
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
```
marcoconti83: Go get 'em, Danger-bot!

Tests/Source/Synchronization/Sync States/ZMBackgroundTaskStateTests.m
```diff
@@ -16,6 +16,7 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 // 
 
+// TODO Sabine reenable
```
marcoconti83: TODO?

Source/Synchronization/ZMSyncStateManager.h
typfel: How about `StatusDirectory`

typfel: But this should be renamed to together with the `ZMAppStateDelegate` from `wire-ios-message-strategy`

marcoconti83: `StatusDirectory` sounds good

typfel: More suggestions:

`ApplicationStatusDirectory` which implements the `ApplicationStatusDelegate`

PR #280, opened 2017-02-23
Make ZMCallStateTranscoder a request strategy

typfel:
For the ongoing project of removing the old state machine we need to convert all transcoder into request strategies.

### DIFF
--- a/Source/Synchronization/ZMCallStateTranscoder.h
+++ b/Source/Synchronization/ZMCallStateTranscoder.h
@@ -20,6 +20,7 @@
 @import Foundation;
 @import CoreData;
 @import WireRequestStrategy;
+@import WireMessageStrategy;
 #import "ZMObjectStrategyDirectory.h"
 
 @class ZMConversation;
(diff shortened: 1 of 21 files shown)


Source/Synchronization/ZMCallStateTranscoder.h
```diff
@@ -33,11 +34,11 @@ typedef NS_ENUM(uint8_t, ZMCallEventSource) {
 
 
 
-@interface ZMCallStateTranscoder : ZMObjectSyncStrategy <ZMObjectStrategy>
+@interface ZMCallStateTranscoder : ZMAbstractRequestStrategy <ZMEventConsumer, ZMContextChangeTrackerSource>
```
mikeger: May be would make sense then to rename the `ZMCallStateTranscoder` to `ZMCallStateRequestStrategy` ?

Tests/Source/Synchronization/Transcoders/ZMCallStateTranscoderTests.h
```diff
@@ -21,14 +21,15 @@
 @import ZMCDataModel;
```
mikeger: I guess having the header is not necessary for the test class, unless if something is inheriting from this test case?

Source/Synchronization/ZMCallStateTranscoder.m
```diff
@@ -101,22 +101,22 @@ - (instancetype)initWithManagedObjectContext:(NSManagedObjectContext *)moc
 }
 _Pragma("clang diagnostic pop")
```
mikeger: Is this dance with init still necessary?

typfel: No I just missed it. We should delete the init

Tests/Source/Synchronization/Transcoders/ZMCallStateTranscoderTests.h
typfel: There's a an extension on the test

typfel: `ZMCallStateTranscoderTests (VideoCalling)`

typfel:
@mikeger updated

PR #281, opened 2017-02-24
Use monochrome heart copy in push action to like a message

daehn:
# What's in this PR?

* Copy change as requested by design.

### DIFF
--- a/Resources/Base.lproj/Push.strings
+++ b/Resources/Base.lproj/Push.strings
@@ -170,7 +170,7 @@
 "push.notification.action.message.reply" = "Reply";
 "push.notification.action.message.reply.button.title" = "Send";
 "push.notification.action.conversation.mute" = "Mute";
-"push.notification.action.message.like" = "♡";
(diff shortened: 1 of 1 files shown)


daehn:
There will be another copy change for notifications when users like a message.

Resources/Base.lproj/Push.strings
```diff
@@ -170,7 +170,7 @@
 "push.notification.action.message.reply" = "Reply";
 "push.notification.action.message.reply.button.title" = "Send";
 "push.notification.action.conversation.mute" = "Mute";
-"push.notification.action.message.like" = "❤️";
+"push.notification.action.message.like" = "♥";
```
typfel: It's so tiny

typfel:
Why the do not merge?

daehn:
Will change it in the push when someone liked your message as well

daehn:
The decision has been made, the heart in the push if someone else liked a message will stay the same.

typfel:
What?!

PR #282, opened 2017-02-24
Turn ZMFlowSync into a request strategy

typfel:
For the ongoing project of removing the old state machine we need to convert all transcoder into request strategies.

### DIFF
--- a/Source/Synchronization/Sync States/ZMBackgroundState.m
+++ b/Source/Synchronization/Sync States/ZMBackgroundState.m
@@ -102,7 +102,7 @@
 {
     id<ZMObjectStrategyDirectory> directory = self.objectStrategyDirectory;
     NSMutableArray *transcoders = @[
-                            directory.flowTranscoder,
+                            directory.callFlowRequestStrategy,
                             ].mutableCopy;
 
     if ([ZMUserSession useCallKit]) {
--- a/Source/Synchronization/Sync States/ZMBackgroundState.m
+++ b/Source/Synchronization/Sync States/ZMBackgroundState.m
@@ -100,14 +100,7 @@
 
 - (ZMTransportRequest *)nextRequest
 {
-    id<ZMObjectStrategyDirectory> directory = self.objectStrategyDirectory;
-    NSMutableArray *transcoders = @[
-                            directory.callFlowRequestStrategy,
-                            ].mutableCopy;
-
-    if ([ZMUserSession useCallKit]) {
-        [transcoders addObject:directory.callStateRequestStrategy];
-    }
+    NSMutableArray *transcoders = @[].mutableCopy;
     
     ZMTransportRequest *nextRequest = [self nextRequestFromTranscoders:transcoders];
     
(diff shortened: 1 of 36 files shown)


Source/Synchronization/ZMObjectStrategyDirectory.h
```diff
@@ -52,8 +52,8 @@
 @property (nonatomic, readonly) ZMPhoneNumberVerificationTranscoder *phoneNumberVerificationTranscoder;
 @property (nonatomic, readonly) ZMLoginTranscoder *loginTranscoder;
 @property (nonatomic, readonly) ZMLoginCodeRequestTranscoder *loginCodeRequestTranscoder;
-@property (nonatomic, readonly) ZMFlowSync *flowTranscoder;
-@property (nonatomic, readonly) ZMCallStateRequestStrategy *callStateRequestStrategy;
+//@property (nonatomic, readonly) ZMCallFlowRequestStrategy *callFlowRequestStrategy;
+//@property (nonatomic, readonly) ZMCallStateRequestStrategy *callStateRequestStrategy;
```
mikeger: Comment 😛

Source/Synchronization/Sync States/ZMBackgroundState.m
```diff
@@ -100,14 +100,7 @@ - (void)didEnterForeground
...
 {
-    id<ZMObjectStrategyDirectory> directory = self.objectStrategyDirectory;
-    NSMutableArray *transcoders = @[
-                            directory.flowTranscoder,
-                            ].mutableCopy;
-
-    if ([ZMUserSession useCallKit]) {
-        [transcoders addObject:directory.callStateRequestStrategy];
-    }
+    NSMutableArray *transcoders = @[].mutableCopy;
```
marcoconti83: This seems useless. Why create an empty array to see if any element inside returns a request? we can just assume there's no request.

Tests/Source/Calling/VoiceChannelV2Tests+VideoCalling.m
```diff
@@ -60,13 +60,13 @@ - (void)testThatItCallsIsSendingVideoForParticipantAndReturnValue_hasFlowManager
     XCTAssertNil(error);
     XCTAssertTrue([self waitForCustomExpectationsWithTimeout:0.5f]);
     
-    ZMFlowSyncInternalFlowManagerOverride = nil;
+    ZMCallFlowRequestStrategyInternalFlowManagerOverride = nil;
```
marcoconti83: I know you didn't touch this code but just renamed it, but isn't it easier to do this

`ZMCallFlowRequestStrategyInternalFlowManagerOverride = nil;`

in the `tearDown` of the test, instead of having to call it in all tests? Maybe it's worth doing it now.

Source/Synchronization/Sync States/ZMBackgroundState.m
typfel: This is change is there to make it compile. This entire class will be removed in the upcoming PR which removes  the state machine background states.

Tests/Source/Calling/VoiceChannelV2Tests+VideoCalling.m
typfel: This I can do.

Source/Synchronization/Sync States/ZMBackgroundState.m
marcoconti83: Ok!

typfel:
@marcoconti83 @mikeger updated

PR #287, opened 2017-02-27
Fix call appearing connected before it is

typfel:
When answering a call via callkit it would show up as connected directly after
tapping the answer button since we fulfilled the answer action instantly and
not after establishing the call.

This PR also cleans up some other callkit callbacks. We were for example
calling `reportOutgoingCall:connectedAtDate` when incoming calls were
established.

https://wearezeta.atlassian.net/browse/ZIOS-8137


Tests/Source/Calling/ZMCallKitDelegateTests.swift
```diff
@@ -82,6 +82,59 @@ class MockCallKitCallController: NSObject, CallKitCallController {
...
+    }
+    
+    override func fail() {
+        hasFailed = true
+    }
+    
+}
+
+@available(iOS 10.0, *)
+class MockStartCallAction : CXStartCallAction {
```
daehn: 👍

typfel:
@daehn yes, very happy the mocking worked. might try to convert the callkit integration into Swift next :)

mikeger:
@typfel be careful with Swift, I tried it back in the day and it was causing the issues with iOS9 compatibility, i.e. the library was causing the crash. The problem was that you can have `#if` checks in swift, but if you import the non-existent module (`CallKit`) it would be unable to load zmessaging on the app start. May be I was holding it wrong.

PR #288, opened 2017-02-28
Do not ring with CallKit for muted conversations

mikeger:
Also the change that allows to end the call even if CallKit is refusing to end.


mikeger:
Adding the test

PR #289, opened 2017-02-28
Only consider client and asset messages in top people calculation

daehn:
# What's in this PR?

* System messages were taking into account in the top people calculation.
* Fetch `ZMAssetClientMessage` and `ZMClientMessage` count instead of `ZMMessage`.
* This will add an extra request per conversation, we will have to see if this performs well and otherwise might want to reverse iterate over the messages in a conversation until the `serverTimestamp` is older than a month.
* Add missing test assertion.

### DIFF
--- a/Source/UserSession/Search/TopConversationsDirectory.swift
+++ b/Source/UserSession/Search/TopConversationsDirectory.swift
@@ -142,8 +142,8 @@ fileprivate extension ZMConversation {
 
     func lastMonthMessageCount() -> Int {
         guard let identifier = remoteIdentifier, let moc = managedObjectContext else { return 0 }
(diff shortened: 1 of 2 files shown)


Source/UserSession/Search/TopConversationsDirectory.swift
```diff
@@ -165,5 +165,11 @@ fileprivate extension ZMMessage {
         return NSCompoundPredicate(andPredicateWithSubpredicates: [lastMonthPredicate, conversationPredicate])
     }
 
+    static func countOfMessageInTheLastMonth(in identifier: UUID, context: NSManagedObjectContext) -> Int {
```
vytis: typo - should be `countOfMessages`

daehn:
@vytis fixed!

PR #296, opened 2017-03-06
Track the call ended reason correctly when hanging up on calling device

daehn:
# What's in this PR?

* We need to check the user in order to know if the reason for the end of the call is `self` or `other`.


daehn:
Waiting for the tests to complete…

PR #297, opened 2017-03-06
Calling degradation

typfel:
## Reason behind this PR?

User needs to know when conversation security gets degraded during the call

## What's in this PR?

* Introduce two new call states for degraded incoming / outgoing calls
* Add methods on the `VoiceChannel` for responding to degraded calls.

### Note
Not all tests there yet.

---

Needs release with changes in:

- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/17
- [x] https://github.com/wireapp/wire-ios-message-strategy/pull/45
- [x] https://github.com/wireapp/wire-ios-data-model/pull/201


Source/Calling/VoiceChannelRouter.swift
```diff
@@ -149,6 +149,22 @@ extension VoiceChannelRouter : CallObservers {
 
 extension VoiceChannelRouter : CallActions {
     
+    public func continueByDecreasingConversationSecurity(userSession: ZMUserSession) {
+        guard let conversation = conversation else { return }
+        conversation.makeNotSecure()
```
marcoconti83: Is it OK that the other function (`leaveAndKeepDegradedConversationSecurity`) switches to the Sync context, but this one doesn't?

daehn: @marcoconti83 `markAsNotSecure` should be called from the UI, see https://github.com/wireapp/wire-ios-data-model/pull/201/files#diff-9411e2b6da1d80670b7926da61b5b6c0R150

Source/Calling/WireCallCenter.swift
```diff
@@ -152,6 +154,13 @@ class VoiceChannelStateObserverFilter : NSObject,  VoiceChannelStateObserver {
         super.init()
         
         self.token = VoiceChannelStateObserverToken(context: context, observer: self)
+        self.conversationToken = ConversationChangeInfo.add(observer: self, for: conversation)
+    }
+    
+    deinit {
+        if let token = conversationToken {
+            ConversationChangeInfo.remove(observer: token, for: observedConversation)
```
daehn: This is not one of the self removing observers?

Source/Synchronization/Transcoders/ZMCallStateLogger.m
```diff
@@ -75,18 +75,20 @@ - (NSString *)descriptionForVoiceChannelState:(VoiceChannelV2State)state
             return @"NoActiveUsers";
         case VoiceChannelV2StateIncomingCall:
             return @"IncomingCall";
+        case VoiceChannelV2StateIncomingCallDegraded:
```
daehn: Not really related to this PR: This `enum` is always confusing me as it has `V2` in its name but is used for both v2 and v3.

vytis: Yeah we should refactor this at some point, probably after removing v2. It has states that are in both v2 and v3, but also some that are version-specific.

Source/Calling/WireCallCenter.swift
vytis: How can we check?

Source/Synchronization/Transcoders/ZMCallStateLogger.m
typfel: The V3 call state is mapped back to V2 because that's what the UI understands. Obviously not ideal.

Source/Calling/WireCallCenter.swift
daehn: Some of the observers internally observe a `NSNotification` and remove themselves as observer in their `deinit`.

daehn: Either way this doesn't block this PR from getting merged.

typfel: Ah good catch. It actually is but the method for  doing manual removal is still there.

Source/Calling/WireCallCenter.swift
```diff
@@ -125,8 +125,9 @@ class VoiceChannelStateObserverToken : NSObject, WireCallCenterV2CallStateObserv
     
     func callCenterDidChange(callState: CallState, conversationId: UUID, userId: UUID?) {
         guard let conversation = ZMConversation(remoteID: conversationId, createIfNeeded: false, in: context) else { return }
-        
-        observer?.callCenterDidChange(voiceChannelState: callState.voiceChannelState, conversation: conversation, callingProtocol: .version3)
+        let voiceChannelState = callState.voiceChannelState(securityLevel: conversation.securityLevel)
+        observer?.callCenterDidChange(voiceChannelState: voiceChannelState, conversation: conversation, callingProtocol: .version3)
+        conversation.securityLevel == .secureWithIgnored
```
typfel: This should be deleted.

Source/Calling/VoiceChannelV2.h
```diff
@@ -43,9 +43,11 @@ typedef NS_ENUM(uint8_t, VoiceChannelV2State) {
     VoiceChannelV2StateInvalid = 0,
     VoiceChannelV2StateNoActiveUsers, ///< Nobody is active on the voice channel
     VoiceChannelV2StateOutgoingCall, ///< We are connecting and nobody is in a connected state on the voice channel yet (ie: we are calling)
+    VoiceChannelV2StateOutgoingCallDegraded, ///<
```
mikeger: No comment?

PR #301, opened 2017-03-07
Manually fire a notification for the list indicator when the call state changes

daehn:
# What's in this PR?

* This is needed in order to properly update the conversation list indicator on the UI.


daehn:
@basine updated, please re-review.

daehn:
Added the `DO NOT MERGE` label to ensure we don't merge this before the release.

PR #295, opened 2017-03-06
Fix/message strategies refactoring

marcoconti83:
# Reason for this pull request
Adopt the new wire-ios-message-strategy from https://github.com/wireapp/wire-ios-message-strategy/pull/37

PR #304, opened 2017-03-08
Insert system messages for performed calls

daehn:
# What's in this PR?

* Add `SystemMessageCallObserver`, a class that observes the call state and inserts a system message of type `.performedCall` if a call happened.

### This PR depends on:
- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/204 has been merged.


Source/Calling/SystemMessageCallObserver.swift
```diff
@@ -0,0 +1,91 @@
...
+
+
+private let log = ZMSLog(tag: "Calling System Message")
+
+
+@objc public final class SystemMessageCallObserver: NSObject, VoiceChannelStateObserver {
+
+    var token: NSObjectProtocol?
+    var callerByConversation = [ZMConversation: ZMUser]()
+    var startDateByConversation = [ZMConversation: Date]()
```
mikeger: I guess it would be nice to clean the list once in a while, for example when we are appending the message. Afterwards the information is not necessary to keep any more.

daehn: @mikeger true, forgot to remove the elements after inserting the message.

daehn: @mikeger updated!

Source/Calling/SystemMessageCallObserver.swift
```diff
@@ -0,0 +1,93 @@
...
+            startDateByConversation[conversation] = Date()
+        default: break
+        }
+    }
+
+    public func callCenterDidEndCall(reason: VoiceChannelV2CallEndReason, conversation: ZMConversation, callingProtocol: CallingProtocol) {
+        if let caller = callerByConversation[conversation], let startDate = startDateByConversation[conversation] {
+            let duration = -startDate.timeIntervalSinceNow
+            log.info("Appending performed call message: \(duration), \(caller.displayName), \"\(conversation.displayName)\"")
+            conversation.appendPerformedCallMessage(with: duration, caller: caller)
```
typfel: Don't you need to enqueue a save here?

daehn: That's done inside the `append` method, but could be moved here.

typfel: Ah ok

Source/Calling/SystemMessageCallObserver.swift
```diff
@@ -0,0 +1,93 @@
...
+        // no-op
+    }
+
+}
+
+
+private extension ZMConversation {
+
+    func callingUser() -> ZMUser? {
+        guard let selfUser = managedObjectContext.map(ZMUser.selfUser) else { return nil }
```
mikeger: So Yoda if syntax was the first step, now it's progressing :)

Source/Calling/SystemMessageCallObserver.swift
```diff
@@ -53,6 +53,8 @@ private let log = ZMSLog(tag: "Calling System Message")
             let duration = -startDate.timeIntervalSinceNow
             log.info("Appending performed call message: \(duration), \(caller.displayName), \"\(conversation.displayName)\"")
             conversation.appendPerformedCallMessage(with: duration, caller: caller)
+            callerByConversation[conversation] = nil
+            startDateByConversation[conversation] = nil
```
mikeger: Should we cleanup also in `else`? Or even in `defer`?

daehn: @mikeger updated!

mikeger:
@daehn why not merge this one?

daehn:
Because I am still running the tests locally

mikeger:
@daehn ok, thanks for the info.

daehn:
And needed to merge https://github.com/wireapp/wire-ios-sync-engine/pull/295 first

daehn:
@mikeger tests are ✅, ready to merge!

PR #305, opened 2017-03-08
Revert answer call handling

typfel:
We need to revert the answer call handling until we have more time investigate why the "correct" api usage breaks incoming calls.


mikeger:
Strange CallKit ways.

typfel:
We already found out why it was not working and Callkit is not to blame. We had a deadlock :)

PR #306, opened 2017-03-09
Add degraded variant for connecting state

typfel:
We need degraded variant for the connecting state. This is handles the case when you find out the call is degraded after you've accepted the call.


typfel:
*running tests*

typfel:
*tests passed*

typfel:
@daehn updated with fixes for tests

PR #307, opened 2017-03-13
Move from using `/search/common` to using `/search/contacts`

daehn:
# What's in this PR?

* We do not care about the actual users and only need to know the count.
* We now perform a search using `/search/contacts` and use the username as query with a size limit of 1.

### This PR depends on:
- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/209 has been merged.


mikeger:
🔥 Conflict

Source/UserSession/Search/ZMCommonContactsSearch.m
```diff
@@ -78,70 +81,75 @@ - (instancetype)initWithTransportSession:(ZMTransportSession *)transportSession
...
             return YES;
         }
     }
     return NO;
 }
 
-- (void)notifyDelegateWithResult:(NSOrderedSet *)users
+- (void)notifyDelegateWithResult:(NSUInteger)numberOfConnections
 {
-    [self.delegate didReceiveCommonContactsUsers:users forSearchToken:self.searchToken];
```
mikeger: I guess the method `didReceiveCommonContactsUsers` should be removed from the delegate then?

daehn: It has been removed, see https://github.com/wireapp/wire-ios-data-model/pull/209/.

daehn:
@mikeger updated!

PR #292, opened 2017-03-01
Profile image v3: tracking state of upload

vytis:
## Reason behind this PR

We want to move profile images to use assets v3.

## What's in this PR?

It is first of few PRs for the full implementation. This part adds `UserProfileImageUpdateStatus` that tracks state of profile image upload. At the moment it handles starting the image resize and collecting the results. 
Request strategy for uploading resized images and another one for updating profile will be coming in separate PRs.

P.S. This PR is being merged to intermediate branch (`feature/profile-picture-v3`) where full changes will be collected before merging to `develop`.

PR #308, opened 2017-03-14
Profile picture v3: Request strategy for uploading

vytis:
This continues on work done in #292 by adding logic for uploading resized images to asset store.

### DIFF
--- a/Tests/Source/Synchronization/Strategies/UserImageAssetUploadStrategyTests.swift
+++ b/Tests/Source/Synchronization/Strategies/UserImageAssetUploadStrategyTests.swift
@@ -115,14 +115,22 @@ class UserImageAssetUploadStrategyTests : MessagingTest {
         updateStatus.dataToConsume[.preview] = "Some".data(using: .utf8)
         
         // THEN
-        XCTAssertNotNil(sut.nextRequest())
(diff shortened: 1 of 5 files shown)


Tests/Source/Synchronization/Strategies/UserImageAssetUploadStrategyTests.swift
```diff
@@ -0,0 +1,180 @@
...
+        
+        // THEN
+        XCTAssertNotNil(sut.nextRequest())
+        
+        // WHEN
+        updateStatus.dataToConsume.removeAll()
+        updateStatus.dataToConsume[.complete] = "Other".data(using: .utf8)
+        
+        // THEN
+        XCTAssertNotNil(sut.nextRequest())
```
daehn: Would be nice to check the path and method of the request here

vytis: As I get the request from requestFactory I thought I will just test that it returns something and then in the next test (`testThatItCreatesRequestWithExpectedData`) check that it's the same as returned by the request factory. But then maybe worth adding a simple check here as well.

PR #309, opened 2017-03-15
Profile picture v3: Integrated request strategy

vytis:
* Added method that is available to UI from ZMUserSession for setting new profile image
* Added request strategy and upload status class to the operation loop
* Few tweaks for ObjC interoperability


Source/UserSession/UserProfileImageOwner.swift
```diff
@@ -52,7 +50,7 @@ public final class UserProfileImageOwner: NSObject, ZMImageOwner {
     }
     
     public func originalImageSize() -> CGSize {
-        return imageSize
+        return .zero
```
mikeger: What is the logic behind the change?

Source/UserSession/ZMUserSession.m
```diff
@@ -346,8 +347,8 @@ - (instancetype)initWithTransportSession:(ZMTransportSession *)session
             self.pingBackStatus = [[BackgroundAPNSPingBackStatus alloc] initWithSyncManagedObjectContext:syncManagedObjectContext
                                                                                   authenticationProvider:self.authenticationStatus];
 
-           self.callStateObserver = [[ZMCallStateObserver alloc] initWithLocalNotificationDispatcher:self.localNotificationDispatcher
-                                                                                managedObjectContext:syncManagedObjectContext];
+            self.callStateObserver = [[ZMCallStateObserver alloc] initWithLocalNotificationDispatcher:self.localNotificationDispatcher
```
mikeger: 👍

Source/UserSession/UserProfileImageOwner.swift
vytis: I don't get the size from UI so I would need to extract it from `NSData`.
I looked through the code and it seems relevant only for assets v2. It's used in https://github.com/wireapp/wire-ios-images/blob/develop/Sources/Image%20Processing/ZMAssetMetaDataEncoder.m. Maybe @daehn can comment more about it?

daehn: Yeah it looks like we do not upload the size with v3 anymore, however I am unsure if the preprocessor uses that information for other purposes.

PR #311, opened 2017-03-20
Remove ClientMessageTranscoder from sync states

daehn:
# What's in this PR?

* The `ClientMessageTranscoder` is not a transcoder anymore and thus can be moved to the strategies array of the `ZMSyncStrategy` and can be removed from the sync states as well.

### This PR depends on:
- [x] A new release of `wire-ios-message-strategy` after https://github.com/wireapp/wire-ios-message-strategy/pull/49 has been merged.


marcoconti83:
Did you run all tests? I think some state machine tests might fail as they check with transcoders are invoked in each state.

daehn:
All ✅ (well expect the `TypingTranscoder` of course)

PR #312, opened 2017-03-21
Mark connected users to be re-downloaded in HotFix

daehn:
# What's in this PR?

* We need to refetch all connected users as they might alreday have uploaded profile pictures using the `/assets/v3` endpoint which we do not yet have locally and couldn't download before we updated to the latest version (The update event has already been processed so the client won't know about the new picture). The `selfUser` might have uploaded a picture using the new endpoint form a different client as well in the meantime and we need to re-download it. This might lead to a race condition, e.g. we want to migrate the profile picture from the old endpoint to the new one if the user does not yet have a v3 asset, but we need to ensure that we re-download the `selfUser` first, otherwise we might redundantly re-upload a picture that was already uploaded from a different client, or, worse upload an old profile picture that is not up-to-date anymore using the new endpoint.
* The same `HotFix` was used with the introduction of usernames, where the client needed to re-download all users to update their usernames which might have been updated in the meantime but could not be processed.

### DIFF
--- a/Source/Synchronization/ZMHotFixDirectory+Swift.swift
+++ b/Source/Synchronization/ZMHotFixDirectory+Swift.swift
@@ -91,9 +91,7 @@ extension ZMHotFixDirectory {
         }
     }
 
-    /// We need to refetch all connected users as they might alreday have updated their username (62.3.1) or uploaded profile
-    /// pictures using the /assets/v3 endpoint which we do not yet have locally and couldn't download before we updated to
-    /// a version supporting them. 
+    /// Marks all connected users (including self) to be refetched.
     /// Unconnected users are refreshed with a call to `refreshData` when information is displayed.
     /// See also the related `ZMUserSession.isPendingHotFixChanges` in `ZMHotFix+PendingChanges.swift`.
     public static func refetchConnectedUsers(_ context: NSManagedObjectContext) {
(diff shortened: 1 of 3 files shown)


Source/Synchronization/ZMHotFixDirectory+Swift.swift
```diff
@@ -91,8 +91,11 @@ extension ZMHotFixDirectory {
         }
     }
 
-    /// We need to refetch all connected users as they might alreday have updated their username before we updated to
-    /// a version supporting them. Unconnected users are refreshed with a call to `refreshData` when information is displayed.
+    /// We need to refetch all connected users as they might alreday have updated their username (62.3.1) or uploaded profile
```
marcoconti83: I would actually move this comment to the call site of the function. This function will fetch all users, in terms of documenting the function, it does not matter why. 
We might end up calling this from many different hot fixes and the comment will just grow. Better to specify the "why" where we are calling this function instead.

daehn:
@marcoconti83 updated!

marcoconti83:
Cool, now it's good to merge. However, I would not merge this until downloading asset v3 is done, I would merge this as the last one. If we merge it earlier, and it gets into an internal release without downloading asset v3 yet, those users will never get the asset v3 picture.

daehn:
@marcoconti83 as this is an intermediate PR to the `feature/profile-picture-v3` bas branch it should be fine to merge (as the base branch should only be merged when the support to download is in as well).

PR #314, opened 2017-03-22
Re-upload existing image when user does not have v3 image yet

daehn:
# What's in this PR?

* Add function to re-upload existing user images without preprocessing them again.
* `UserProfileImageUpdateStatus` is a `ZMContextChangeTracker` now and will check if it needs to re-upload the image to v3 if the self user changes, does not need to be updated from the backend and if there are  no medium and small image data as well as no asset v3 identifiers in the database yet.
* This functionality is still missing integration tests (which in turn require `wire-ios-mocktransport` to handle the v3 upload).

### DIFF
--- a/Source/UserSession/UserProfileImageUpdateStatus.swift
+++ b/Source/UserSession/UserProfileImageUpdateStatus.swift
@@ -249,7 +249,7 @@ extension UserProfileImageUpdateStatus: UserProfileImageUpdateProtocol {
     }
 }
 
-// Called internally with existing image data to reupload to v3 (no proprocessing needed)
+// Called internally with existing image data to reupload to v3 (no preprocessing needed)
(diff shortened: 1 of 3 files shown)


Source/UserSession/UserProfileImageUpdateStatus.swift
```diff
@@ -237,12 +242,51 @@ extension UserProfileImageUpdateStatus {
...
 }
 
+// Called from the UI to update a v3 image
 extension UserProfileImageUpdateStatus: UserProfileImageUpdateProtocol {
     public func updateImage(imageData: Data) {
         setState(state: .preprocess(image: imageData))
     }
 }
 
+// Called internally with existing image data to reupload to v3 (no proprocessing needed)
```
marcoconti83: typo `proprocessing` -> `preprocessing`

PR #321, opened 2017-03-29
Fix liking a message from a push notification

daehn:
# What's in this PR?

* We did not return after the call to kick off the like action, which could result in the completion handler being called too early.
* More importantly we did not start a proper background task on the `ZMOperationLoop`.
* Shuffled around some properties and include some files in the public header to expose the operation loop and relevant methods to Swift.

### DIFF
--- a/Source/UserSession/ZMUserSession+Background.m
+++ b/Source/UserSession/ZMUserSession+Background.m
@@ -204,6 +204,7 @@ static NSString *ZMLogTag = @"Push";
     }
     if ([identifier isEqualToString:ZMMessageLikeAction]) {
         [self likeMessageForNotification:notification WithCompletionHandler:completionHandler];
+        return;
     }
     
     if (floor(NSFoundationVersionNumber) > NSFoundationVersionNumber_iOS_8_4) {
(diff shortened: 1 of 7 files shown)


Source/UserSession/ZMUserSession+Background.m
```diff
@@ -204,6 +204,7 @@ - (void)application:(id<ZMApplication>)application handleActionWithIdentifier:(N
     }
     if ([identifier isEqualToString:ZMMessageLikeAction]) {
         [self likeMessageForNotification:notification WithCompletionHandler:completionHandler];
+        return;
```
mikeger: Should it all be if-else?

daehn:
Still running all tests.

Source/UserSession/ZMUserSession+MessageLikeAction.swift
```diff
@@ -17,21 +17,35 @@
...
+            guard let `self` = self else { return }
+            switch result {
+            case .failed: fallthrough // We might want to display a local notification to indicate failure in the future
+            case .unavailable: log.error("Error performing background task liking from notification: \(result.rawValue)")
+            default: break
+            }
+
+            self.managedObjectContext.performGroupedBlock {
+                activity?.end()
+                completion?()
```
basine: Shouldn't that be called in the startBackGroundTask block?  Otherwise we end the activity before we have the result?

Source/UserSession/ZMUserSession+Background.m
daehn: @mikeger updated!

PR #323, opened 2017-03-29
Fixing test failures after mock transport update

vytis:
Mock transport returns phone and email only for self user. Few methods that check users would fail as it would compare the current data with the response and see that values are missing.


daehn:
@vytis is this supposed to go into `develop` and not into the profile picture v3 branch?

vytis:
Yeah to `develop` - this is not related to assetv3, just another change that was made to mocktransport

PR #328, opened 2017-04-03
Profile picture using assets V3

vytis:
This is the main branch where the intermediate PRs were merged to. All of the code has been reviewed separately.

Depends on:
- [ ] https://github.com/wireapp/wire-ios-data-model/pull/218


Source/UserSession/Search/SearchDirectoryUserIDTable.swift
```diff
@@ -0,0 +1,163 @@
...
+    override public var hashValue: Int {
+        return userId.hashValue ^ asset.hashValue
+    }
+}
+
+
+@objc public final class SearchDirectoryUserIDTable: NSObject {
+
+    private let isolation = DispatchQueue(label: "SearchDirectoryUserIDTable.isolation", attributes: .concurrent)
+    private var entries = [NSObject: Set<SearchUserAndAsset>]() // TODO: Use ZMSearchResultStore instead of NSObject
```
daehn: The issue here was that before we were using a mapping with the protocol `ZMSearchResultsStore` as key type, which is not feasible in Swift as members of this protocol are not guaranteed to be `Hashable` (`ZMSearchResultsStore` & `Hashable` did not work either unfortunately). 

Another thing to note with this mapping here is that the Objective-C implementation used an `NSMutableDictionary` and boxed the keys unretained in `NSValues`, the test checking this still passes however.

PR #330, opened 2017-04-05
Delete background states

typfel:
This intermediate PR delete the background states from the state machine, they are replaced by the `OperationStatus` which keep track of when the app is in `foreground`/`background` or doing a `call`, `background fetch` or `background task`.

Depends on:
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/23
- [x] https://github.com/wireapp/wire-ios-message-strategy/pull/51


Source/UserSession/SyncStatus.swift
```diff
@@ -17,8 +17,11 @@
 //
 
 
+
+
+
```
vytis: Quite a bit of whitespace

Source/Synchronization/ApplicationStatusDirectory.swift
```diff
@@ -0,0 +1,109 @@
...
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+        apnsConfirmationStatus.tearDown()
+        clientRegistrationStatus.tearDown()
+        clientUpdateStatus.tearDown()
+    }
+    
+    public var deliveryConfirmation: DeliveryConfirmationDelegate {
+        get {
+            return apnsConfirmationStatus
```
vytis: Ca you simply return stuff here?

Source/Synchronization/ApplicationStatusDirectory.swift
```diff
@@ -0,0 +1,109 @@
...
+    public let userProfileUpdateStatus : UserProfileUpdateStatus
+    public let clientRegistrationStatus : ZMClientRegistrationStatus
+    public let clientUpdateStatus : ClientUpdateStatus
+    public let pingBackStatus : BackgroundAPNSPingBackStatus
+    public let accountStatus : ZMAccountStatus
+    public let proxiedRequestStatus : ProxiedRequestsStatus
+    public let syncStatus : SyncStatus
+    public let operationStatus : OperationStatus
+    public let requestCancellation: ZMRequestCancellation
+
```
vytis: Does it make sense to add the `UserProfileImageUpdateStatus` introduced for asset v3 in here also?

typfel: This branch is behind `develop` quite a bit. But yes I suppose I should add it here.

Source/Synchronization/ApplicationStatusDirectory.swift
typfel: let's try

Source/UserSession/OperationStatus.swift
```diff
@@ -0,0 +1,168 @@
...
+        backgroundFetchTimer?.invalidate()
+        backgroundFetchTimer = nil
+    }
+    
+    public func finishBackgroundTask(withTaskResult result: BackgroundTaskResult) {
+        DispatchQueue.main.async {
+            self.backgroundTaskHandler?(result)
+            self.backgroundTaskHandler = nil
+        }
+        backgroundTaskTimer?.invalidate()
```
marcoconti83: Should we invert the order and put the dispatch at the end, as with this code the handler might fire before we invalidate the timer? Not sure if that's a problem, but potentially we could call the handler and the timer could expire before we have a chance to call `invalidate()`

zmessaging-cocoa.xcodeproj/xcshareddata/xcschemes/zmessaging-ios.xcscheme
```diff
@@ -113,8 +113,8 @@
          </EnvironmentVariable>
          <EnvironmentVariable
             key = "ZMLOG_TAGS"
-            value = "MockTransportRequests"
-            isEnabled = "NO">
+            value = "MockTransportRequests,Authentication,State machine"
+            isEnabled = "YES">
```
marcoconti83: This should be "NO"

zmessaging-cocoa.xcodeproj/xcshareddata/xcschemes/zmessaging-ios.xcscheme
```diff
@@ -23,7 +23,7 @@
          <BuildActionEntry
             buildForTesting = "YES"
             buildForRunning = "NO"
-            buildForProfiling = "NO"
+            buildForProfiling = "YES"
```
marcoconti83: Is this needed?

typfel: No I will revert all xcscheme changes

Source/UserSession/OperationStatus.swift
typfel: Yep let's do that it sounds safer.

Source/UserSession/SyncStatus.swift
```diff
@@ -124,28 +123,6 @@ extension SyncStatus {
 // MARK: Quick Sync
 extension SyncStatus {
     
-    public func didEnterForeground() {
```
basine: This is not needed anymore?

Source/UserSession/SyncStatus.swift
```diff
@@ -16,9 +16,8 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-
 @objc public enum SyncPhase : Int {
-    case idle, fetchingLastUpdateEventID, fetchingConnections, fetchingConversations, fetchingUsers, fetchingMissedEvents, done
```
basine: So when you initialize the class, you will always be in fetchingLastUpdateEventID?

Source/UserSession/SyncStatus.swift
typfel: The idea now is that synchronisation is not tied to being in the foreground since we might want to do it in the background while in a call. It will stop the synchronisation when going into the background since the websocket will be closed.

Source/UserSession/SyncStatus.swift
typfel: can also be `fetchingMissingEvents` if has a previous lastUpdateEventID.

typfel: I thought I could remove the `idle`state since the strategies them self will decide if they are allowed to sync.

Source/UserSession/OperationStatus.swift
```diff
@@ -0,0 +1,168 @@
...
+        RequestAvailableNotification.notifyNewRequestsAvailable(self)
+    }
+    
+    public func startBackgroundTask(withCompletionHandler completionHandler: @escaping BackgroundTaskHandler) {
+        startBackgroundTask(timeout: 30.0, withCompletionHandler: completionHandler)
+    }
+    
+    public func startBackgroundTask(timeout: TimeInterval, withCompletionHandler completionHandler: @escaping BackgroundTaskHandler) {
+        guard backgroundTaskHandler == nil, isInBackground else {
+            return completionHandler(.failed)
```
basine: How does it work if you e.g. reply to two push notifications under bad network?

typfel: Second will fail

Source/UserSession/OperationStatus.swift
```diff
@@ -0,0 +1,168 @@
...
+        backgroundTaskHandler = completionHandler
+        backgroundTaskTimer = Timer.scheduledTimer(timeInterval: timeout, target: self, selector: #selector(backgroundTaskTimeout), userInfo: nil, repeats: false)
+    }
+    
+    dynamic func backgroundFetchTimeout() {
+        finishBackgroundFetch(withFetchResult: .failed)
+    }
+    
+    dynamic func backgroundTaskTimeout() {
+        finishBackgroundTask(withTaskResult: .failed)
```
basine: I always wondered if we should also cancel all requests we are currently doing. This would be a different PR though.

Source/UserSession/SyncStatus.swift
basine: Yeah they should "know" if they are authenticated

Source/UserSession/SyncStatus.swift
basine: So you said, we stop the sync, don't we need to restart it at some point?

Source/UserSession/OperationStatus.swift
basine: Would it be better to have a queue of tasks then?

Source/UserSession/OperationStatus.swift
typfel: Might be a good yes. Currently OperationStatus has no knowledge of request strategies which is nice.

Source/UserSession/OperationStatus.swift
typfel: Yep, might need to introduce an `task identifier` to separate the completion handlers if keep multiple around.

typfel: Queue doesn't really work since the request strategies wouldn't now which one is allowed to run.

Source/UserSession/SyncStatus.swift
typfel: We still do when the websocket opens, no?

typfel:
I'll merge this know and merge develop into all branches. Support for multiple background tasks will come in a separate PR.

PR #331, opened 2017-04-05
Refactor WireCallCenter V3 for better testing

basine:


Source/Calling/AVSWrapper.swift
```diff
@@ -0,0 +1,108 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import Foundation
+import avs
+
+public protocol AVSWrapperType {
+    init(userId: UUID, clientId: String, observer: UnsafeMutableRawPointer?)
+    func getCallState(conversationId: UUID) -> CallState
```
typfel: I think `callState(conversationId:)` is a more swifty name

Source/Calling/WireCallCenterV3.swift
```diff
@@ -146,115 +127,168 @@ public enum CallState : Equatable {
...
+    guard let contextRef = contextRef, let convID = UUID(cString: conversationId) else { return }
+    let userID = UUID(cString: userId)
+    
+    let callCenter = Unmanaged<WireCallCenterV3>.fromOpaque(contextRef).takeUnretainedValue()
+    callCenter.handleCallState(callState: .terminating(reason: CallClosedReason(reason: reason)), conversationId: convID, userId: userID)
+}
+
+/// Handles sending call messages
+/// In order to be passed to C, this function needs to be global
+internal func SendCallHandler(token: UnsafeMutableRawPointer?, conversationId: UnsafePointer<Int8>?, userId: UnsafePointer<Int8>?, clientId: UnsafePointer<Int8>?, data: UnsafePointer<UInt8>?, dataLength: Int, contextRef: UnsafeMutableRawPointer?) -> Int32
```
typfel: Maybe would be clear what this does if we call it something like `SendCallMessageHandler`

Source/Calling/WireCallCenterV3.swift
```diff
@@ -264,7 +298,10 @@ private typealias WireCallMessageToken = UnsafeMutableRawPointer
     
     public weak var transport : WireCallCenterTransport? = nil
     
-    public private(set) var callingProtocol : CallingProtocol = .version2
+    public fileprivate(set) var callingProtocol : CallingProtocol = .version2
+    
+    var avsWrapper : AVSWrapperType!
+    weak var uiMOC : NSManagedObjectContext!
```
typfel: Does this really need to be weak? uiMOC doesn't have reference to the call center, or?

Source/Calling/WireCallCenterV3.swift
```diff
@@ -264,7 +298,10 @@ private typealias WireCallMessageToken = UnsafeMutableRawPointer
     
     public weak var transport : WireCallCenterTransport? = nil
     
-    public private(set) var callingProtocol : CallingProtocol = .version2
+    public fileprivate(set) var callingProtocol : CallingProtocol = .version2
+    
+    var avsWrapper : AVSWrapperType!
```
typfel: Can't we lose the `!`?

Source/Calling/WireCallCenterV3.swift
```diff
@@ -273,269 +310,107 @@ private typealias WireCallMessageToken = UnsafeMutableRawPointer
...
         let transformedData = Data(buffer: bytes)
         
-        transport?.send(data: transformedData, conversationId: UUID(uuidString: conversationId)!, userId: UUID(uuidString: userId)!, completionHandler: { status in
+        transport?.send(data: transformedData, conversationId: conversationId, userId: userId, completionHandler: { status in
             wcall_resp(Int32(status), "", token)
         })
         
+        let currentState = callState(conversationId: conversationId)
+        if let uiMOC = uiMOC {
+            currentState.postNotificationOnMain(conversationID: conversationId, userID: userId, uiMOC: uiMOC)
```
typfel: Why do we post a notification when AVS sends a call message? Chris was showing me some sound effect being played at the wrong time which was triggered by this.

Source/Calling/WireCallCenterV3.swift
```diff
@@ -273,269 +310,107 @@ private typealias WireCallMessageToken = UnsafeMutableRawPointer
...
     
-    private func send(token: WireCallMessageToken, conversationId: String, userId: String, clientId: String, data: UnsafePointer<UInt8>, dataLength: Int) -> Int32 {
+    fileprivate func send(token: WireCallMessageToken, conversationId: UUID, userId: UUID, clientId: String, data: UnsafePointer<UInt8>, dataLength: Int) -> Int32 {
         
         let bytes = UnsafeBufferPointer<UInt8>(start: data, count: dataLength)
         let transformedData = Data(buffer: bytes)
         
-        transport?.send(data: transformedData, conversationId: UUID(uuidString: conversationId)!, userId: UUID(uuidString: userId)!, completionHandler: { status in
+        transport?.send(data: transformedData, conversationId: conversationId, userId: userId, completionHandler: { status in
             wcall_resp(Int32(status), "", token)
```
typfel: shouldn't this call also be wrapped by the AVS wrapper?

Source/UserSession/CallStateObserver.swift
```diff
@@ -55,13 +55,15 @@ extension CallStateObserver : WireCallCenterCallStateObserver, WireCallCenterMis
     public func callCenterDidChange(callState: CallState, conversationId: UUID, userId: UUID?) {
         notifyIfWebsocketShouldBeOpen(forCallState: callState)
         
-        managedObjectContext.performGroupedBlock {
+        managedObjectContext.performGroupedBlock { [weak self] in
```
typfel: Is weak `self` necessary here since the block isn't escaping?

Source/Calling/WireCallCenterV3.swift
basine: Not really, the closures need a reference to self and self is not fully initialized because... avsWrapper is not - devils circle

Source/Calling/WireCallCenterV3.swift
basine: You implemented it that way in the first place ;) Afaik the sendMessage is called when we join or start a call. It might be enough to post it in the join or startCall method. The bug that Chris mentioned can be handled in the UI. We already keep track of the previous callState there and could just check if the callState actually changed and only then play the sound.

Source/Calling/WireCallCenterV3.swift
basine: yes, it would probably be better to do that

Source/Calling/WireCallCenterV3.swift
basine: Yeah you're right - I must have added it accidentally when refactoring.

Source/UserSession/CallStateObserver.swift
basine: No, not really, but it will also do no harm, will it?

typfel: From my point of view the harm is in people reading the code will think you have to do this when doing a `performGroupedBlock` and will copy it and spread uncertainty when it's really important to have a weak `self`

Source/Calling/WireCallCenterV3.swift
```diff
@@ -146,115 +127,168 @@ public enum CallState : Equatable {
...
+}
 
-public protocol WireCallCenterCBRCallObserver : class {
-    func callCenterCallIsCBR()
+/// Handles established calls
+/// In order to be passed to C, this function needs to be global
+internal func EstablishedCallHandler(conversationId: UnsafePointer<Int8>?, userId: UnsafePointer<Int8>?,contextRef: UnsafeMutableRawPointer?)
+{
+    guard let contextRef = contextRef, let convID = UUID(cString: conversationId), let userID = UUID(cString: userId) else { return }
+    if wcall_is_video_call(convID.transportString()) == 1 {
```
typfel: Could also be called through the wrapper?

basine: yep, just wrapped. Someone should write tests for that though ;)

Source/Calling/WireCallCenterV3.swift
```diff
@@ -264,278 +299,112 @@ private typealias WireCallMessageToken = UnsafeMutableRawPointer
...
-    private func send(token: WireCallMessageToken, conversationId: String, userId: String, clientId: String, data: UnsafePointer<UInt8>, dataLength: Int) -> Int32 {
+    fileprivate func send(token: WireCallMessageToken, conversationId: UUID, userId: UUID, clientId: String, data: UnsafePointer<UInt8>, dataLength: Int) -> Int32 {
         
         let bytes = UnsafeBufferPointer<UInt8>(start: data, count: dataLength)
         let transformedData = Data(buffer: bytes)
         
-        transport?.send(data: transformedData, conversationId: UUID(uuidString: conversationId)!, userId: UUID(uuidString: userId)!, completionHandler: { status in
-            wcall_resp(Int32(status), "", token)
+        transport?.send(data: transformedData, conversationId: conversationId, userId: userId, completionHandler: { status in
+            self.avsWrapper.handleResponse(httpStatus: status, reason: "", context: token)
```
typfel: Now I guess we should weakify `self` here.

basine: haha, yeah I thought of it and then forgot :D

Source/Calling/AVSWrapper.swift
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/// Wraps AVS calls for dependency injection and better testing
+public class AVSWrapper : AVSWrapperType {
+    
+    required public init(userId: UUID, clientId: String, observer: UnsafeMutableRawPointer?) {
+        let resultValue = wcall_init(
+            userId.transportString(),
+            clientId,
+            ReadyHandler,
```
daehn: Confused me a bit at first that the function names are capitalized here.

Source/Calling/WireCallCenterV3+Notifications.swift
```diff
@@ -0,0 +1,161 @@
...
+
+
+
+/// MARK - CBR observer
+
+public protocol WireCallCenterCBRCallObserver : class {
+    func callCenterCallIsCBR()
+}
+
+struct WireCallCenterCBRCallNotification {
```
daehn: Can this be a `SelfPostingNotification` as well?

Source/Calling/AVSWrapper.swift
typfel: Yes, maybe it more idiomatic to have lower case function names even for global functions.

"Swift 3 API guidelines state that "Names of types and protocols are UpperCamelCase. Everything else is lowerCamelCase."

https://swift.org/documentation/api-design-guidelines/

basine: I can change it in the group calling PR ;)

Source/Calling/WireCallCenterV3+Notifications.swift
basine: yes

PR #332, opened 2017-04-05
Profile picture v3: handle failures

vytis:
* Updated the implementation to be able to try uploading profile picture again after a failure.
* Updated tests and implementation to use correct MOCs which should hopefully make tests more reliable
* Removed Data Model added as a subproject

### DIFF
--- a/Source/UserSession/UserProfileImageUpdateStatus.swift
+++ b/Source/UserSession/UserProfileImageUpdateStatus.swift
@@ -141,16 +141,14 @@ public final class UserProfileImageUpdateStatus: NSObject {
 // MARK: Main state transitions
 extension UserProfileImageUpdateStatus {
     internal func setState(state newState: ProfileUpdateState) {
-        managedObjectContext.performGroupedBlock {
-
-            let currentState = self.state
-            guard currentState.canTransition(to: newState) else {
-                // Trying to transition to invalid state - ignore
-                return
-            }
-            self.state = newState
-            self.didTransition(from: currentState, to: newState)
+        let currentState = self.state
+        guard currentState.canTransition(to: newState) else {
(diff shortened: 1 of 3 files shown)


Source/UserSession/UserProfileImageUpdateStatus.swift
```diff
@@ -201,16 +200,14 @@ extension UserProfileImageUpdateStatus {
...
-        managedObjectContext.performGroupedBlock {
-            let currentState = self.imageState(for: imageSize)
-            guard currentState.canTransition(to: newState) else {
-                // Trying to transition to invalid state - ignore
-                return
-            }
-            
-            self.imageState[imageSize] = newState
-            self.didTransition(from: currentState, to: newState, for: imageSize)
+        let currentState = self.imageState(for: imageSize)
```
marcoconti83: The `performGroupedBlock` is gone now, is it safe to remove it? I see some `setState` that are not wrapped inside a `performedGoupedBlock`, e.g. in `startPreprocessing`

vytis: The assumption was that the methods that are called from other than `syncMOC` thread are wrapped in `performGroupedBlock`, but maybe this assumption is too weak. Are there any problems if we call `performGroupedBlock` from inside another block? Also is there a way to assert if we call this on a wrong thread?

marcoconti83: When you run the app, if the code uses Core Data somehow (accessing the user info or any object), it will assert if it's on the wrong thread.
It is generally safe to have nested `performGroupedBlock`, but there might be some side effect depending on the usage because you are now making the code asynchronous.

vytis:
@marcoconti83 added some comments to the implementation but left the implementation the same.

PR #333, opened 2017-04-06
Enable group calling V3

basine:
- [x] Add a switch for switching between V2 and V3
- [x] Use public avs version


Cartfile
```diff
@@ -1,3 +1,2 @@
 github "wireapp/wire-ios-message-strategy" ~> 32.0.0
-github "wireapp/avs-ios-binaries" == 3.3.93
-
+github "wireapp/avs-ios-binaries-appstore" == 3.4.33
```
typfel: This will break for anyone outside wire

Source/Calling/VoiceChannelV3.swift
```diff
@@ -81,31 +95,41 @@ public class VoiceChannelV3 : NSObject, CallProperties {
...
+              let remoteIdentifier = conversation.remoteIdentifier
+        else { return false }
+        
+        let isGroup = (conversation.conversationType == .group)
         var joined = false
         
         switch state {
-        case .incomingCall:
-            joined = WireCallCenterV3.activeInstance?.answerCall(conversationId: remoteIdentifier) ?? false
+        case .incomingCall, .incomingCallInactive:
```
typfel: Can a call be `incomingCallInactive` and degraded?

Cartfile
basine: Yes, it's still a WIP therefore the "do not merge"

Source/Calling/VoiceChannelV3.swift
basine: I suppose so? I haven't thought about the degradation cases yet.

typfel: No it can't see now :)

Source/Calling/WireCallCenterV3.swift
```diff
@@ -301,6 +344,19 @@ public typealias WireCallMessageToken = UnsafeMutableRawPointer
     
     public fileprivate(set) var callingProtocol : CallingProtocol = .version2
     
+    /// We keep a snapshot of all participants so that we can notify the UI when a user is connected or when the stereo sorting changes
+    fileprivate var participantSnapshots : [UUID : VoiceChannelParticipantV3Snapshot] = [:]
+    
+    /// If a user ignores a call in a group conversation, it is added to this list
+    /// The list is reset when the call is closed or a new call is started by the selfUser
+    fileprivate var ignoredConversations = Set<UUID>()
```
typfel: This keeps tracks of which conversations got a call state with `shouldRing == false` in order to be able to return the full `CallState` when calling `callState(conversationId:)`?

basine: Yes that, too. It's used when the user ignored a call, or when we get a `shouldRing == false` since the behaviour for the UI is the same.

Source/Calling/VoiceChannelV3.swift
basine: OK, checked - if the conversation is degraded and ignored, the `VoiceChannelState` will be `.incomingCallDegraded` - which makes sense. The conversationListIndicator however should still show `.inactiveCall` since this one is set checking the `CallState` which does not take security levels into account

Source/Calling/WireCallCenterV3.swift
typfel: When I wrote this I was just wondering why SE needs to keep track of this and not AVS

basine: AVS is not keeping track of this unfortunately :-/

typfel: Not saying we need to this immediately but maybe it's worth discussing with AVS if it makes sense for them to keep track of this state as well. One of the selling points of calling 3.0 was that SE shouldn't maintain any call state.

Source/Calling/ZMCallKitDelegate.m
```diff
@@ -362,7 +362,7 @@ - (void)requestStartCallInConversation:(ZMConversation *)conversation videoCall:
         }];
     }
     else {
-        [self endAllOngoingCallKitCallsExcept:conversation];
+        [self endAllOngoingCallKitCallsExcept:conversation voiceChannelState:state];
```
typfel: I would find it easier to under this logic if we would do:

`[self endAllOngoingCallKitCallsExcept:state == VoiceChannelV2StateNoActiveUsers ? nil : conversation];`

mikeger:
@basine @typfel let's merge it now and cry later, i.e. start testing it?

PR #335, opened 2017-04-07
Fix for profile picture not being set for v2

vytis:
We were missing a save to kick it off. Also added a bit more debug info to the request.


Source/UserSession/UserProfileImageUpdateStatus.swift
```diff
@@ -256,6 +256,7 @@ extension UserProfileImageUpdateStatus: UserProfileImageUpdateProtocol {
             uiMOC.performGroupedBlock {
                 let editableUser = ZMUser.selfUser(in: uiMOC) as ZMEditableUser
                 editableUser.originalProfileImageData = imageData
+                uiMOC.saveOrRollback()
```
typfel: So this is initiated by the UI? shouldnt the call which sets the new profile picture be wrapped in an `enqueueChanges` which would perform a save?/

vytis:
@typfel updated

PR #336, opened 2017-04-07
Fix checking the wrong return value 

daehn:
# What's in this PR?

* Fixes comparing the return value of some `wcall` C function to to the wrong value, from the documentation: 
>  Returns 0 if successful

* Do not append additional performed call system message for terminating v3 calls. A `ZMSystemMessage` of type `.missedCalll` is already appended by the SystemMessageCallObserver when the call is terminated (or times out).


daehn:
Running tests 🏃🏼…

PR #338, opened 2017-04-10
Fix: crash when unarchiving objects saved before framework rename

vytis:
For Swift classes `NSKeyedArchiver` encodes class names with format `moduleName.className`. In an unfortunate case that it can't find this class it throws an (ObjC) exception. 
We recently renamed the project to `WireSyncEngine` and this causes problems if we have saved sessions on disk.

In addition there was a bug in `SessionTracker` (mistyped key ` "iniatorID"` instead of ` "initiatorID"`) that prevented any sessions to be read from disk altogether.


daehn:
x-all-the-things-keypath-all-the-strings

PR #340, opened 2017-04-11
Fix leaks in TypingStrategy

daehn:
# What's in this PR?

* When debugging the flaky `TypingStrategyTests` I noticed that the amount of `TypingStrategies` being alive is the same as the amount of tests that have previously run.
* When running the whole test suite there are a lot of tests running before the `TypingStrategyTests`, which means there could be hundreds of leaked `TypingStrategy`s, as the `isTyping` request generation is time sensitive this could be the root cause of the flakiness (e.g. when the huge amount of leaked strategies receive the notification first then the actual sut in the tests might get it too late).
* The `TypingStrategy` was never torn down in tests. The problem was that we were using the `NotificationCenter` block API to add observers, but were not holding on to the observer tokens and deregistering them in `tearDown`. We were also passing a function pointer, which is nice syntactic sugar but also has the caveat that it captures `self` strongly in the case the function is an instance member. We now use the selector API.
* This also removes an unnecessary call to `RequestAvailableNotification.notifyNewRequestsAvailable(self)` in `addConversationForNextRequest`, as it is already called internally form within a `performGroupedBlock`.

### DIFF
--- a/Source/Synchronization/Strategies/TypingStrategy.swift
+++ b/Source/Synchronization/Strategies/TypingStrategy.swift
@@ -66,8 +66,8 @@ public class TypingStrategy : NSObject {
         self.clientRegistrationDelegate = clientRegistrationDelegate
         self.typing = typing ?? ZMTyping(userInterfaceManagedObjectContext: uiContext, syncManagedObjectContext: syncContext)
         super.init()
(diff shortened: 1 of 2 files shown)


daehn:
Shouldn't be merged until Circle-CI is finished and passed.

PR #341, opened 2017-04-11
Set start time before posting established notification

basine:


Source/Calling/WireCallCenterV3.swift
```diff
@@ -128,12 +128,12 @@ public enum CallState : Equatable {
         }
     }
     
-    func postNotificationOnMain(conversationID: UUID, userID: UUID?, uiMOC: NSManagedObjectContext, completion: ((Void) -> Void)? = nil){
+    func postNotificationOnMain(conversationID: UUID, userID: UUID?, uiMOC: NSManagedObjectContext, performBeforePost: ((Void) -> Void)? = nil){
```
typfel: I'd write out this label when using it since it's not obvious that this block is performed before posting the notification.

basine: @typfel what do you mean with writing it out?

typfel: writing the out `performBeforePost` label where this block is used in the established handler.

basine: @typfel like this?

Source/Calling/WireCallCenterV3.swift
```diff
@@ -347,9 +347,9 @@ public typealias WireCallMessageToken = UnsafeMutableRawPointer
 
         switch callState {
         case .established:
-            callState.postNotificationOnMain(conversationID: conversationId, userID: userId, uiMOC: uiMOC){
+            callState.postNotificationOnMain(conversationID: conversationId, userID: userId, uiMOC: uiMOC, performBeforePost: {
```
basine: @typfel like this?

typfel: Yes

typfel:
This PR will conflict with my PR where dispatch onto the main queue in the call handlers.

PR #342, opened 2017-04-11
Manually nil put observer since it won't deinit otherwise

typfel:


typfel:
Tests didn't hang at least

PR #345, opened 2017-04-11
We need to dispatch onto the main queue before we do anything in the AVS handlers

typfel:
This fixes an issue where the outgoing video was missing when accepting a call.

https://wearezeta.atlassian.net/browse/ZIOS-8400

### DIFF
--- a/Source/Calling/WireCallCenterV3.swift
+++ b/Source/Calling/WireCallCenterV3.swift
@@ -128,15 +128,6 @@ public enum CallState : Equatable {
         }
     }
     
-    func postNotificationOnMain(conversationID: UUID, userID: UUID?, uiMOC: NSManagedObjectContext, completion: ((Void) -> Void)? = nil){
-        uiMOC.performGroupedBlock {
-            WireCallCenterCallStateNotification(callState: self,
-                                                conversationId: conversationID,
-                                                userId: userID).post()
-            completion?()
-        }
-    }
-    
     func logState(){
         switch self {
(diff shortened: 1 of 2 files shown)


typfel:
This also fixes https://github.com/wireapp/wire-ios-sync-engine/pull/341

mikeger:
Before AVS used to synchronize everything on their own, changed with the new API?

typfel:
@mikeger AVS still marshals the calls to the correct thread when we mute the audio for example. But these are callback handlers and they called on the AVS thread, they expect us dispatch away immediately from that thread. Calling any AVS functions for that thread may result in a deadlock.

Source/Calling/WireCallCenterV3.swift
```diff
@@ -339,22 +359,21 @@ public typealias WireCallMessageToken = UnsafeMutableRawPointer
...
+            establishedDate = Date()
+            
+            if avsWrapper.isVideoCall(conversationId: conversationId) {
+                avsWrapper.setVideoSendActive(userId: conversationId, active: true)
             }
-        default:
-            callState.postNotificationOnMain(conversationID: conversationId, userID: userId, uiMOC: uiMOC)
         }
+    
+        callState.postNotificationOnMain(conversationID: conversationId, userID: userId, uiMOC: uiMOC)
```
basine: We wrap the posting of the notification in `postNotificationOnMain` in another performGroupedBlock. I guess that can go then as well.

PR #348, opened 2017-04-19
Refactor like notification action to use the new background task API

typfel:
Refactor the like action from a local notification to use the background task API which comes with the sync states refactoring.

Unfortunately I had re-write it in objective-c to avoid issues with objective-c bridging.


daehn:
https://teespring.com/shop/dinosaur_objectivec#pid=76&cid=5854&sid=front

PR #351, opened 2017-04-21
Always include local seach for connected users

mikeger:
- The bug is introduced due to the recent changes on the backend.
- We can safely assume that we can search locally for the connections.


daehn:
@mikeger are any tests affected but his change?

marcoconti83:
No tests?
image

mikeger:
@marcoconti83 no, thank you.
493a592d-e9c0-db34-5137-488ab854d476

mikeger:
@marcoconti83 Added the test.

PR #353, opened 2017-04-24
Attempt to simplify the SyncStatus

typfel:
This PR tries simplify the public API in the `SyncStatus` and reduce the state handling in the request strategies which perform the slow sync.


vytis:
Only single test failed - impressive!

typfel:
@vytis all tests passing

PR #354, opened 2017-04-25
Adding new iPad models to devices list

jenskasemets:


Source/Utility/Device.swift
```diff
@@ -74,6 +74,7 @@ extension UIDevice {
         case "iPad2,1", "iPad2,2", "iPad2,3", "iPad2,4":return "iPad 2"
         case "iPad3,1", "iPad3,2", "iPad3,3":           return "iPad 3"
         case "iPad3,4", "iPad3,5", "iPad3,6":           return "iPad 4"
+        case "iPad6,11", "iPad6,12":                    return "iPad 5"
```
daehn: I think it's only called `iPad` from now on?

mikeger: Yeah, let's call it "iPad (5th generation)" or iPad (2017) : https://support.apple.com/kb/SP751?locale=en_US

jenskasemets: You are right, will fix. 
Internal naming still uses numeric identificator:
https://www.theiphonewiki.com/wiki/Models

PR #355, opened 2017-04-25
use new AVS version

basine:
AVS added a new callEndReason which we can use to decide whether or not a call has been left or actually ended.

### DIFF
--- a/Source/Calling/WireCallCenterV3.swift
+++ b/Source/Calling/WireCallCenterV3.swift
@@ -43,6 +43,8 @@ public enum CallClosedReason : Int32 {
     case canceled
     /// Incoming call was answered on another device
     case anweredElsewhere
+    /// Call left by the selfUser but continues until everyone else leaves or AVS closes it
+    case stillOngoing
     /// Call was closed for an unknown reason. This is most likely a bug.
     case unknown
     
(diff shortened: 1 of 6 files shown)


Source/Calling/WireCallCenterV3.swift
```diff
@@ -504,31 +519,15 @@ public struct CallEvent {
...
     public func rejectCall(conversationId: UUID, isGroup: Bool) {
         avsWrapper.rejectCall(conversationId: conversationId, isGroup: isGroup)
         ignoredConversations.insert(conversationId)
-
-        if isGroup {
-            WireCallCenterCallStateNotification(callState: .incoming(video: false, shouldRing: false),
-                                                conversationId: conversationId,
-                                                userId: initiatorForCall(conversationId: conversationId) ?? selfUserId).post()
-        } else {
-            WireCallCenterCallStateNotification(callState: .terminating(reason: .canceled),
```
typfel: AVS changed their behaviour again on this? Previously they would not call the close handler when you reject a call.

typfel: Ok, read the chat history. This been changed like 3-4 times now

mikeger:
@basine some tests failed on Circle, real issue or the bug in tests?

Source/Calling/WireCallCenterV3.swift
basine: Exactly

PR #357, opened 2017-04-28
do not report calls to callKit in silenced conversations

basine:
**Motivation for this PR**
Incoming calls in muted conversations are not supposed to activate CallKit nor show the internal call UI.

**Also included in this PR:** 
If you are trying to answer a call with callKit and the the callController does not have a call for the conversation yet, the answer would fail with UnknownCallUUID error. 
Solution: When we receive this error we request starting a new call.

Likewise if we try to create a new call, but the callUUID already exists, it currently fails with callUUIDAlreadyExists error code. 
Solution: When we receive this error we request answering the existing call.


basine:
@typfel @mikeger  I looked into the option of creating a new incoming call and immediately accepting it. From a user perspective it feels a bit weird, because you briefly get the callKit overlay and then our internal overlay. Another issue is that iOS's missed call list will not include the missed call if the conversation is silenced, because we never report it to the system. It would be cool if Apple allowed adding silent missed calls.

PR #358, opened 2017-04-28
Remove APNS performance tracking, add pingback logs

daehn:
# What's in this PR?

* Adds a missing `isEqual` implementation to `EventsWithIdentifier`.
* Remove classes and logic used to track APNS performance.
* Add more logs to the `BackgroundAPNSPingbackStatus` with the tag `"Pingback"` in order to see when and which notification IDs are added to the list of Ids to be fetched. This will be used to debug request loops to the `/notifications` endpoint and will be enabled for internal build by default in https://github.com/wireapp/wire-ios/pull/980.


daehn:
@typfel please re-review, a missing `isEqual` implementation was added in dec5a0c as well.

typfel:
Is adding the missing isEqual implementation fixing a bug? Should we maybe have a test for it.

vytis:
@typfel adding some tests for it

daehn:
It's a case of "How did this ever work", as we are comparing two objects of this type in multiple places…

daehn:
@typfel @mikeger please re-review, would be good to get this into internal fast to get more meaningful logs.

PR #362, opened 2017-05-02
Insert missed call message when group call ends

basine:
Previously systemMessages for V3 calls were handled by two different objects:
`CallingSystemMessageGenerator` and `SystemMessageCallObserver` where one was repeating part of the other object's behaviour. Both were subscribing the changes in the callState, manipulating the conversation which can lead to unwanted effects due to race conditions.

The `SystemMessageCallObserver` is now only handling V2 calls. And a new object which is owned by the `CallStateObserver` is handling V3 calls to avoid race conditions. This currently leads to some code duplication, but also removes some ambiguity when converting between `CallState` and `VoiceChannelV2State` as well as `CallClosedReason` and `VoiceChannelV2CallEndReason`. As soon as we retire V2 calls, we can remove the duplicated code.


Tests/Source/Calling/CallSystemMessageGeneratorTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        selfUser.remoteIdentifier = UUID()
+        selfUserID = selfUser.remoteIdentifier
+        clientID = "foo"
+        
+        sut = WireSyncEngine.CallSystemMessageGenerator()
+        mockWireCallCenterV3 = WireCallCenterV3Mock(userId: selfUserID, clientId: clientID, uiMOC: uiMOC)
+    }
+    
+    override func tearDown() {
+        sut = nil
```
vytis: If tests still work fine, try to `nil` all the variables that were created in `setUp` to reduce memory usage.

Tests/Source/Calling/CallSystemMessageGeneratorTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        let messageCount = conversation.messages.count
+        
+        // when
+        sut.callCenterDidChange(callState: .outgoing, conversation: conversation, user: selfUser, timeStamp: nil)
+        sut.callCenterDidChange(callState: .established, conversation: conversation, user: selfUser, timeStamp: nil)
+        sut.callCenterDidChange(callState: .terminating(reason: .canceled), conversation: conversation, user: user, timeStamp: nil)
+        
+        // then
+        XCTAssertEqual(conversation.messages.count, messageCount+1)
+        if let message = conversation.messages.lastObject as? ZMSystemMessage {
```
daehn: I would add an `else` clause and fail the test in that case to ensure that the last message is not a system message as expected.

Tests/Source/Calling/SystemMessageCallObserverTests.swift
```diff
@@ -43,7 +43,7 @@ class SystemMessageCallObserverTests : MessagingTest {
         selfUserID = selfUser.remoteIdentifier
         clientID = "foo"
         
-        sut = WireSyncEngine.SystemMessageCallObserver(managedObjectContext:uiMOC)
+        sut = WireSyncEngine.SystemMessageCallObserverV2(managedObjectContext:uiMOC)
```
vytis: Can't comment on the correct line but good opportunity to `nil` all the things in `tearDown` of this test class as well

Source/Calling/SystemMessageCallObserver.swift
```diff
@@ -82,18 +89,70 @@ private let log = ZMSLog(tag: "Calling System Message")
...
+    var startDateByConversation = [ZMConversation: Date]()
+    var connectDateByConversation = [ZMConversation: Date]()
+
+    public func callCenterDidChange(callState: CallState, conversation: ZMConversation, user: ZMUser?, timeStamp: Date?) {
+        switch callState {
+        case .outgoing:
+            log.info("Setting call start date for \(conversation.displayName)")
+            startDateByConversation[conversation] = Date()
+            fallthrough
+        case .incoming(video: _, shouldRing: _):
```
vytis: `case .incoming:` is a bit shorted if you don't use any of the associated values.

mikeger:
🔥 🚒  Conflict

Tests/Source/Calling/CallSystemMessageGeneratorTests.swift
basine: The XCTAssertEqual in the line above would fail otherwise. However, it does not test the message type.

PR #361, opened 2017-05-02
Remove Clusterizer class

daehn:
# What's in this PR?

* This class has been moved to `wire-ios-data-model` in https://github.com/wireapp/wire-ios-data-model/pull/243.

### This PR depends on:
- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/243 has been merged.

PR #364, opened 2017-05-03
Track metrics JSON from AVS call metrics handler

daehn:
# What's in this PR?

* Track call metrics JSON from AVS using Localytics.


daehn:
@typfel yeah, we should merge https://github.com/wireapp/wire-ios-sync-engine/pull/361 first and then merge develop back into this branch here.

daehn:
@typfel updated!

PR #366, opened 2017-05-03
Post APNS about missed call when CallKit is enabled

basine:
If we miss a call we want to display a missed Call APNS even if we are using CallKit.


vytis:
CI failed for some reason, happened to me several times today... Restart should fix it.

PR #368, opened 2017-05-04
Refactor fetching the notification stream after receiving a notification

daehn:
# What's in this PR?

There were multiple issues related to fetching `/notifications` after receiving a push (and performing the ping back request to cancel the fallback alert push the backend will sent after 30 seconds). 

- This PR refactors the logic in `ZMMissingUpdateEventsTranscoder` and ensures we can handle a paginated notification stream in this case as well.
- We were seeing request loops to `/notifications?size=500&since={id}&cancel_fallback={id}`, sometimes with the same notification ids, even though the response was successful. One issue was that we were reseting the list paginator every time a new push arrived. There also were multiple ways of getting into an inconsistent state, in which the local `notificationEventsToCancel` were not in sync with the actual state of the ping back status.
- Another case has been added to `ZMPushPayloadResult`. The case `.needsMoreRequests` is currently only used in case we are fetching the notification stream in response to a push notification and do not want to call the fetch completion handler the OS hand over to us, yet. In this case we forward the events in `ZMOperationLoop+Background` and wait to be called with a `.success` in order to call the system handler.

**There are still a couple of things related to fetching the notification stream in the background which should be addressed separately:**

- In case the strategy needs to fetch multiple pages, which can happen if a client has been offline for a while, other strategies might already create requests. Currently the only scenario I can think of where this might happen is when sending delivery receipts. After fetching the first page (currently set to 500 update events), the `ClientMessageTranscoder` will start inserting and sending delivery confirmation messages. This can be an issue as the used session might not yet be up to date.
- While this PR also ensures that we do not call the push fetch completion handler the OS hands to us when fetching multiple pages, it does not address the issue that we might not wait until all delivery receipts have been sent. This is not a critical issue, but a a nice to have.


Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
```diff
@@ -254,29 +253,45 @@ - (void)processEvents:(NSArray<ZMUpdateEvent *> *)events
...
-        if (self.listPaginator.status != ZMSingleRequestInProgress) {
+        // We only reset the paginator if it is neither in progress nor has more pages to fetch.
+        if (self.listPaginator.status != ZMSingleRequestInProgress && !self.listPaginator.hasMoreToFetch) {
             [self.listPaginator resetFetching];
         }
-        
+
+        // We need to add the id before asking the list paginator for a request,
+        // as it will ask us for additional query items.
+        BOOL didSetNotificationID = NO;
```
daehn: The logic here is still very convoluted and I we should think about good ways to change that. The issue here is that we need to set the `notificationEventsToCancel` as they are used to create additional query items for the request. However we do not want to set them if no request will be created.

daehn: I will try to refactor the way we ask for additional query items to declutter this.

daehn:
@typfel I updated the `nextRequest` logic in c534919. We know check `hasMoreToFetch` which indicates if we will be able to generate a request.

Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
```diff
@@ -276,21 +276,14 @@ - (ZMTransportRequest *)nextRequestIfAllowed
 
         // We need to add the id before asking the list paginator for a request,
         // as it will ask us for additional query items.
-        BOOL didSetNotificationID = NO;
-        if (nil == self.notificationEventsToCancel && fetchingForAPNS) {
+        // Also we need to ensure that we will be able to generate a request (checking hasMoreToFetch),
+        // to avoid setting the notificationEventsToCancel when we're unable to create a request.
+        if (nil == self.notificationEventsToCancel && fetchingForAPNS && self.listPaginator.hasMoreToFetch) {
```
typfel: Should there be a `!` before `self.listPaginator.hasMoreToFetch`?

daehn: We need to ensure that we will create a request, which is not the case if `hasMoreToFetch` is `NO`.

daehn: The naming is a bit confusing, as the `hasMoreToFetch` property of the `ZMSimpleListRequestPaginatorSync` is not equivalent to the `"has_more"` flag of the paginated response (the property is also `YES` after resetting the sync to perform the initial request for example).

basine:
@daehn Regarding your concern that we might already send delivery receipts - this could be circumvented if we entered quicksync, since the message transcoders do not create requests during syncing in general. We would have to "remember" somehow that we are only syncing to fetch notifications though.

daehn:
Yeah, @typfel and I were discussing entering quick sync, but it's very much tied to be happening in the foreground with the web socket open at the moment.
Maybe we can add different background state for this, in general the `BackgroundAPNSPingbackStatus` already exposes this with its `state` property. As long as this property is `inProgress` no delivery receipts should be sent. It might make sense to extract sending delivery confirmations from the `ClientMessageTranscoder` into a separate strategy and inject this state to do this though.

PR #369, opened 2017-05-04
stop typing when typing in other conversation starts

basine:
This should handle missing calls from the UI when typing in a conversation ended, but isTyping wasn't called.


Cartfile.resolved
```diff
@@ -4,15 +4,15 @@ github "wireapp/libPhoneNumber-iOS" "0.9.2-swift3.1"
 github "wireapp/ocmock" "v3.3.2"
 github "wireapp/ono" "1.4.0"
 github "wireapp/protobuf-objc" "1.9.13"
+github "wireapp/wire-ios-system" "17.0.0"
```
typfel: Curious why the dependencies move around so much. Are you using an update to date version of Carthage?

basine: yes, I run `carthage update` once in a while to make sure all our tests are still running despite changes in other frameworks.

typfel: Yes, makes sense to do it but I was wondering why the dependency sort order seems to change and was thinking maybe it's because we run `carthage update` with different Carthage versions installed.

Source/Synchronization/Strategies/TypingStrategy.swift
```diff
@@ -36,22 +36,70 @@ public struct TypingEvent {
...
+    /// Adds the conversation to the "queue"
+    /// If `isTyping` is true, it turns all other conversation events to endTyping events
+    func addItem(conversationID: NSManagedObjectID, isTyping: Bool) {
+        if isTyping {
+            // end all previous typings
+            conversations.forEach {
+                conversations[$0.key] = false
+            }
+            unbalancedConversations.forEach {
+                conversations[$0] = false
```
typfel: Isn't this already set to false by the previous `forEach`?

basine: Not really, conversations grows when we receive a typing notification and shrinks when the request for the conversation is sent out. We only keep track of the last event. If you imagine this scenario:
1) startTyping in convA (add it to unbalaced and conversations) -> remove from conversations when request is sent, lastEvent is convA
2) endTyping in convB (due to misfunctioning of UI) -> remove from conversations when request is sent, lastEvent is convB
3) startTyping in convB -> convA is unbalanced, but not in the conversations dictionary anymore, because we already sent the request. We should still tell the other side convA stopped typing

There should never be more than 2 conversations in these collections anyway, therefore the extra loop wouldn't matter much.

typfel: Ok, I follow now. 👍

PR #370, opened 2017-05-08
Sanitise request paths for loop tracking, add tests

daehn:
# What's in this PR?

* Sanitise request paths for loop tracking (replace UUIDs and client IDs with `"{id}"`).
* Add tests for `RequestLoopAnalyticsTracker`.

### DIFF
--- a/Source/UserSession/ZMUserSession.m
+++ b/Source/UserSession/ZMUserSession.m
@@ -240,12 +240,10 @@ ZM_EMPTY_ASSERTING_INIT()
     
     RequestLoopAnalyticsTracker *tracker = [[RequestLoopAnalyticsTracker alloc] initWithAnalytics:analytics];
     session.requestLoopDetectionCallback = ^(NSString *path) {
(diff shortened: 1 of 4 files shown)


Source/UserSession/ZMUserSession.m
```diff
@@ -240,12 +240,10 @@ - (instancetype)initWithMediaManager:(id<AVSMediaManager>)mediaManager
     
     RequestLoopAnalyticsTracker *tracker = [[RequestLoopAnalyticsTracker alloc] initWithAnalytics:analytics];
     session.requestLoopDetectionCallback = ^(NSString *path) {
-        if ([path hasSuffix:@"/typing"]) {
+        // The tracker will return NO in case the path should be ignored.
+        if (! [tracker tagWithPath:path]) {
```
mikeger: Can tracker be nil if localytics is disabled?

daehn: No, the tracker will still be there, but will return `NO` in this case.

PR #375, opened 2017-05-09
Require email on second sign in

mikeger:
The backend is not sending an error for the second client creation when email and password are not set for that user: https://github.com/wearezeta/backend-issues/issues/553 

The idea is to use the same SE API that is indicating the special `ZMUserSessionNeedsToRegisterEmailToRegisterClient ` log in error. This is now generated not as the result of the BE request, but as the result of the internal check after fetching `/self`.

**Known limitation:**

We are fetching the `/self` every 30-something seconds. Then the sign in callback is called every time we receive the result of GET `/self`. UI is ready to handle this.

I did not figured out how to fix this, I guess we would have to save the previous reported error?


Source/UserSession/ZMAuthenticationStatus.m
```diff
@@ -452,9 +452,20 @@ - (ZMEmailCredentials *)emailCredentials
     return nil;
 }
 
-- (BOOL)registeredOnThisDeviceOnContext:(NSManagedObjectContext *)moc;
+@end
+
+
+
+@implementation NSManagedObjectContext (Registrtaion)
```
typfel: typo: `Registrtaion ` > Registration

mikeger: Fixed, thanks.

Source/UserSession/ZMClientRegistrationStatus.m
```diff
@@ -32,24 +32,28 @@
 
 static NSString *ZMLogTag ZM_UNUSED = @"Authentication";
 
+static BOOL overrideNoAddingEmailNecessary = NO;
```
basine: Is this really necessary? A partial mock could have done the job espescially since we call it from ObjC

Source/UserSession/ZMUserSession+Registration.m
```diff
@@ -51,7 +52,7 @@ - (void)updateFromRegistrationUser:(ZMCompleteRegistrationUser *)registrationUse
 @implementation ZMUserSession (Registration)
 
 - (BOOL)registeredOnThisDevice {
-    return [self.authenticationStatus registeredOnThisDeviceOnContext:self.managedObjectContext];
+    return [self.managedObjectContext isRegisteredOnThisDevice];
```
basine: Is this called from with the same managedObjectContext where you are also setting it?

mikeger:
@basine correct, we fetch `/self` and check if it has email or not, then decide how to proceed, instead of trying to insert the new client and process the result. Basically the logic implementation is moved to the client.

mikeger:
@basine @typfel updated PR for review.

mikeger:
@typfel @vytis @daehn I think this is good now, please review.

mikeger:
Let's wait for Sabine to re-review it next week.

PR #378, opened 2017-05-11
Expose callingProtocol on the VoiceChannel

typfel:
- prefix caller name with `(V2)` in callkit if we make/receive V2 calls.

https://wearezeta.atlassian.net/browse/ZIOS-8556


vytis:
Should this be available for development builds only?

typfel:
@vytis if it was easy do then yes otherwise it was ok to leave it in since most people shouldn't do V2 calls anymore

vytis:
Didn't realize it is harder on SE than in UI app, makes sense

PR #379, opened 2017-05-12
Add call snapshots

typfel:
We are adding call snapshots so that we don't have to query AVS about the previous
call state. This means we will no longer call the following methods:
```
wcall_get_state
wcall_is_video_call
wcall_iterate_state
```

The reason for this change is that the above mentioned methods do not return correct
values at all times. In addition to that AVS is not happy about how frequently we
call them atm.


Source/Calling/AVSWrapper.swift
```diff
@@ -22,16 +22,13 @@ import avs
 
 public protocol AVSWrapperType {
     init(userId: UUID, clientId: String, observer: UnsafeMutableRawPointer?)
-    func callState(conversationId: UUID) -> CallState
```
basine: I wouldn't necessarily remove that since the methods are still there and might be useful some time in the future.

Source/Calling/WireCallCenterV3.swift
```diff
@@ -449,40 +452,48 @@ public struct CallEvent {
...
+            callSnapshots[conversationId] = CallSnapshot(callState: callState, isVideo: video)
+            
+            participantSnapshots[conversationId] = VoiceChannelParticipantV3Snapshot(conversationId: conversationId,
+                                                                                     selfUserID: selfUserId,
+                                                                                     members: [CallMember(userId: userId!, audioEstablished: false)],
+                                                                                     initiator: userId)
+            break
+            
+        case .terminating(reason: let reason):
+            clearSnapshot(conversationId: conversationId)
```
basine: You don't want to clear the snapshot for every reason. e.g. when the user leaves the call but the call is still ongoing, we get an end reason .still_ongoing. We should only clear snapshots when we really know the call ended completely.

Source/Calling/AVSWrapper.swift
basine: but for the sake of not reverting all your changes, it's ok if you remove it ;)

typfel: I wanted to be sure that it wasn't still  used somewhere

Source/Calling/WireCallCenterV3.swift
typfel: This methods gets the `finalCallState` so the still-ongoing call state will never arrive here.

Tests/Source/Calling/WireCallCenterV3Tests.swift
```diff
@@ -328,21 +333,31 @@ extension WireCallCenterV3Tests {
...
         
         // when
-        sut.mockAVSCallState = .incoming(video: false, shouldRing: true)
+        WireSyncEngine.incomingCallHandler(conversationId: conversationIdRef, messageTime: 0, userId: userIdRef, isVideoCall: 0, shouldRing: 1, contextRef: context)
+        XCTAssert(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
         sut.rejectCall(conversationId: conversationId, isGroup: false)
         
         // then
-        XCTAssertNotEqual(sut.callState(conversationId: conversationId), .incoming(video: false, shouldRing: true))
+        XCTAssertNotEqual(sut.callState(conversationId: conversationId), .incoming(video: false, shouldRing: false))
```
typfel: @basine this was bug before I think.

Source/Calling/WireCallCenterV3.swift
basine: ah ok, you could just write `.terminating:` since you don't use reason, but it's ok like that.

Source/Calling/AVSWrapper.swift
daehn: @typfel you could mark it as `@available (*, unavailable)`

typfel: @daehn true, next time ;)

Tests/Source/Calling/WireCallCenterV3Tests.swift
basine: I think in a 1on1 that's actually what we wanted to do, but AVS now handles it the same way as group calls: You get a still_ongoing followed by a cancelled afaik.

typfel: Hmm ok, so we should show the call as still ongoing in the UI as well after ending a call?

basine: tbh I have no idea, but I don't see a reason why we shouldn't. The call times out after a while anyway.

basine:
@typfel As discussed, I would merge after the release.

Tests/Source/Calling/WireCallCenterV3Tests.swift
typfel: I guess bit it's a bit weird. You can't join the call again because it's in the process being shutdown...

basine: No it's not - it only ends when it times out on the other side. The caller does not get notified the receiver rejected his call.

typfel: Ah right, was thinking about close call. So I should revert this change then.

daehn:
@typfel @basine can this be merged now?

typfel:
@daehn @basine ah this thing, maybe it's better to wait until after the pilot release since we probably won't have time to fix any call issues.

mikeger:
@typfel should we merge this one now?

typfel:
@mikeger yes, know we should have some time to fix the fall out.

typfel:
@mikeger wait still need do some fixes here.

mikeger:
@basine let's merge this one?

PR #380, opened 2017-05-12
Number of notification produced has changed after the connection re-ordering was fixed

typfel:
depends on 
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/31


Tests/Source/Integration/ConversationsTests.m
```diff
@@ -1162,7 +1162,7 @@ - (void)testThatConversationGoesOnTopAfterARemoteUserAcceptsOurConnectionRequest
         XCTAssertEqual(note.deletedIndexes.count, 0u);
         XCTAssertEqual(note.insertedIndexes.count, 0u);
     }
-    XCTAssertEqual(updatesCount, 1);
+    XCTAssertGreaterThanOrEqual(updatesCount, 1);
```
typfel: This is flakey. See line: 1153

daehn: Can we get the diff before the action and after and assert on that?

typfel: How do you mean? If don't care about the number of notification updates we can also remove this line.

daehn: Yeah, I misunderstood what we're doing there

PR #386, opened 2017-05-22
Support for searching within a team

typfel:
This PR includes a re-write for the `ZMSearchDirectory` into swift. Not all functionality has been ported to swift since it's no longer used.

_deleted functionality:_
- deleting suggestions
- common connections
- address book search

_added functionality:_
- filter conversation by team
- search among team members
- support for canceling searches

_TODO:_
- ~~port address book search if needed~~
- caching of search results


Source/UserSession/Search/SearchRequest.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+    
+}
+
+public struct SearchRequest {
+    
+    let maxQueryLength = 200
+    
+    public init(query: String, searchOptions: SearchOptions, team: Team? = nil) {
+        self.query = query.truncated(at: maxQueryLength)
```
daehn: What happens if the passed `SearchOptions` do not include `teamMembers` but a non-nil `Team` is passed in here? Thinking about a way how this could be wither or, but might be enough to add a `precondition` as this should never happen if I'm not mistaken?

typfel: What's the harm in allowing it? Maybe we to have screenonly for searching for team conversations.

Source/UserSession/Search/SearchTask.swift
```diff
@@ -0,0 +1,157 @@
...
+    fileprivate var result : SearchResult = SearchResult(contacts: [], teamMembers: [], directory: [], conversations: [])
+    
+    public init(request: SearchRequest, context: NSManagedObjectContext, session: ZMUserSession) {
+        self.request = request
+        self.session = session
+        self.context = context
+    }
+    
+    func onResult(_ resultHandler : @escaping ResultHandler) {
+        resultHandlers.append(resultHandler)
```
typfel: I'm still thinking about what the life cycle of the result handlers should be. Do we require you to hold on to the task or will it retain it self until all the results have arrived.

Source/UserSession/Search/SearchRequest.swift
daehn: Yeah, I think there's no real harm, was just thinking that it might be weird to pass in `teamMembers` but no team to search on, but it will just return an empty array for the members result in that case.

Source/UserSession/Search/SearchTask.swift
daehn: I think if the caller retains it it should be fine, also enforces the caller to think about only having one active task at a time.

Source/UserSession/Search/SearchTask.swift
```diff
@@ -0,0 +1,157 @@
...
+extension SearchTask {
+    
+    func performRemoteSearch() {
+        guard request.searchOptions.contains(.directory) else { return }
+        
+        context.performGroupedBlock {
+            let request = self.searchRequestInDirectory(withQuery: self.request.query)
+            
+            request.add(ZMCompletionHandler(on: self.session.managedObjectContext, block: { (response) in
+                guard
```
daehn: Maybe `[weak self]` in the completion and task created handler.

daehn:
@typfel the old classes are not deleted yet?

typfel:
@daehn no, the plan is to delete them after I've integrated the UI so I can run the app while I'm doing the integration.

Source/Public/ZMUserSession.h
```diff
@@ -132,6 +132,10 @@ extern NSString * const ZMTransportRequestLoopNotificationName;
 /// The URL of the shared container that has been determinned using the passed in application group identifier
 @property (nonatomic, readonly) NSURL *sharedContainerURL;
 
+@property (nonatomic, readonly) ZMTransportSession *transportSession;
```
mikeger: Swift compatibility? Should we expose the private header instead as we did with other swifty stuff?

typfel: Exactly, yes I'll try that.

mikeger:
🔥 Conflict 🐜 small one

Source/UserSession/Search/SearchDirectory.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    deinit {
+        assert(isTornDown, "`tearDown` must be called before SearchDirectory is deinitialized")
+    }
+    
+    public init(userSession: ZMUserSession) {
+        self.userSession = userSession
+        self.searchContext =  NSManagedObjectContext.createSearchWithStore(at: userSession.storeURL)
+        
+        isTornDown = true
```
mikeger: Not sure I am following the logic, also seems that `isTornDown` is not really used.

Source/UserSession/Search/SearchRequest.swift
```diff
@@ -0,0 +1,56 @@
...
+        self.searchOptions = searchOptions
+        self.team = team
+    }
+    
+    var query : String = "" {
```
mikeger: Would it be better to have only `let` allowed in this structure? Reason is to reassure the immutability.

Source/UserSession/Search/SearchTask.swift
```diff
@@ -0,0 +1,162 @@
...
+        
+        conversations.forEach { (conversation) in
+            if matchingPredicate.evaluate(with: conversation) {
+                matching.append(conversation)
+            } else {
+                nonMatching.append(conversation)
+            }
+        }
+        
+        return matching + nonMatching
```
mikeger: I think we should not add the nonMatching to the result.

typfel: This code is ported from the old search directory. It's a work-around for sorting matches on `userDefinedName` above matching conversations without a `userDefinedName`

Source/UserSession/Search/SearchRequest.swift
typfel: Yes, was thinking initially it would be useful to mutate query but actually it's not :)

Source/UserSession/Search/SearchDirectory.swift
typfel: see `deinit`

Source/UserSession/Search/SearchTask.swift
mikeger: All right! May be would make sense to add the comment, so nobody else like me would get confused and `fix` it :)

Source/UserSession/Search/SearchDirectory.swift
typfel: Oh right :) nice catch

Source/UserSession/Search/SearchRequest.swift
```diff
@@ -44,13 +44,8 @@ public struct SearchRequest {
...
     
-    var query : String = "" {
-        didSet {
-            query = query.truncated(at: maxQueryLength)
-        }
-    }
-    
     var team : Team? = nil
-    var searchOptions: SearchOptions = []
+    let query : String
```
daehn: No truncation anymore?

typfel: happens when `query` i set in `init`

mikeger:
@daehn @typfel let's merge?

typfel:
@mikeger  waiting for tests

PR #388, opened 2017-05-24
API changes after mock transport update

vytis:
API changes got introduced by https://github.com/wireapp/wire-ios-mocktransport/pull/37


daehn:
@vytis why the `"DO NOT MERGE"`?

PR #389, opened 2017-05-24
Adjust order of slow sync steps

daehn:
# What's in this PR?

* Move the `.fetchingTeams` slow sync phase before fetching connections and conversations.
* Before it could happen that when a team conversation a user is guest in would be fetched, the team would be set as `needsToBeUpdatedFromBackend`, but then would be fetched in the `.fetchingTeams` sync phase anyways. This means that after completing the slow sync said team unnecessarily would have been refetched.


daehn:
I think https://github.com/wireapp/wire-ios-sync-engine/pull/388 should be merged first as there might be some conflicts

PR #390, opened 2017-05-24
Localisation for the release 2.37

jenskasemets:


Resources/ar.lproj/Push.stringsdict
```diff
@@ -197,7 +197,7 @@
         <key>push.notification.knock.group</key>
         <dict>
             <key>NSStringLocalizedFormatKey</key>
-            <string>%1$@ pinged%3$#@new_knocks@في %2$@</string>
+            <string>%1$@ رنّ%3$#@new_knocks@في %2$@</string>
```
basine: Not sure if that's just a display issue, but the string `%3$#@new_knocks@` needs to stay intact in order for this to work. Same for `%2$@`. We probably need to test this on the client.

Resources/ar.lproj/Push.stringsdict
```diff
@@ -213,7 +213,7 @@
         <key>push.notification.knock.group.noconversationname</key>
         <dict>
             <key>NSStringLocalizedFormatKey</key>
-            <string>%1$@ pinged%2$#@new_knocks@in a conversation</string>
+            <string>%1$@ رنّ%2$#@new_knocks@in a conversation</string>
```
basine: Same problem as above.

Resources/ar.lproj/Push.stringsdict
```diff
@@ -255,13 +255,13 @@
...
                 <key>other</key>
-                <string>%1$@ pings</string>
+                <string>%1$@ رنّات</string>
             </dict>
         </dict>
         <key>push.notification.knock.oneonone</key>
         <dict>
             <key>NSStringLocalizedFormatKey</key>
-            <string>%1$@ pinged%3$#@new_knocks@</string>
+            <string>%1$@ رنّ%3$#@new_knocks@</string>
```
basine: Same problem as above.

Resources/ar.lproj/Push.stringsdict
jenskasemets: Looks like this is just rendering issue when view shows it with RTL support. Order of the actual characters is correct.
I would suggest to merge it in and then test it ourselves first.

PR #395, opened 2017-05-26
Add isCompleted parameter to search result handler

typfel:
The `isCompleted`parameter can be used to update the UI after a search completes, for example to show a `no results found` message.


Source/UserSession/Search/SearchTask.swift
```diff
@@ -19,15 +19,15 @@
 import Foundation
 import WireUtilities
 
-public typealias ResultHandler = (_ result: SearchResult, _ isCompleted: Bool) -> Void
```
daehn: Could also be nested inside the `SearchTask` then it could be used like `SearchTask.ResultsHandler`

Source/UserSession/Search/SearchTask.swift
```diff
@@ -19,27 +19,30 @@
...
     }
     
+    /// Add a result handler
     public func onResult(_ resultHandler : @escaping ResultHandler) {
         resultHandlers.append(resultHandler)
     }
     
+    /// Cancel a previously started task
     public func cancel() {
         resultHandlers.removeAll()
```
basine: should tasksRemaining be reset to 0 here?

typfel: You are not supposed to re-start tasks but let's do it for  consistency :)

PR #397, opened 2017-05-26
Re-fetch the team on the right context before accessing members

typfel:


Source/UserSession/Search/SearchTask.swift
```diff
@@ -93,7 +93,14 @@ extension SearchTask {
     }
     
     func teamMembers(matchingQuery query : String) -> [Member] {
-        return request.team?.members(matchingQuery: query) ?? []
+        guard
+            let teamObjectID = request.team?.objectID,
+            let team = context.object(with: teamObjectID) as? Team
```
daehn: `object(with:)` will throw an exception in case there is no object with that id, if we want to be sure we can use `existingObject(with:)`, see https://developer.apple.com/reference/coredata/nsmanagedobjectcontext/1506197-object. But then again, this object should always be there, right?

typfel: Ah interesting.

basine: In line 112 you are also accessing `request.team` - I suppose this might crash too?

typfel: It doesn't seem to crash but good catch!

PR #399, opened 2017-05-29
Ensure we delete teams we did not receive a deletion event for

daehn:
# What's in this PR?

It can happen that a user is offline for longer than 4 weeks, in this case said user will miss events and we will perform a slow sync. In the case the user got removed from a team that the user already has locally in that period, we want to ensure that we delete that team from the client.
The `/teams` request will only return the teams the user is still a member in of course, so we fetch all local teams and mark them as `needsToBeUpdatedFromBackend`, we reset this flag for all teams we fetch during the slow sync, (we just fetched them so there is no need to do it again after the slow sync completed). After the slow sync we will try to fetch the deleted teams and receive a 4xx response and delete the team locally.

### DIFF
--- a/Tests/Source/Integration/TeamTests.swift
+++ b/Tests/Source/Integration/TeamTests.swift
@@ -319,8 +319,7 @@ extension TeamTests {
     func testThatYouCanRemoveAMemberFromATeamConversation_SelfIsMember(){
     
     }
-    
-    
+
     func testThatYouCanNotAddAMemberToATeamConversation_SelfIsGuest(){
         
     }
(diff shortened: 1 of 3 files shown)


Tests/Source/Integration/TeamTests.swift
```diff
@@ -319,13 +319,47 @@ extension TeamTests {
...
+            return nil
+        }
+
+        // When
+        XCTAssert(logInAndWaitForSyncToBeComplete())
+        XCTAssert(waitForEverythingToBeDone())
+
+        // Then
+        // Assert that the local team got deleted after trying to refetch it AFTER the slow sync was performed.
+        // FIXME: The team is not deleted on uiMOC at this point,
```
basine: @daehn does `isZombieObject` return true?

basine: Ah I guess not, because you just fetched it...

daehn: @basine yes, it does.

PR #405, opened 2017-05-31
Add a HotFix to enforce a slow sync when updating sync-engine to v88

daehn:
# What's in this PR?

* Users might receive team related update events while not being able to parse them because they didn't update yet. To work around this we now force a slow sync in a HotFix to refetch all teams and all (team) conversations which might have been added before updating.

### DIFF
--- a/Source/UserSession/SyncStatus.swift
+++ b/Source/UserSession/SyncStatus.swift
@@ -65,6 +65,16 @@
 
 private let zmLog = ZMSLog(tag: "SyncStatus")
 
+
+extension Notification.Name {
+
+    public static var ForceSlowSync = {
+        Notification.Name("restartSlowSyncNotificationName")
+    }()
+    
+}
+
+
 public class SyncStatus : NSObject {
 
     fileprivate var previousPhase : SyncPhase = .done
@@ -99,8 +109,13 @@ public class SyncStatus : NSObject {
         super.init()
         
         currentSyncPhase = hasPersistedLastEventID ? .fetchingMissedEvents : .fetchingLastUpdateEventID
-        
         self.syncStateDelegate.didStartSync()
+        NotificationCenter.default.addObserver(self, selector: #selector(forceSlowSync), name: .ForceSlowSync, object: nil)
+    }
+
+    public func forceSlowSync() {
+        currentSyncPhase = .fetchingLastUpdateEventID
+        syncStateDelegate.didStartSync()
     }
 }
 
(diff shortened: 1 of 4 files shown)


Source/UserSession/SyncStatus.swift
```diff
@@ -65,6 +65,16 @@
 
 private let zmLog = ZMSLog(tag: "SyncStatus")
 
+
+extension Notification.Name {
+
+    public static var ForceSlowSync = {
```
basine: actually a `static let` would work here, too.

Source/UserSession/SyncStatus.swift
```diff
@@ -99,8 +109,13 @@ public class SyncStatus : NSObject {
         super.init()
         
         currentSyncPhase = hasPersistedLastEventID ? .fetchingMissedEvents : .fetchingLastUpdateEventID
-        
         self.syncStateDelegate.didStartSync()
+        NotificationCenter.default.addObserver(self, selector: #selector(forceSlowSync), name: .ForceSlowSync, object: nil)
+    }
+
+    public func forceSlowSync() {
+        currentSyncPhase = .fetchingLastUpdateEventID
```
basine: I am a bit concerned that we might miss some events if we enter slowsync in this phase. Imagine we enter the background as soon as we call this. In the background we would not receive webSocket events. We might miss some events and then never try to fetch them again, because we fetch the lastUpdateEvent and only events newer than that. So maybe we should go back to the phase after `. fetchingLastUpdateEventID ` instead. Since it's an Int, we could just add 1 to the rawValue. Maybe a helper function nextPhase() on the enum might be good. Since we use it in line 126 as well.

daehn: Good catch, will update the logic.

daehn:
@mikeger @basine updated!

Source/UserSession/SyncStatus.swift
```diff
@@ -99,16 +119,22 @@ public class SyncStatus : NSObject {
         super.init()
         
         currentSyncPhase = hasPersistedLastEventID ? .fetchingMissedEvents : .fetchingLastUpdateEventID
-        
         self.syncStateDelegate.didStartSync()
+        NotificationCenter.default.addObserver(self, selector: #selector(forceSlowSync), name: .ForceSlowSync, object: nil)
     }
+
+    public func forceSlowSync() {
+        currentSyncPhase = SyncPhase.fetchingLastUpdateEventID.nextPhase!
```
daehn: I force unwrap here, but I would argue that it would be a programmer error if the initial phase does not have a `nextPhase`.

PR #407, opened 2017-06-01
Fetch selfUser at the end of slowSync

basine:
When fetching users during slowSync, the user payload does not contain an email address thereby deleting the selfUser's email address. During slowSync we should therefore also fetch the selfUser at the end.

Also changed: The slowSync strategies are now calling back the slowSyncStatus callBack with their respective phase - This way we avoid entering a wrong phase before the current phase has completed.

### DIFF
--- a/Source/Synchronization/Transcoders/ZMSelfStrategy.m
+++ b/Source/Synchronization/Transcoders/ZMSelfStrategy.m
@@ -51,6 +51,7 @@ NSTimeInterval ZMSelfStrategyPendingValidationRequestInterval = 5;
 @property (nonatomic) ZMSingleRequestSync *downstreamSelfUserSync;
 @property (nonatomic) NSPredicate *needsToBeUdpatedFromBackend;
 @property (nonatomic, weak) ZMClientRegistrationStatus *clientStatus;
+@property (nonatomic, weak) SyncStatus *syncStatus;
 
 @end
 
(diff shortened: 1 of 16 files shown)


Source/Synchronization/Transcoders/ZMSelfStrategy.m
```diff
@@ -332,6 +347,12 @@ - (void)didReceiveResponse:(ZMTransportResponse *)response forSingleRequest:(ZMS
                 self.timedDownstreamSync.timeInterval = 0;
             }
         }
+        
+        if (self.isSyncing) {
+            [syncStatus finishCurrentSyncPhaseWithPhase:self.expectedSyncPhase];
+        }
+    } else if (self.isSyncing) {
```
typfel: Should only do this if it's a permanent error?

typfel:
Let's also remove the fetching the self user in the `ZMUserTranscoder`

Source/UserSession/SyncStatus.swift
```diff
@@ -125,7 +129,8 @@ public class SyncStatus : NSObject {
 // MARK: Slow Sync
 extension SyncStatus {
     
-    public func finishCurrentSyncPhase() {
+    public func finishCurrentSyncPhase(phase : SyncPhase) {
+        guard phase == currentSyncPhase else { return }
```
typfel: Seems like it should assert here then because this a programmer error if it occurs.

basine: I would be careful to assert here, tbh... We don't have all tests covered and it might potentially crash a lot of clients.

typfel: asserts doesn't crash clients

basine: Maybe after this release when we don't force everyone into a slow sync.

vytis:
The tests are behaving bad lately - crashes without running all of them. Retrying seems to fix it...

PR #409, opened 2017-06-02
Fixed assert breaking tests

mikeger:
- Replaced the assert with `ZMLogError`.

Idea for the future improvement: 
Create the precondition method that checks the condition and fails the test but crash on production.

### DIFF
--- a/Source/Synchronization/Transcoders/ZMLastUpdateEventIDTranscoder.m
+++ b/Source/Synchronization/Transcoders/ZMLastUpdateEventIDTranscoder.m
@@ -130,7 +130,9 @@
     NOT_USED(sync);
     SyncStatus *status = self.syncStatus;
     if(response.payload == nil) {
-        [status failCurrentSyncPhaseWithPhase:self.expectedSyncPhase];
+        if (status.isSyncing) {
+            [status failCurrentSyncPhaseWithPhase:self.expectedSyncPhase];
+        }
         return;
     }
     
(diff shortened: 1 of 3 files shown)


typfel:
Most transcoders actually check if they are in the correct sync phase before finishing/failing it. `ZMLastUpdateEventIDTranscoder` doesn't do this and that's why it asserts.

typfel:
`line: 133` in `ZMLastUpdateEventIDTranscoder ` should we wrapped in a `if (self.isSyncing) {}` to prevent this.

mikeger:
@typfel ok, I would try to fix it that way then.

mikeger:
@typfel updated. Waiting for CircleCI.

mikeger:
Works!

Source/Synchronization/Transcoders/ZMLastUpdateEventIDTranscoder.m
```diff
@@ -139,7 +141,10 @@ - (void)didReceiveResponse:(ZMTransportResponse *)response forSingleRequest:(ZMS
         self.lastUpdateEventID = lastNotificationID;
         if (status.currentSyncPhase == self.expectedSyncPhase) {
             [status updateLastUpdateEventIDWithEventID:lastNotificationID];
-            [status finishCurrentSyncPhaseWithPhase:self.expectedSyncPhase];
+            
+            if (status.isSyncing) {
```
typfel: this is the same as `status.currentSyncPhase == self.expectedSyncPhase` in the outer `if` statement

PR #410, opened 2017-06-02
Add address book search support to SearchDirectory

typfel:
- delete old search directory implementation
- delete common contacts search
- update integration tests that relied on the old search directory


Source/UserSession/Search/ZMSearchResult+AddressBook.swift
```diff
@@ -21,70 +21,70 @@ import Contacts
...
-        
-        for user in users {
-            identifiersToContact.removeValue(forKey: user.addressBookEntry.localIdentifier!)
-        }
-        
-        return (users: users, nonMatchedContacts: Array(identifiersToContact.values))
-    }
-}
+//
+//extension ZMSearchResult {
```
vytis: Seems that this code was moved to another place?

typfel: Oops :) Forgot to delete

PR #414, opened 2017-06-15
do not call back automatically when acting on a missed call notification

basine:
In some cases when a call timed out AVS would notify us about the timeout the next time the app enters the foreground. We would then create a notification (while the applicationStatus is still .background). iOS would then call `application: didReceiveLocalNotification:` which is usually only called when the user either taps on a notification while the app is already running. This in turn causes an automatic callback, because we assume the user actively tapped on the notification. 

To avoid automatic callbacks, missed call notifications now only open the conversation - but do not call back automatically anymore unless the call is still ongoing.


daehn:
@basine @mikeger I kicked off a new patch release.

PR #416, opened 2017-06-16
Removing v2 profile picture upload

vytis:
V2 download is still needed, so only uploading part was removed.

Needs release with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/289


daehn:
@vytis can we bump data-model and merge this one?

vytis:
Currently doing the releases of all dependencies

vytis:
@daehn updated, good to merge after test finishes

PR #418, opened 2017-06-19
Return BackgroundAPNSPingBackStatus as NotificationFetchStatusProvider

daehn:
# What's in this PR?

* Update `ApplicationStatus` by adding the necessary `notificationFetchStatus`.
* Add `.duringNotificationStreamFetch` as configuration to strategies that need to create requests while the notification stream is fetched. 

### This PR depends on:
- [x] A new release of `wire-ios-message-strategy` after https://github.com/wireapp/wire-ios-message-strategy/pull/79 has been merged.


marcoconti83:
In which case you need to create requests while you are  downloading the notification stream?

PR #421, opened 2017-06-22
Fix some request loops, allow message sending

mikeger:
# Issue with past participants

Due to the change in the backend, the `/conversations/{id}` is permanently returning 404 for requests from the users that are not the part of the conversation any more.

Solution is to mark such conversations with `needsToBeUpdatedFromBackend = false`, same applies for users, teams and connections.

# Issue with multiple `UserClient`

Some users accumulated the big amount of duplicated* user clients in the databases due to the unknown bug. 

Solution is to merge the user clients for now and monitor if the issue would continue to happen in the future, since the unknown bug might be fixed by the previous attempts.

* Duplicated clients are the clients with the exact same user and remoteIdentifier inserted in the database multiple times.

# Remarks

- `DiskDatabaseTest` is added since the CoreData's `GROUP BY` is only working on the disk storage.

# Depends on

- [x] https://github.com/wireapp/wire-ios-message-strategy/pull/81
- [x] https://github.com/wireapp/wire-ios-data-model/pull/293
- [x] https://github.com/wireapp/wire-ios-data-model/pull/299


mikeger:
@basine good point, would add!

BTW currently blocked by changes in https://github.com/wireapp/wire-ios-sync-engine/pull/418

mikeger:
@basine updated, please re-review!

mikeger:
Tests are breaking, needs some fixing ❌

mikeger:
@basine @daehn please re-review

Source/Synchronization/ZMHotFixDirectory+Swift.swift
```diff
@@ -110,4 +132,24 @@ extension ZMHotFixDirectory {
...
+        context.findDuplicated(by: #keyPath(UserClient.remoteIdentifier)).forEach { (remoteId: String?, clients: [UserClient]) in
+            // Group clients having the same remote identifiers by user
+            clients.filter { !($0.user?.isSelfUser ?? true) }.group(by: ZMUserClientUserKey).forEach { (user: ZMUser, clients: [UserClient]) in
+                guard let firstClient = clients.first, clients.count > 1 else {
+                    return
+                }
+                
+                let tail = clients.dropFirst()
+                // Merge clients having the same remote identifier and same user
+                let _ = tail.reduce(firstClient) {
```
daehn: I would avoid using `reduce` to perform side effects. I guess the same result could be achieved using a simple `for` loop?

daehn:
@mikeger conflicting 👻

mikeger:
Can be merged now.

WireSyncEngine.xcodeproj/project.pbxproj
```diff
@@ -2332,7 +2332,6 @@
 				5447E4631AECDC2E00411FCD /* ZMUserSessionRegistrationTests.m in Sources */,
 				543ED0011D79E0EE00A9CDF3 /* ApplicationMock.swift in Sources */,
 				878F63E51EF91E7C006E819B /* HotFixDuplicatesTests.swift in Sources */,
-				BF1F52C81ECC8720002FB553 /* TeamSyncRequestStrategyTests.swift in Sources */,
```
daehn: Not sure, this one should have been re-added with @basine 's latest changes.

mikeger: May be but check the final diff, I think it was added at some point but then removed.

PR #422, opened 2017-06-26
Fix CoreData multithreading violation when sending a reply from a notification

daehn:
# What's in this PR?

* We were passing a conversation from the UI context to the `LocalNotificationDispatcher`. The dispatcher internally only operates on the sync context.
* There is still another issue regarding the reply logic, we currently add a message observer to get notified whether the message was sent successfully or not. However, this observer does never fire at the moment, resulting in our background task timeout timer firing and finishing the background task with a result of `ZMBackgroundTaskResultFailed`. This bug is the cause for us always showing a `"Message failed to send"` notification when replying from a notification (this does not apply to liking from a notification as we are not showing any failure notifications in that case). I am still investigating the reason why the observer does not fire in this case, the fix will come in a separate PR.


daehn:
The second issue described in the PR description is being addressed here:
https://github.com/wireapp/wire-ios-data-model/pull/294
https://github.com/wireapp/wire-ios-sync-engine/pull/423

PR #423, opened 2017-06-26
Use new lightweight context change observer when reacting to a push

daehn:
# What's in this PR?

When replying to a message from a notification, users were always shown a `"Unable to send notification"` local notification which we display once the background task timer fires and we did not get notified that the message got sent, yet. We didn't get notified by the message observer because our observer system is not active when the application is in the background. As it would be quite some overhead to start up our observer system (creating snapshots from a lot of objects and so on) just to observe the delivery state of one message, we now use a lightweight wrapper around context change notifications and remove said observer as soon as the delivery state changed to delivered or sent.

### This PR depends on:
- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/294 has been merged.


daehn:
This needs to be a major release as dependencies were bumped in their major version.

PR #424, opened 2017-06-26
Move team strategies, add slowSync to team downstream strategy

basine:
do not merge yet
Needs

- [x] Release of dataModel with https://github.com/wireapp/wire-ios-data-model/pull/295
- [x] Release or MessageStrategy with https://github.com/wireapp/wire-ios-message-strategy/pull/82
- [x] Release of MockTransport with https://github.com/wireapp/wire-ios-mocktransport/pull/50


Source/Synchronization/Strategies/TeamDownloadRequestStrategy+Events.swift
```diff
@@ -0,0 +1,156 @@
...
+
+extension TeamDownloadRequestStrategy: ZMEventConsumer {
+
+    public func processEvents(_ events: [ZMUpdateEvent], liveEvents: Bool, prefetchResult: ZMFetchRequestBatchResult?) {
+        events.forEach(process)
+    }
+
+    private func process(_ event: ZMUpdateEvent) {
+        switch event.type {
+        case .teamCreate: createTeam(with: event)
```
daehn: Is this update event still there? AFAIK the only way to get added to a team is to directly create an account with that team specified now.

basine: I know that `.teamMemberLeave` is definitely still in action as it is sent if you are removed from the team. About the other ones, I am not quite sure.

PR #425, opened 2017-06-27
Do not refetch a team when receiving team update events

daehn:
# What's in this PR?

* Ignore `team.create` update events.
> With the new multi-account model this event should not be sent anymore, and if it is we should not act on it. An account will either have a team since registration or not, currently there is no way to get added to a team after registering.

* Do not create or refetch teams when receiving update events anymore.


daehn:
bildschirmfoto 2017-06-27 um 14 57 00

daehn:
There are some integration tests failing (slow sync), but as this is a PR to an intermediary branch and the slow sync logic is about to change again this seems fine.

Source/Synchronization/Strategies/TeamDownloadRequestStrategy+Events.swift
```diff
@@ -83,23 +82,15 @@ extension TeamDownloadRequestStrategy: ZMEventConsumer {
...
         _ = Member.getOrCreateMember(for: user, in: team, context: managedObjectContext)
 
-        // In case we just created this team locally or the self user got added,
-        // we want to ensure that we refetch all of its metadata, as well as all member permissions.
-        team.needsToBeUpdatedFromBackend = created || user.isSelfUser
-
         // cf. https://github.com/wireapp/architecture/issues/13
         // We want to refetch the members in case we didn't just create the team as the payload does not include permissions.
-        team.needsToRedownloadMembers = !team.needsToBeUpdatedFromBackend
+        team.needsToRedownloadMembers = true
```
basine: why do we set this to `true`?

Source/Synchronization/Strategies/TeamDownloadRequestStrategy+Events.swift
```diff
@@ -62,11 +62,10 @@ extension TeamDownloadRequestStrategy: ZMEventConsumer {
...
     private func createTeam(with event: ZMUpdateEvent) {
-        guard let identifier = event.teamId else { return }
-        guard let team = Team.fetchOrCreate(with: identifier, create: true, in: managedObjectContext, created: nil) else { return }
-        team.needsToBeUpdatedFromBackend = true
-        // FIXME: The backend started returning the team data in the response payload which we could parse
-        // to avoid having to refetch the team here (see https://github.com/wireapp/architecture/pull/26/files#diff-64b45f100a728fb2e5e24168a3c50de4R294).
+        // With the new multi-account model this event should not be sent anymore,
+        // and if it is we should not act on it.
+        // An account will either have a team since registration or not, 
+        // currently there is no way to get added to a team after registering.
```
basine: 👍🏻

Source/Synchronization/Strategies/TeamDownloadRequestStrategy+Events.swift
daehn: The `. teamMemberJoin` event does not contain the member permissions, which is why we always need to refetch the members from `/members`.

PR #427, opened 2017-06-29
fix payload parsing, bump components

basine:


Source/Synchronization/Strategies/TeamDownloadRequestStrategy.swift
```diff
@@ -71,7 +71,7 @@ extension TeamDownloadRequestStrategy: ZMDownstreamTranscoder {
             let team = object as? Team,
             let payload = response.payload?.asDictionary() as? [String: Any] else { return }
         
-        guard payload["binding"] as? Int == 1 else {
+        guard let isBound = payload["binding"] as? Bool, isBound == true else {
```
mikeger: Just to check, when binding = true we have to delete the team?

daehn: Then we should not even insert it in the first place. We should only delete a team if we receive a 4xx when trying to fetch it.

basine: When it's false it should be deleted. I added this in case someone for whatever reasons still has old teams lying around.

basine: But actually Marco never got back to me about whether or requests to /teams/{id} return the same payload

daehn: Or actually not even that, if we end up in the situation that we receive a 4xx when trying to fetch a team, then it is either our team, or we were added to a conversation with that team id and it does not match our local team id. If the latter is the case then we know we are a guest and should not even try to fetch a team (this is not implemented yet I think, but that's the desired behaviour in the future). So if we end up in the situation that we receive a 4xx for our local it means that the team got deleted our we got removed as a member and we should log out.

basine: OK, just checked, we do get the "binding" parameter.

PR #434, opened 2017-07-04
WIP: Unauthenticated session

vytis:
## Reason behind this PR

To support multi-accounts we need to support multiple CoreData stores. They will be stored in subfolders named by `user_id`. This poses a problem when logging in, because we don't know the `user_id` until after the login. The solution was to have registration and login flow to be independent of current persistence and delay loading it until the login/registration is completed.

## What's in this PR

The biggest change is the introduction of `UnauthenticatedSession` which has its own CoreData stack with in-memory store and a separate operation loop. It is used during registration/login until the full `ZMUserSession` is setup.

`ZMCookie` class was removed and `ZMPersistentCookieStorage` is injected into `ZMTransportSession` to support multiple cookies for the same server in the future.

**NOTE**
This is pointing to an integration branch `feature/registration-improvements`, so it's safe to merge after review


vytis:
To decrease the number of intermediate branches we merge this PR and will track all changes in a separate one

PR #438, opened 2017-07-04
Disk storage-free registration

vytis:
## Reason behind this PR

To support multi-accounts we need to support multiple CoreData stores. They will be stored in subfolders named by `user_id`. This poses a problem when logging in, because we don't know the `user_id` until after the login. The solution was to have registration and login flow to be independent of current persistence and delay loading it until the login/registration is completed.

## What's in this PR

The biggest change is the introduction of `UnauthenticatedSession` which has its own CoreData stack with in-memory store and a separate operation loop. It is used during registration/login until the full `ZMUserSession` is setup.

`ZMCookie` class was removed and `ZMPersistentCookieStorage` is injected into `ZMTransportSession` to support multiple cookies for the same server in the future.

We need releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/52
- [x] https://github.com/wireapp/wire-ios-data-model/pull/308


Source/Synchronization/Transcoders/ZMLastUpdateEventIDTranscoder.m
```diff
@@ -130,15 +130,13 @@ - (void)didReceiveResponse:(ZMTransportResponse *)response forSingleRequest:(ZMS
...
-        if (status.isSyncing) {
-            [status failCurrentSyncPhaseWithPhase:self.expectedSyncPhase];
-        }
-        return;
-    }
     
     NSUUID *lastNotificationID = [[response.payload asDictionary] optionalUuidForKey:@"id"];
-    if(lastNotificationID != nil) {
+    
+    if (response.HTTPStatus == 404 && status.currentSyncPhase == self.expectedSyncPhase) {
```
marcoconti83: Why did this change with this PR?

typfel:
So should we merge this?

vytis:
We probably need to merge all the other PRs before

PR #442, opened 2017-07-05
Fix phone registration tests

typfel:
Since the `ZMUserSession` has been replaced by the `SessionManager` as the entry point to SE we also need to update our integration tests to reflect that. Modifying the existing `IntegrationTestBase` was difficult since it inherits from `MessagingTest` which relies heavily on MOCs.

I therefore decided to create a `IntegrationTest` base class which tries to remain as close as possible to a real setup. The class is defined in obj-c but the implementation is in swift because obj-c classes can't inherit from swift classes.

### DIFF
--- a/Source/UnauthenticatedSession/UnauthenticatedOperationLoop.swift
+++ b/Source/UnauthenticatedSession/UnauthenticatedOperationLoop.swift
@@ -32,6 +32,20 @@ class UnauthenticatedOperationLoop: NSObject {
         super.init()
         RequestAvailableNotification.addObserver(self)
     }
+    
+    deinit {
+        for requestStrategy in requestStrategies {
+                        
+            if let requestStrategy = requestStrategy as? NSObject {
+                let tearDownSelector = Selector("tearDown") // TODO referer to a teardown protocol
+                
+                if requestStrategy.responds(to: tearDownSelector) {
+                    requestStrategy.perform(tearDownSelector)
+                }
+            }
+            
+        }
+    }
 }
 
 extension UnauthenticatedOperationLoop: RequestAvailableObserver {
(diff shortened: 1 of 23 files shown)


typfel:
Still working on tearing down everything correctly between tests

Tests/Source/Integration/PhoneRegistrationTests.m
```diff
@@ -244,11 +249,12 @@ - (void)testThatItLogsInWithAPhoneNumberIfThePhoneNumberIsAlreadyRegisteredToAno
     // expect
     [[self.authenticationObserver expect] loginCodeRequestDidSucceed];
     [[self.authenticationObserver expect] authenticationDidSucceed];
+    [[self.authenticationObserver expect] authenticationDidSucceed]; // client was registered
```
typfel: @vytis  been wondering about this one. So the change is that we now also call this on `ZMAuthenticationStatus loginSucceed` which why we get this twice. But we were anyway talking about separating authentication from "login".

Source/UnauthenticatedSession/UnauthenticatedOperationLoop.swift
```diff
@@ -32,15 +32,31 @@ class UnauthenticatedOperationLoop: NSObject {
         super.init()
         RequestAvailableNotification.addObserver(self)
     }
+    
+    deinit {
+        for requestStrategy in requestStrategies {
+                        
+            if let requestStrategy = requestStrategy as? NSObject {
+                let tearDownSelector = Selector("tearDown") // TODO referer to a teardown protocol
```
daehn: We could move the [`TearDownCapable` protocol](https://github.com/wireapp/wire-ios-share-engine/blob/dd07e1acaa7024ff7da6dff359fd4ba6fdd83eea/Sources/Conversation.swift#L46) to `wire-ios-message-strategy`.

Tests/Source/Integration/PhoneRegistrationTests.m
vytis: Yeah we should probably not call this method twice as it doesn't really make much sense to do it, let alone rely on it...

typfel:
@vytis whole suite passes in one go now.

vytis:
Wow that's amazing! 🏖 So we didn't break anything on the way then?

typfel:
I'm just talking about the PhoneRegistration tests so there's still a lot of things left to break ;)

vytis:
Ah I see, a bit too early to celebrate then:)

PR #446, opened 2017-07-11
Fix login integration tests

typfel:
Issues fixed while migrating tests:
* We didn't create an unauthenticated session if you got logged out.
* If your client got deleted we we will receive a event which logs you out. If you tried to log in again the first attempt would fail since we would again detect that your client was deleted.

### DIFF
--- a/Tests/Source/Integration/LoginFlowTests.m
+++ b/Tests/Source/Integration/LoginFlowTests.m
@@ -21,26 +21,21 @@
 @import WireDataModel;
 
 #import "NSError+ZMUserSessionInternal.h"
-#import "IntegrationTestBase.h"
-
 #import "ZMUserSession+Internal.h"
-#import "ZMUserSession+Authentication.h"
-#import "ZMUserSession+Registration.h"
 #import "ZMUserSession+OTR.h"
+#import <WireSyncEngine/WireSyncEngine-Swift.h>
+#import "WireSyncEngine_iOS_Tests-Swift.h"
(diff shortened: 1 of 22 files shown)


Tests/Source/Integration/LoginFlowTests.m
```diff
@@ -91,39 +73,34 @@ - (void)testThatItNotifiesIfTheClientNeedsToBeRegistered
...
-    
+    // expect
     id authenticationObserver = [OCMockObject mockForProtocol:@protocol(ZMAuthenticationObserver)];
-    [[[authenticationObserver expect] andDo:provideCredentials] authenticationDidFail:[NSError userSessionErrorWithErrorCode:ZMUserSessionNeedsCredentials userInfo:nil]];
-    id token = [self.userSession addAuthenticationObserver:authenticationObserver];
+    id token = [ZMUserSessionAuthenticationNotification addObserver:authenticationObserver];
     
-    XCTestExpectation *expectation = [self expectationWithDescription:@"Authentication did succeed"];
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Login did succeed"];
+    [[authenticationObserver expect] authenticationDidSucceed]; // authentication
```
daehn: Does it expect `authenticationDidSucceed` to be called twice now?

typfel: @daehn yes it does. there's also two expectations in this test.

PR #447, opened 2017-07-12
More tests are working

typfel:
issues found during migration:
* we lost a `!` during the re-factoring which caused us not to process pushes.

depends on:
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/54


Tests/Source/E2EE/UserClientRequestStrategyTests.swift
```diff
@@ -233,7 +233,7 @@ extension UserClientRequestStrategyTests {
         XCTAssertEqual(receivedAuthenticationNotifications.count, 1, "should only receive one notification")
         let note = receivedAuthenticationNotifications.first
         AssertOptionalNotNil(note, "Authentication should fail. Observers should be notified") { note in
-            XCTAssertEqual(note.error as NSError, expectedError)
+            XCTAssertEqual(note.error! as NSError, expectedError)
```
daehn: Could also be `XCTAssertEqual(note.error as NSError?, expectedError)` to avoid force unwrapping.

Tests/Source/Integration/IntegrationTestBase+Search.swift
```diff
@@ -115,3 +116,101 @@ extension IntegrationTestBase {
     }
     
 }
+
+//
+
+extension IntegrationTest {
+    
+    @objc
+    public func searchAndConnectToUser(withName name: String, searchQuery: String) {
```
vytis: would be good to add `line` and  `file` parameters and forward them to XCTAsserts to get more meaningful error messages

typfel: Does that work when used from obj-c?

typfel: I tried and It wouldn't compile

vytis: Default parameters don't work, but maybe there is a way to add a wrapper of some sort as there is `__FILE__` and `__LINE__` in ObjC

daehn:
Many properties of the integration base class changed their optionality and became optional, maybe it makes sense to mark them as implicitly unwrapped optionals instead of having to unwrap them anywhere else?

typfel:
@daehn yes, `mockTransportSession` we can assume is always there. I will mark it as implicitly unwrapped.

typfel:
@daehn I'll try this in a follow up PR since this has already grown too big.

Tests/Source/Integration/IntegrationTestBase+Search.swift
typfel: I'll try this out in a follow up PR.

PR #457, opened 2017-07-14
Fix TextSearchTests

daehn:
# What's in this PR?

* Fix `TextSearchTests`.


typfel:
was also fixed in https://github.com/wireapp/wire-ios-sync-engine/pull/456

daehn:
Closed in favour of #456

PR #465, opened 2017-07-17
Fix retain cycle in user profile observer

mikeger:
# Issue

Automatic retaining closure capture strikes the memory again.


Tests/Source/UserSession/UserProfileUpdateStatusTests.swift
```diff
@@ -57,6 +57,25 @@ class UserProfileUpdateStatusTests : MessagingTest {
     }
 }
 
+extension UserProfileUpdateStatusTests {
+    func testThatItDoesNotReatinObserver() {
```
daehn: Typo `reatin` → `retain`

PR #466, opened 2017-07-17
Move store and caches

basine:
This uses the new WireDataModel API that allows specifying an accountIdentifier for the location of the database. It also moved the caches on startup if this is needed. 
NOTE: This PR is probably prone to change to integrate the changes introduced with the SessionManager and AccountManager

TODOs:
- [x] needs release with https://github.com/wireapp/wire-ios-data-model/pull/317
- [x] should use WireMessageStrategy release with this fix https://github.com/wireapp/wire-ios-message-strategy/pull/89


Source/SessionManager/SessionManager.swift
```diff
@@ -61,7 +61,7 @@ public class SessionManager : NSObject {
                                                   cookieStorage: cookieStorage,
                                                   initialAccessToken: nil,
                                                   sharedContainerIdentifier: nil)
-        let localStoreProvider = LocalStoreProvider()
+        let localStoreProvider = LocalStoreProvider(userIdentifier: UUID()) // TODO: we need to provide real user id
```
daehn: This missing user id is an issue at the moment. We don't have it until we opened the database, but need it here in order to move the database to its new location (as the directory will be named after the user id).

PR #475, opened 2017-07-25
Initial sync fixes

typfel:
This PR add new property `hasCompletedInitialSync`. We currently have a similar property in the UI but it's unreliable since it's based on the `initialSyncCompleted` notification which doesn't guarantee the order in which the observers are executed.

Additionally we now also skip applying all hot fixes for on a fresh install and therefore avoids some unnecessary requests.

### DIFF
--- a/Source/Synchronization/Transcoders/ZMSelfStrategy.m
+++ b/Source/Synchronization/Transcoders/ZMSelfStrategy.m
@@ -299,6 +299,9 @@ NSTimeInterval ZMSelfStrategyPendingValidationRequestInterval = 5;
             }
         }
         
+        // Save to ensure self user is update to date when sync finishes
+        [self.managedObjectContext saveOrRollback];
+        
         if (self.isSyncing) {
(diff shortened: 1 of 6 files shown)


Source/Synchronization/Transcoders/ZMSelfStrategy.m
```diff
@@ -299,6 +299,9 @@ - (void)didReceiveResponse:(ZMTransportResponse *)response forSingleRequest:(ZMS
             }
         }
         
+        // Save to ensure self user is update to date when sync finishes
+        [self.managedObjectContext saveOrRollback];
```
typfel: I added this because I had the issue that `isPendingHotFixChanges` (`ZMUser.needsToBeUpdatedFromBackend`) was still true when the initial sync finishes since save haven't been done yet at that point. Maybe there's a better place to do this save?

typfel:
*checking tests*

typfel:
Not applying hotfixes if `lastSavedVersion` is missing could mean that we don't apply hotfixes if someone upgrades a two year old client.

PR #478, opened 2017-07-27
Fail login immediately when offline

mikeger:
# Issue

On automation it was noticed that the login error is not always appearing. The error is produced by the timer in `ZMAuthenticationStatus`. I checked the timer and it worked correctly.

The improvement proposed here is to fail the login if the device appears to be offline.

### DIFF
--- a/Tests/Source/UserSession/UnauthenticatedSessionTests.swift
+++ b/Tests/Source/UserSession/UnauthenticatedSessionTests.swift
@@ -1,0 +1,104 @@
+//
+// Wire
+// Copyright (C) 2017 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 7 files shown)


daehn:
@mikeger A test would be good!

mikeger:
@daehn I added some tests, more (for `UnauthenticatedSession`) to come.

Tests/Source/UserSession/UnauthenticatedSessionTests.swift
```diff
@@ -0,0 +1,104 @@
...
+    var sut: UnauthenticatedSession!
+    
+    public override func setUp() {
+        super.setUp()
+        transportSession = TestTransportSession()
+        sut = UnauthenticatedSession(transportSession: transportSession, delegate: nil)
+    }
+    
+    public override func tearDown() {
+        super.tearDown()
```
vytis: I think you are supposed to call this after the code in the method

PR #479, opened 2017-07-27
Fix core data multithreading assertion when setting context type on eventMOC

daehn:
# What's in this PR?

* Fix core data multithreading assertion when setting context type on eventMOC.


Source/Synchronization/NSManagedObjectContext+EventDecoder.swift
```diff
@@ -34,8 +34,9 @@ extension NSManagedObjectContext {
         let managedObjectContext = NSManagedObjectContext(concurrencyType: .privateQueueConcurrencyType)
         managedObjectContext.persistentStoreCoordinator = eventPersistentStoreCoordinator
         managedObjectContext.createDispatchGroups()
-        managedObjectContext.isEventMOC = true
-        
+        managedObjectContext.performGroupedBlock {
```
daehn: We don't need to wait here as this is the first block being enqueued and every subsequent one will have to dispatch onto the context's queue in order to access the value we set here.

PR #480, opened 2017-07-31
add search request for team user handle

johnxnguyen:
## Fix:
- allow search for team users directly with their handles


Source/UserSession/Search/SearchTask.swift
```diff
@@ -190,3 +191,77 @@ extension SearchTask {
...
+        if handle.hasPrefix("@") {
+            handle = handle.substring(from: handle.index(after: handle.startIndex))
+        }
+        
+        handle = handle.lowercased()
+        var queryCharacterSet = CharacterSet.urlQueryAllowed
+        queryCharacterSet.remove(charactersIn: "=&+")
+        let encodedQuery = handle.addingPercentEncoding(withAllowedCharacters: queryCharacterSet) ?? ""
+        
+        return ZMTransportRequest(getFromPath: "/users?handles=\(encodedQuery)")
```
mikeger: I think it would be cool to use [NSURLComponents](http://nshipster.com/nsurl/)

Source/UserSession/Search/SearchTask.swift
```diff
@@ -190,3 +191,77 @@ extension SearchTask {
...
+                    let name = userPayload["name"] as? String,
+                    let id = userPayload["id"] as? String
+                    else {
+                        return
+                }
+                
+                let document = ["handle": handle, "name": name, "id": id]
+                let documentPayload = ["documents": [document]]
+                guard let result = SearchResult(payload: documentPayload, query: query, userSession: session) else {
+                    return
```
mikeger: The `tasksRemaining` would not decrease in this case. I think it would be worth putting it in `defer` block then.

Source/UserSession/Search/SearchTask.swift
```diff
@@ -190,3 +191,77 @@ extension SearchTask {
...
+                }
+                
+                if let prevResult = self?.result {
+                    if !prevResult.directory.contains(result.directory.first!) {
```
mikeger: Should we also `if let` the `result.directory.first` just to be on the safe side?

Tests/Source/UserSession/SearchTaskTests.swift
```diff
@@ -47,6 +47,34 @@ class SearchTaskTests : MessagingTest {
         return conversation
     }
     
+    func testThatItFindsASingleUnconnectedUserByHandle() {
+        
+        // given
+        let remoteResultArrived = expectation(description: "received remote result")
+        
+        mockTransportSession.performRemoteChanges { (remoteChanges) in
+            let mockUser = remoteChanges.insertUser(withName: "Sponge Bob")
```
mikeger: I think it would make sense to give user the name that does not contain "bob", since the old search would also find him by the name.

Tests/Source/UserSession/SearchTaskTests.swift
```diff
@@ -47,6 +47,34 @@ class SearchTaskTests : MessagingTest {
         return conversation
     }
     
+    func testThatItFindsASingleUnconnectedUserByHandle() {
```
mikeger: Just an observation, is this a MockTransportSession / integration test? If no, then it's not doing any queries, so I have doubts if it is really testing the code path intended to be tested.

johnxnguyen: could you elaborate? My intention is to be able to test the single search request (per handle). Am I missing something?

mikeger: I think I am missing something, let me check :)

PR #488, opened 2017-08-09
Moving event store to user-specific folder

vytis:
On first launch it will try to copy existing store from the old location and initialize it in the user-specific folder.


Source/Synchronization/NSManagedObjectContext+EventDecoder.swift
```diff
@@ -29,17 +30,28 @@ extension NSManagedObjectContext {
...
         managedObjectContext.createDispatchGroups()
         managedObjectContext.performGroupedBlock {
             managedObjectContext.isEventMOC = true
         }
-        addPersistentStore(eventPersistentStoreCoordinator!, appGroupIdentifier: appGroupIdentifier)
+        addPersistentStore(eventPersistentStoreCoordinator!, withSharedContainerURL: sharedContainerURL, userIdentifier: userIdentifier)
         return managedObjectContext
     }
+    
+    fileprivate static func migrateOldStoreIfNeeded(oldStoreURL: URL, newStoreURL: URL) {
```
daehn: When reading this for the first time I thought this meant migrating to a new model version, maybe `moveOldStore` or ` relocateOldStore` makes more sense here.

vytis: makes sense, will tweak wording a bit

PR #497, opened 2017-08-16
Share reachability between both sessions

vytis:
This is needed so we would be consistent when the SyncEngine thinks the backend is reachable and what gets exposed to UI.

Depends on:
- [x] https://github.com/wireapp/wire-ios-transport/pull/54


Source/SessionManager/SessionFactories.swift
```diff
@@ -34,22 +34,23 @@ open class AuthenticatedSessionFactory {
...
         self.analytics = analytics
         self.apnsEnvironment = apnsEnvironment
         self.application = application
-        ZMBackendEnvironment.setupEnvironments()
-        self.environment = ZMBackendEnvironment(userDefaults: .standard)
+        self.environment = environment
     }
 
-    func session(for account: Account, storeProvider: LocalStoreProviderProtocol) -> ZMUserSession? {
+    func session(for account: Account, storeProvider: LocalStoreProviderProtocol, reachability: ReachabilityProvider) -> ZMUserSession? {
```
typfel: couldn't `reachability` be property on the factory so we don't need to pass it around as an argument?

vytis: Good idea, will make the API cleaner

PR #503, opened 2017-08-28
Fixed push notifications registration logic

johnxnguyen:
## What's in this PR?

The client was not registering for push notifications. Now it is :)


mikeger:
@johnxnguyen let's look on those tests today

PR #504, opened 2017-08-28
Observe file system access when creating session manager

johnxnguyen:
# What's in this PR?

## Problem
The `SessionManager` attempts to access the file system in order to retrieve account information. We must not attempt to access the file system before it is unlocked.

## Solution
The static `create` method should be called when in order to instantiate a `SessionManager` object. It will only return the instance once the file system is unlocked.


mikeger:
@johnxnguyen I think the tests failed here as well: the test target is not building

PR #506, opened 2017-08-29
Updating account details

vytis:
Added listening for user changes so that we could get the team name on initial launch. Team name changes are observed as well and updated accordingly. 
Currently the team picture is not downloaded here and is nil in account.

Needs release with:
- [x] https://github.com/wireapp/wire-ios-data-model/pull/354


Source/SessionManager/SessionManager.swift
```diff
@@ -327,6 +335,43 @@ public typealias LaunchOptions = [UIApplicationLaunchOptionsKey : Any]
...
+        guard let session = userSession else { return }
+        let selfUser = ZMUser.selfUser(in: session.syncManagedObjectContext)
+        if let account = accountManager.accounts.first(where: { $0.userIdentifier == selfUser.remoteIdentifier }) {
+            if let name = team?.name {
+                account.teamName = name
+            }
+            if let userName = userName {
+                account.userName = userName
+            }
+            accountManager.add(account)
```
typfel: Adding the `account` will update it if it already exists?

Source/SessionManager/SessionManager.swift
```diff
@@ -273,9 +275,12 @@ public typealias LaunchOptions = [UIApplicationLaunchOptionsKey : Any]
         guard let session = authenticatedSessionFactory.session(for: account, storeProvider: provider) else {
             preconditionFailure("Unable to create session for \(account)")
         }
+        let selfUser = ZMUser.selfUser(inUserSession: session)
+        teamObserver = TeamChangeInfo.add(observer: self, for: nil)
+        selfObserver = UserChangeInfo.add(observer: self, forBareUser: selfUser!)
```
typfel: Shouldn't we release the observers when logging out or switching accounts?

Source/SessionManager/SessionManager.swift
vytis: Yes, it is basically `createOrUpdate` under the hood, only named slightly confusing. Also has tests for update behaviour in DataModel.

Source/SessionManager/SessionManager.swift
vytis: Good point

vytis:
@typfel updated

mikeger:
@vytis we are not setting the `imageData` on the account yet? We could use the user image for non-team users.

typfel:
What about if the account already exists? I guess the name wouldn't get set until it's updated?

vytis:
@typfel I think it should update with the supplied name then - https://github.com/wireapp/wire-ios-data-model/blob/develop/Tests/Source/Model/AccountStoreTests.swift#L157

typfel:
@vytis I phrased it weirdly. I mean when the scenario when you upgrade and you already have a "synced" account. Then you wouldn't get any change notifications.

vytis:
@typfel update the code. The tests are lacking for now, but that needs a bit of refactoring to be able to inject dependencies.

typfel:
@mikeger tests were still failing

mikeger:
Sorry, I implied from comment that the tests are missing and not failing. From the log I see that the test task hanged. @vytis @typfel  should we revert?

PR #509, opened 2017-08-29
Fix threading and tests for account manager

mikeger:
- Issue due to the different sources of updates, now uses the MOC that is given by the update event.


Source/SessionManager/SessionManager.swift
```diff
@@ -405,18 +404,24 @@ extension SessionManager {
 extension SessionManager: TeamObserver {
     public func teamDidChange(_ changeInfo: TeamChangeInfo) {
         let team = changeInfo.team
-        updateCurrentAccount(with: team)
+        guard let managedObjectContext = (team as? Team)?.managedObjectContext else {
```
vytis: Aren't change notifications always posted on UI context? CC @typfel

typfel: Yes they are.

mikeger: May be we should decide from which thread it's allowed to change the accounts and assert in the setters? It looks like it is not thread-safe at the moment. Should we limit it to use only main queue? Or pass in the queue?

vytis: It's not core data entity, so not sure how we could assert inside. Probably would be best to add an isolation queue of some sort in the account manager to make reads/writes serial.

mikeger: I'll do it as a separate PR then. I think if we would assert the calls to be done from the same queue then the isolation queue is not necessary.

vytis: I think there is no documented and not deprecated way to get current queue though. In other places we assert based on which NSManagedObjectContext the current managed object uses, but since accounts are not persisted in CoreData we can't do it.

PR #510, opened 2017-08-30
Fixing leaks 

vytis:


marcoconti83:
I think you can kill the MemoryVC and that target

marcoconti83:
Sorry, not the target :D

vytis:
Yeah probably, not too useful anyway.

Source/Synchronization/ZMSyncStrategy.m
```diff
@@ -277,9 +277,24 @@ - (void)didInterruptUpdateEventsStream
...
+    self.callFlowRequestStrategy = nil;
+    self.connectionTranscoder = nil;
+    self.conversationTranscoder = nil;
+    self.eventsBuffer = nil;
+    self.userTranscoder = nil;
+    self.selfStrategy = nil;
+    self.clientMessageTranscoder = nil;
+    self.lastUpdateEventIDTranscoder = nil;
+    self.systemMessageEventConsumer = nil;
+    self.allChangeTrackers = nil;
```
mikeger: tumblr_mtaghz8ukr1r8oyaho1_500

Source/Synchronization/ZMOperationLoop.m
```diff
@@ -130,9 +130,10 @@ - (instancetype)initWithTransportSession:(ZMTransportSession *)transportSession
         
         [ZMRequestAvailableNotification addObserver:self];
         
+        NSManagedObjectContext *moc = self.syncMOC;
         // this is needed to avoid loading from syncMOC on the main queue
-        [self.syncMOC performGroupedBlock:^{
-            [self.transportSession configurePushChannelWithConsumer:self groupQueue:self.syncMOC];
+        [moc performGroupedBlock:^{
+            [self.transportSession configurePushChannelWithConsumer:self groupQueue:moc];
```
typfel: we are processing push channel events on the main context now?

vytis: It just fixes an error because `self.syncMOC` is weak - `moc` is a variable defined few lines above

typfel: Oh right 😳

circle.yml
```diff
@@ -1,8 +1,8 @@
 machine:
   xcode:
-    version: "8.3.2"
+    version: "8.3.3"
   environment:
-    DEPENDENCIES_BASE_URL: "https://raw.githubusercontent.com/wireapp/wire-ios-shared-resources/master"
+    DEPENDENCIES_BASE_URL: "https://raw.githubusercontent.com/wireapp/wire-ios-shared-resources/new-xcode"
```
typfel: new-xcode?

vytis: Reverted already. Update out CircleCI images to the latest Xcode and tried to test if it works on this repo!

marcoconti83:
Why the drop in tests? -14% sounds quite a lot

typfel:
@vytis so what's holding this back now? The tests are not passing with XCode 8.2?

vytis:
They are passing on the local machine, but leaks checking fail on CI. However the fail is pretty subtle - we have an `assert` that fires in between test cases so there is no test to fail. Xcode helpfully restarts the process and continues from the next test. Or sometimes doesn't restart... Still brainstorming with myself how to avoid this situation and make leaks more visible on CI.

vytis:
I will downgrade `wire-ios-testing` to the version that doesn't have the leak check and will try to fix the issue separately.
Should be good to merge after the tests pass.

typfel:
@vytis transport session is torn down twice now. Can probably remove: https://github.com/wireapp/wire-ios-sync-engine/blob/fix/leaks/Source/UserSession/ZMUserSession+Authentication.m#L83

vytis:
@typfel good point, will remove

PR #511, opened 2017-08-30
guard against registering for remote notifications when using simulator

johnxnguyen:
## Problem
When registering for remote notifications, we were not checking if the app was being run on the device, which has no push notifications.

## Solution
Before registering for notifications, we now check if the simulator is being used, if so, we log a message and return early.


johnxnguyen:
@mikeger the APNS tests are failing now because we dont register for notifications (since tests are run on the simulator)...

PR #512, opened 2017-08-31
Added account deletion

mikeger:
- Plus some fixes for threading during logout.


Source/SessionManager/SessionManager.swift
```diff
@@ -311,6 +311,13 @@ public typealias LaunchOptions = [UIApplicationLaunchOptionsKey : Any]
         }
     }
 
+    public func delete(account: Account) {
+        let accountID = account.userIdentifier
+        self.accountManager.remove(account)
+        
+        try! FileManager.default.removeItem(at: StorageStack.accountFolder(accountIdentifier: accountID, applicationContainer: sharedContainerURL))
```
typfel: Maybe log an error if it the `try` fails?

mikeger:
@typfel updated

PR #513, opened 2017-08-31
Switching to the right context to perform the teardown

marcoconti83:


typfel:
Why do we put the persistence coordinator in static variable in the EventDecoder?

marcoconti83:
I don't know, there is a "try again" logic here that the second time after a failure will use the static PSC. Needs investigation.

PR #498, opened 2017-08-17
Remove the V2 calling API and expose the V3 API directly

typfel:
### What changed?
The V2 API for calling was still around since a lot of code is built on top of it. This PR removes the V2 wrapper on top of V3 calling. Since the V3 API is based upon enums with associated values some code has to be ported to Swift.

### What's left
UI needs to be updated for the new API

PR #516, opened 2017-09-01
Support for AVS 3.6

typfel:
Update calling implementation for API changes in AVS 3.6.

in short:
- call config retrieval has moved from the flow manager to the`CallingRequestStrategy`
- new API for multi account support


Cartfile.resolved
```diff
@@ -13,6 +13,6 @@ github "wireapp/wire-ios-mocktransport" "49.1.0"
 github "wireapp/wire-ios-protos" "11.0.1"
 github "wireapp/wire-ios-request-strategy" "55.0.0"
 github "wireapp/wire-ios-system" "18.2.2"
-github "wireapp/wire-ios-testing" "12.0.0"
```
vytis: let's revert to `12.0.0`, test suite crashes because of leak check currently...

Source/Calling/AVSWrapper.swift
```diff
@@ -23,106 +23,122 @@ import avs
...
             let serverTime = UInt32(callEvent.serverTimestamp.timeIntervalSince1970)
             
-            wcall_recv_msg(bytes, callEvent.data.count, currentTime, serverTime, callEvent.conversationId.transportString(), callEvent.userId.transportString(), callEvent.clientId)
+            wcall_recv_msg(handle, bytes, callEvent.data.count, currentTime, serverTime, callEvent.conversationId.transportString(), callEvent.userId.transportString(), callEvent.clientId)
+        }
+    }
+    
+    public func update(callConfig: String?, httpStatusCode: Int) {
+        if let callConfig = callConfig {
+            wcall_config_update(handle, httpStatusCode == 200 ? 0 : EPROTO, callConfig)
```
vytis: I guess if let is not really needed here as the else part simply passes nil

vytis:
So we will still keep `V3` suffixes in class names?

Source/Calling/AVSWrapper.swift
typfel: True I will revisit this part. But there's was issue passing an optional String into the C-function.

typfel:
@vytis I'll remove the V3 prefixes when I have time to finish https://github.com/wireapp/wire-ios-sync-engine/pull/498

vytis:
@typfel ah true, got confused which PR does what.

PR #517, opened 2017-09-05
Migrate event decoder store to new folder structure

typfel:
As described in https://github.com/wireapp/wire-ios-data-model/pull/358
we are moving to new folder structure.

depends on:
- [x] https://github.com/wireapp/wire-ios-data-model/pull/358


Source/Synchronization/NSManagedObjectContext+EventDecoder.swift
```diff
@@ -29,23 +29,50 @@ extension NSManagedObjectContext {
     
     /// Creates and returns the `ManagedObjectContext` used for storing update events, ee `ZMEventModel`, `StorUpdateEvent` and `EventDecoder`.
     /// - parameter appGroupIdentifier: Optional identifier for a shared container group to be used to store the database,
-    /// if `nil` is passed a default of `group. + bundleIdentifier` will be used (e.g. when testing)
-    public static func createEventContext(withSharedContainerURL sharedContainerURL: URL, userIdentifier: UUID?) -> NSManagedObjectContext {
-        let previousStoreURL = storeURL(withSharedContainerURL: sharedContainerURL, userIdentifier: nil) // Passing no user identifier resolves to the old location before multiple accounts
-        let newStoreURL = storeURL(withSharedContainerURL: sharedContainerURL, userIdentifier: userIdentifier)
-        FileManager.default.createAndProtectDirectory(at: newStoreURL.deletingLastPathComponent())
-        relocateStoreIfNeeded(previousStoreURL: previousStoreURL, newStoreURL: newStoreURL)
+    /// if `nil` is passed a default of `group. + bundleIdentifier` will be used (e.g. when testing) TODO
```
mikeger: TODO for comment, or code should actually be changed too?

typfel: TODO for comment. accountIdentifier is no longer optional.

Source/Synchronization/NSManagedObjectContext+EventDecoder.swift
```diff
@@ -76,7 +103,24 @@ extension NSManagedObjectContext {
         return NSPersistentStoreCoordinator(managedObjectModel: mom)
     }
     
-    fileprivate static func addPersistentStore(_ psc: NSPersistentStoreCoordinator, withSharedContainerURL sharedContainerURL: URL, userIdentifier: UUID?, isSecondTry: Bool = false) {
+    fileprivate static func addPersistentStore(_ psc: NSPersistentStoreCoordinator, at location: URL, isSecondTry: Bool = false) {
```
vytis: Did we figure out the reason of having two tries here?

typfel: No we did not yet. Need to to do some git history investigation because code has moved between files.

daehn: @typfel @vytis IIRC when adding a persistent store without migrating we first try to open an existing one (which works in case there is no new model version / migration needed). If (for some reason) this operation fails, we remove the store from disk and create a new one, that's the second try (and the user will loose all of their data so this might be a relict from the non-e2ee times).

PR #519, opened 2017-09-06
Triggering authenticationDidSucceed also when registering with phone

vytis:
During the last refactoring we changed the behaviour and the notification is not called anymore when registering with phone number.

### DIFF
--- a/Source/Synchronization/Transcoders/ZMRegistrationTranscoder.m
+++ b/Source/Synchronization/Transcoders/ZMRegistrationTranscoder.m
@@ -148,11 +148,10 @@
     ZMAuthenticationStatus *authenticationStatus = self.authenticationStatus;
     if (response.result == ZMTransportResponseStatusSuccess) {
         BOOL shouldParseUserInfo = authenticationStatus.currentPhase == ZMAuthenticationPhaseRegisterWithPhone;
-        [authenticationStatus didCompleteRegistrationSuccessfully];
-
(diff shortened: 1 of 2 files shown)


Source/Synchronization/Transcoders/ZMRegistrationTranscoder.m
```diff
@@ -148,11 +148,10 @@ - (void)didReceiveResponse:(ZMTransportResponse *)response forSingleRequest:(ZMS
     ZMAuthenticationStatus *authenticationStatus = self.authenticationStatus;
     if (response.result == ZMTransportResponseStatusSuccess) {
         BOOL shouldParseUserInfo = authenticationStatus.currentPhase == ZMAuthenticationPhaseRegisterWithPhone;
-        [authenticationStatus didCompleteRegistrationSuccessfully];
-
         if (shouldParseUserInfo) {
             [self.userInfoParser parseUserInfoFromResponse:response];
```
typfel: This will trigger `authenticationDidSucceed`?

vytis: Yes, because inside `[authenticationStatus didCompleteRegistrationSuccessfully]` the `currentPhase` will be `authenticated`. This happens because we call `[self.userInfoParser parseUserInfoFromResponse:response]` before completing registration.
In the previous case we would set the cookie after, so the `currentPhase` was still registering with phone.

PR #521, opened 2017-09-06
Remove user image for account with team

mikeger:
- Not sure how the issue could happen, but for me the team account after some time got the image set, this should forbid it.

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -404,7 +404,7 @@ extension SessionManager {
             if let userName = selfUser.name {
                 account.userName = userName
             }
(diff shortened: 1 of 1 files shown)


typfel:
Maybe because we sync the self user and picture before the team?

mikeger:
@typfel this could be the case, then I guess I can check the team_id on self user.

mikeger:
@typfel we don't have the team id property any more, so I think checking if the team relation is there via `isTeamMember` is correct way to do it.

typfel:
Does that membership get created when the self user is synced? Didn't look like it in`updateWithTransportData:` in data model because then it as the same behaviour has before.

mikeger:
I checked now and we first fetch /self, then /teams, so it can happen that user has no team for some time, then this fix makes sense.

PR #523, opened 2017-09-07
Feature/block suspended account

johnxnguyen:
When attempting to sign in with a *suspended* account, backend will return the error code `403` with  label `suspended`. This PR adds the new error type and checks for it in the request response and sends the appropriate notification.


typfel:
Conflict!

🔥 🔥 🔥🚒 🚒 🚒

typfel:
@johnxnguyen a test perhaps in `LoginTranscoderTests`

johnxnguyen:
@typfel yeah I was just thinking about that. We need to make the account suspended, that would require changes to `MockTransport`?

typfel:
No not necessary these are not integration tests.

PR #526, opened 2017-09-07
Switch to second account if possible when deleting account

typfel:


Source/SessionManager/SessionManager.swift
```diff
@@ -292,6 +292,16 @@ public typealias LaunchOptions = [UIApplicationLaunchOptionsKey : Any]
         logoutCurrentSession(deleteCookie: false, error: NSError.userSessionErrorWith(.addAccountRequested, userInfo: nil))
     }
     
+    public func delete(account: Account) {
+        if let secondAccount = accountManager.accounts.first(where: { $0.userIdentifier != account.userIdentifier }) {
+            select(secondAccount)
+            deleteAccountData(for: account)
+        } else {
+            logoutCurrentSession(deleteCookie: true)
+            deleteAccountData(for: account)
```
mikeger: `deleteAccountData(for: account)` can be moved out of the if scope

PR #527, opened 2017-09-08
Use `ZMSingleRequestSync` to fetch the calls config

mikeger:
- Fixes the issue with automation of the calling tests.


Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -69,6 +71,32 @@ public final class CallingRequestStrategy : NSObject, RequestStrategy {
     
 }
 
+
+extension CallingRequestStrategy : ZMSingleRequestTranscoder {
+    public func request(for sync: ZMSingleRequestSync) -> ZMTransportRequest? {
+        if !self.callConfigFetched {
```
marcoconti83: Why is the bool needed? It will stop asking for requests once it completes successfully. Or do you need to request only once per application launch, even if AVS keeps asking you to fetch it?

mikeger: Updated

marcoconti83:
Assuming that: (AVS does not call this multiple times || when they do, it does not matter if you call the completion handler only for the last one), then it looks good now!

PR #529, opened 2017-09-13
Invalidate CallKit provider when tearing down a user session

typfel:
It wasn't possible to receive or make calls after switching accounts, this seems to be caused by us not invalidating the old CallKit provider.


mikeger:
When running tests: ```type 'MockCallKitProvider' does not conform to protocol 'CallKitProviderType'```

PR #530, opened 2017-09-15
Feature/unread count

johnxnguyen:
## What's in this PR?
- The session manager now listens for changes to `ConversationList`, finds the `Account` that this list belongs to, then updates the unread conversation count for this account. 

**DEPENDS ON:** https://github.com/wireapp/wire-ios-data-model/pull/361


mikeger:
🔥  Conflict 🚒

johnxnguyen:
build failing due to tests on the parent branch

PR #531, opened 2017-09-15
Multi account push

mikeger:
This PR has 2 logical parts

# Move the push handling logic out of Sync Engine to the level of the Session Manager

New `PushDispatcher` implements the logic for registering the push token and forwarding the registered token and the push notifications to multiple user sessions.

# Allow loading the multiple user sessions

The user sessions are loaded on demand and kept in memory. The memory warning triggers the deallocation of the background sessions.


Source/SessionManager/PushDispatcher.swift
```diff
@@ -0,0 +1,169 @@
...
+            if payload.isPayloadMissingUserInformation() {
+                self.callbackQueue.async {
+                    self.fallbackClient?.receivedPushNotification(with: payload, from: source, completion: completion)
+                }
+            }
+            else {
+                self.callbackQueue.async {
+                    let possibleHandlers = self.clients.filter { $0.mustHandle(payload: payload) }
+                    
+                    if let handler = possibleHandlers.last {
```
typfel: We can register multiple handlers per account or why is this a list?

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+    
+    fileprivate func wakeAllAccounts(for payload: [AnyHashable: Any],
+                                 from source: ZMPushNotficationType,
+                                 completion: ZMPushNotificationCompletionHandler?) {
+        log.error("Push is not specifict to account, fetching all")
+        self.accountManager.accounts.forEach { account in
+            self.withSession(for: account, perform: { userSession in
+                userSession.receivedPushNotification(with: payload, from: source, completion: completion)
```
typfel: If we can't identify to which account a push belongs we forward it to all accounts?

mikeger: Yes, it's to support the old pushes, not sure if relevant any more. Basically previously the backend was not specifying the user.

Source/SessionManager/PushDispatcher.swift
mikeger: Yep, it's a list, so you can add as many observers as you want, and the observer must say if it should handle the push for the payload, so this class has no logic for understanding if the push is for certain client. Clients are implementing the logic.

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -103,9 +105,40 @@ extension SessionManager {
...
-            self.withSession(for: account, perform: { userSession in
+            self.withSession(for: account) { userSession in
                 userSession.receivedPushNotification(with: payload, from: source, completion: completion)
-            })
+            }
+        }
+    }
+    
+    fileprivate func activateAccount(for session: ZMUserSession, completion: @escaping () -> ()) {
+        self.backgroundUserSessions.forEach { account, backgorundSession in
```
typfel: typo: `backgorundSession`

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -0,0 +1,179 @@
...
+        log.error("Push is not specifict to account, fetching all")
+        self.accountManager.accounts.forEach { account in
+            self.withSession(for: account) { userSession in
+                userSession.receivedPushNotification(with: payload, from: source, completion: completion)
+            }
+        }
+    }
+    
+    fileprivate func activateAccount(for session: ZMUserSession, completion: @escaping () -> ()) {
+        self.backgroundUserSessions.forEach { account, backgorundSession in
```
typfel: Isn't it possible that the background sessions have been unloaded in the time between that the push has arrived and the user taps the notifications?

Source/SessionManager/SessionManager.swift
```diff
@@ -376,6 +418,67 @@ public typealias LaunchOptions = [UIApplicationLaunchOptionsKey : Any]
...
+                migration: { [weak self] in self?.delegate?.sessionManagerWillStartMigratingLocalStore() },
+                completion: { provider in
+                    self.activateBackgroungSession(for: account, with: provider, completion: action)
+                }
+            )
+        }
+    }
+    
+    // Creates the user session for @c account given, calls @c completion when done.
+    fileprivate func activateBackgroungSession(for account: Account, with provider: LocalStoreProviderProtocol, completion: @escaping (ZMUserSession)->()) {
```
typfel: typo: `activateBackgroungSession`

typfel: This functions seems to duplicate a lot of from `createSession`. Should this function really notify the delegate (UI) about when it creates a session.

Source/SessionManager/SessionManager.swift
```diff
@@ -376,6 +418,67 @@ public typealias LaunchOptions = [UIApplicationLaunchOptionsKey : Any]
...
+        }
+        self.backgroundUserSessions[account] = newSession
+        newSession.requestToOpenViewDelegate = self
+        pushDispatcher.add(client: newSession)
+
+        log.debug("Created ZMUserSession for account \(String(describing: account.userName)) — \(account.userIdentifier)")
+        let authenticationStatus = unauthenticatedSession?.authenticationStatus
+        
+        newSession.syncManagedObjectContext.performGroupedBlock {
+            newSession.setEmailCredentials(authenticationStatus?.emailCredentials())
```
typfel: setting the `newSession.setEmailCredentials` is only necessary if want to register a new client.

Source/SessionManager/SessionManager.swift
```diff
@@ -376,6 +418,67 @@ public typealias LaunchOptions = [UIApplicationLaunchOptionsKey : Any]
...
+            if let registered = authenticationStatus?.completedRegistration {
+                newSession.syncManagedObjectContext.registeredOnThisDevice = registered
```
typfel: I think this can be removed since we don't call this function as a part of the registration.

Source/SessionManager/SessionManager+Push.swift
mikeger: It is unlikely to happen, since: we receive the user session as an argument, and it is just loaded from `self.withSession(for: account) { userSession ...` in `didReceiveLocalNotification` or `handleAction`. I would add an assert to check the case.

Source/SessionManager/SessionManager.swift
mikeger: Good point, I would remove it.

Source/SessionManager/SessionManager+Push.swift
typfel: Ah I see.

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -0,0 +1,185 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+private let log = ZMSLog(tag: "SessionManager")
+
+extension AccountManager {
+    func account(for id: UUID) -> Account? {
```
daehn: Could also expose and use https://github.com/wireapp/wire-ios-data-model/blob/develop/Source/Model/AccountManager.swift#L134

mikeger: Good point!

PR #543, opened 2017-09-22
Update for change in notification API

typfel:
context parameter is now more type safe since it must conform to NotificationContext protocol.


Tests/Source/UserSession/ZMClientRegistrationStatusTests.m
```diff
@@ -387,6 +387,7 @@ - (void)testThatItNotifiesTheUIAboutSuccessfulRegistration
     [self.sut didRegisterClient:client];
     WaitForAllGroupsToBeEmpty(0.5);
     
+    [self spinMainQueueWithTimeout:2];
```
marcoconti83: Debug leftover?

PR #554, opened 2017-09-25
Unread messages dot

nicorsm:
A dot will be shown on the badge of the user, if a new message is sent to any of the authenticated accounts.


marcoconti83:
Also remove setting the badge count, this is now done at the UI level.

PR #559, opened 2017-09-26
Request to open view can be executed when the delegate is nil

mikeger:


johnxnguyen:
Closing because this was implemented in UI

PR #570, opened 2017-09-27
Fixed possible multi session creation via push

mikeger:
Session loading is async, this makes it possible to accidentally load the session twice when the session loading event (push) is happening quickly one after another.

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -126,6 +126,9 @@ public protocol LocalMessageNotificationResponder : class {
     
     internal var authenticatedSessionFactory: AuthenticatedSessionFactory
     internal let unauthenticatedSessionFactory: UnauthenticatedSessionFactory
+    
+    fileprivate let sessionLoadingQueue : DispatchQueue = DispatchQueue(label: "sessionLoadingQueue")
+    
     fileprivate let sharedContainerURL: URL
     fileprivate let dispatchGroup: ZMSDispatchGroup?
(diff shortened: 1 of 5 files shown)


Source/SessionManager/SessionManager.swift
```diff
@@ -477,25 +480,31 @@ public protocol LocalMessageNotificationResponder : class {
...
-                userIdentifier: account.userIdentifier,
-                dispatchGroup: dispatchGroup,
-                migration: { [weak self] in self?.delegate?.sessionManagerWillStartMigratingLocalStore() },
-                completion: { provider in
-                    self.activateBackgroundSession(for: account, with: provider, completion: action)
-                }
-            )
-        }
+    public func withSession(for account: Account, perform completion: @escaping (ZMUserSession)->()) {
+        self.sessionLoadingQueue.serialAsync(do: { finally in
```
vytis: Not sure if I understand the layered callback structure. Wouldn't just serial dispatch queue work the same?

mikeger: If we would use the `plain` serial queue then the loading would not be executed on the main thread; we can have a serial queue that dispatches synchronously to the main queue the work block, you mean doing it this way?

Tests/Source/Utility/DispatchQueueSerialAsyncTests.swift
```diff
@@ -0,0 +1,57 @@
+//
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
```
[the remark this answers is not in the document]
typfel: 😱

PR #572, opened 2017-09-28
filter system messages for push/chathead notifications

johnxnguyen:
The only system message notifications that were being dispatched were concerning members joining/leaving group conversations. But we only want to be notified if someone adds/deletes us from a conversation.

Additionally, we no longer need to bundle these notifications since we are not collecting participation notifications from other members (bundled notifications lead to "People were added to the group")

### DIFF
--- a/Source/Notifications/Push notifications/LocalNotificationDispatcher+Messages.swift
+++ b/Source/Notifications/Push notifications/LocalNotificationDispatcher+Messages.swift
@@ -131,16 +131,26 @@ extension LocalNotificationDispatcher {
     
     fileprivate func localNotificationForSystemMessage(_ message : ZMSystemMessage) -> ZMLocalNotificationForSystemMessage? {
         
-        // We might want to "bundle" notifications, e.g. member join / leave events
-        if let newNote: ZMLocalNotificationForSystemMessage = messageNotifications.copyExistingMessageNotification(message) {
-            return newNote;
-        }
-        
+        // we only want participation messages concerning self user
+        guard message.isGroupParticipationMessageForSelf else { return nil }
+                
         if let newNote = ZMLocalNotificationForSystemMessage(message: message, application:self.application) {
             messageNotifications.addObject(newNote)
             return newNote;
         }
         return nil
+    }
+}
+
+private extension ZMSystemMessage {
+    
+    /// Returns true if the system message notifies the self user was added or removed
+    /// from a conversation.
+    ///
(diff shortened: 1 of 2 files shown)


Source/Notifications/Push notifications/LocalNotificationDispatcher+Messages.swift
```diff
@@ -131,11 +131,9 @@ extension LocalNotificationDispatcher {
     
     fileprivate func localNotificationForSystemMessage(_ message : ZMSystemMessage) -> ZMLocalNotificationForSystemMessage? {
         
-        // We might want to "bundle" notifications, e.g. member join / leave events
-        if let newNote: ZMLocalNotificationForSystemMessage = messageNotifications.copyExistingMessageNotification(message) {
```
mikeger: I guess bundling of the notifications is not necessary any more?

johnxnguyen: yeah that's right :)

Source/Notifications/Push notifications/LocalNotificationDispatcher+Messages.swift
```diff
@@ -131,11 +131,9 @@ extension LocalNotificationDispatcher {
...
-            return newNote;
-        }
-        
+        // we only want participation messages concerning self user
+        guard message.isGroupParticipationMessageForSelf else { return nil }
```
typfel: Wouldn't this remove all notifications for any system message which are not `participantsAdded`, or `participantsRemoved`, like `connectionUpdate` for example. Is that what we want?

johnxnguyen: ah you're right. Just looking at it now, the only other system message type that produces a notification with an an `alertBody` is `.connectionRequest` (all other types put an empty string in the alert body). I'll fix the code to include this type

typfel: Would change the name of the function then as well.

johnxnguyen: when is the connection request system message sent in a conversation? At the moment, if I receive a connection request, the push notification is generated from the event, not the system message. Is the system message for something else?

johnxnguyen:
@typfel ok so now we return early iff it is a `participantsAdded` or `participantsRemoved` system message and it is not exclusively for the self user.

PR #573, opened 2017-09-28
Conversation list off fix

mikeger:
# Issue

Problem appears when 2 user sessions are active (A, B) -- for example user started the app and switched from A to B and back: A is active, B on background.

1. User sends the app to the background -> Observation system is paused
2. Someone creates the conversation with user from B and writes the message while the app is in the 
3. User activates the app
4. The observation of the list has to catch up, and now it's done in UI and only for the active session with `    [ZMConversationList refetchAllListsInUserSession:ZMUserSession.sharedSession]`
5. The list on the session A is up to date, but on B it's missing the new conversation

Consequences: conversation is missing from the list if user activates B, the dot is absent for B.

# Solution

Make `ZMUserSession` refresh it's lists when the app is going to the foreground.

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -417,6 +417,7 @@ public protocol LocalMessageNotificationResponder : class {
 
         if let backgroundSession = self.backgroundUserSessions[account.userIdentifier] {
             session = backgroundSession
+            ZMConversationList.refetchAllLists(inUserSession: session)
         }
         else {
             guard let newSession = authenticatedSessionFactory.session(for: account, storeProvider: provider) else {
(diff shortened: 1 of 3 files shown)


Source/SessionManager/SessionManager.swift
```diff
@@ -389,7 +389,7 @@ public protocol LocalMessageNotificationResponder : class {
             dispatchGroup: dispatchGroup,
             migration: { [weak self] in self?.delegate?.sessionManagerWillStartMigratingLocalStore() },
             completion: { [weak self] provider in
-                self?.createSession(for: account, with: provider) { session in
+                self?.activateSession(for: account, with: provider) { session in
```
typfel: What does this fix?

typfel: Ah renamed

PR #574, opened 2017-09-28
Files whitelist

nicorsm:
Added a whitelist of files, in order to avoid folders like `com.apple.nsurlsessiond` or files like `.DS_Store` to be included in the dictionary grouping - which splits folders into `assigned` and `unassigned` - available in the `FileRelocator` extension. 

I'll try to write 1-2 tests about this whitelist.

Related to https://github.com/wireapp/wire-ios-share-engine/pull/47 and https://github.com/wireapp/wire-ios-message-strategy/pull/98


Source/UserSession/FileRelocator.swift
```diff
@@ -60,7 +64,7 @@ extension ZMUserSession {
                 }
             }
         } else if result.unassigned.count > 0 {
-            requireInternal(false, "Caches folder contains items that have not been assigned to an account. Items should always be assigned to an account. Use `FileManager.cachesURLForAccount(with accountIdentifier:, in sharedContainerURL:)` to get the default Cache location for the current account")
```
typfel: This could stay. `requireInternal` will only crash on internal builds.

Source/UserSession/FileRelocator.swift
```diff
@@ -18,6 +18,10 @@
 
 import Foundation
 
+//whitelisted files, so the FileRelocator doesn't consider to check these system files.
```
typfel: a space after the `//` would make it easier to read.

Tests/Source/UserSession/ZMUserSessionTest+FileRelocation.swift
```diff
@@ -43,4 +40,48 @@ class ZMUserSessionRelocationTests : ZMUserSessionTestsBase {
...
+        // given
+        let oldLocation = FileManager.default.cachesURLForAccount(with: nil, in: self.sut.sharedContainerURL)
+        clearFolder(at: oldLocation)
+        
+        // when
+        let newLocation = try writeTestFile(name: "example", at: oldLocation)
+        ZMUserSession.moveCachesIfNeededForAccount(with: self.userIdentifier, in: self.sut.sharedContainerURL)
+        
+        //then
+        XCTAssertFalse(FileManager.default.fileExists(atPath: newLocation.path))
```
vytis: Also might be useful to check that the file doesn't exist in the previous place

nicorsm: Maybe names might be confusing... The previous location is the Caches folder. I just append  the name of the file (in this case, `example`) in order to have the "new" location. I can rename these variables for more clarity.

vytis: I mean that to cover more situations in tests you could check both things - that the file you just moved is in the new place and also that it doesn't exist in the old location. This would catch a case where we are copying file instead of moving them.

PR #576, opened 2017-09-29
Fix/group missed call ZIOS 8951

johnxnguyen:
## Issue
If one account is logged into different two iOS devices, and one device is used to answer a group call, then the other device receives a notification that they missed a call from themselves.

## Reason
We process a missed call notification if we do not join a group call, and this call ends normally. However, the condition `callState == .terminating(reason: .normal)` will be true for any terminating call state, because definition of call state equality does not consider the reason (see `WireCallCenterV3.swift` line 106). As a result, we we're processing missed call notifications with the reason `answeredElsewhere` where the user object refers to the self user.

## Solution
Using a `switch` to correctly check the `callState` value.

## Extra
The condition `callState == .terminating(reason: .canceled)` was removed as it is unnecessary and handled by the `WireCallCenterMissedCallObserver`

### DIFF
--- a/Source/UserSession/CallStateObserver.swift
+++ b/Source/UserSession/CallStateObserver.swift
@@ -79,13 +79,15 @@ extension CallStateObserver : WireCallCenterCallStateObserver, WireCallCenterMis
             
             self.updateConversationListIndicator(convObjectID: conversation.objectID, callState: callState)
             
-            let systemMessage = self.systemMessageGenerator.appendSystemMessageIfNeeded(callState: callState, conversation: conversation, user: user, timeStamp: timeStamp)
-            if systemMessage?.systemMessageType == .missedCall
-                && (callState == .terminating(reason: .canceled) || callState == .terminating(reason: .normal) && conversation.conversationType == .group)
-            {
-                // group calls we didn't join, end with reason .normal. We should still insert a missed call in this case.
-                // since the systemMessageGenerator keeps track whether we joined or not, we can use it to decide whether we should show a missed call APNS
-                self.localNotificationDispatcher.processMissedCall(in: conversation, sender: user)
+            if let systemMessage = self.systemMessageGenerator.appendSystemMessageIfNeeded(callState: callState, conversation: conversation, user: user, timeStamp: timeStamp) {
+                switch (systemMessage.systemMessageType, callState, conversation.conversationType) {
+                case (.missedCall, .terminating(reason: .normal), .group):
+                    // group calls we didn't join, end with reason .normal. We should still insert a missed call in this case.
+                    // since the systemMessageGenerator keeps track whether we joined or not, we can use it to decide whether we should show a missed call APNS
+                    self.localNotificationDispatcher.processMissedCall(in: conversation, sender: user)
+                default:
+                    break
+                }
             }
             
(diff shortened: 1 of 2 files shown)


typfel:
I know It's a bit confusing but `.terminating(reason: .canceled)` does not only happen in the missedCall observer. The missed call observer is for calls older that a certain limit that you'll find out about when come back online and start catching up on the notification stream.

johnxnguyen:
@typfel ok good to know, I'll put it back in the condition 👍

Source/UserSession/CallStateObserver.swift
```diff
@@ -79,13 +79,15 @@ extension CallStateObserver : WireCallCenterCallStateObserver, WireCallCenterMis
...
-            let systemMessage = self.systemMessageGenerator.appendSystemMessageIfNeeded(callState: callState, conversation: conversation, user: user, timeStamp: timeStamp)
-            if systemMessage?.systemMessageType == .missedCall
-                && (callState == .terminating(reason: .canceled) || callState == .terminating(reason: .normal) && conversation.conversationType == .group)
-            {
-                // group calls we didn't join, end with reason .normal. We should still insert a missed call in this case.
-                // since the systemMessageGenerator keeps track whether we joined or not, we can use it to decide whether we should show a missed call APNS
-                self.localNotificationDispatcher.processMissedCall(in: conversation, sender: user)
+            if let systemMessage = self.systemMessageGenerator.appendSystemMessageIfNeeded(callState: callState, conversation: conversation, user: user, timeStamp: timeStamp) {
+                switch (systemMessage.systemMessageType, callState, conversation.conversationType) {
+                case (.missedCall, .terminating(reason: .normal), .group):
```
typfel: this will not end using the `==` function when it matches the `callState` against  `terminating(reason: .normal)`?

typfel: But since you have test for this I guess It's safe to it's doesn't :)

johnxnguyen: no it considers the associated reason. I just did a sanity check in a playground to confirm 😄

Source/UserSession/CallStateObserver.swift
```diff
@@ -79,13 +79,18 @@ extension CallStateObserver : WireCallCenterCallStateObserver, WireCallCenterMis
...
-                && (callState == .terminating(reason: .canceled) || callState == .terminating(reason: .normal) && conversation.conversationType == .group)
-            {
-                // group calls we didn't join, end with reason .normal. We should still insert a missed call in this case.
-                // since the systemMessageGenerator keeps track whether we joined or not, we can use it to decide whether we should show a missed call APNS
-                self.localNotificationDispatcher.processMissedCall(in: conversation, sender: user)
+            if let systemMessage = self.systemMessageGenerator.appendSystemMessageIfNeeded(callState: callState, conversation: conversation, user: user, timeStamp: timeStamp) {
+                switch (systemMessage.systemMessageType, callState, conversation.conversationType) {
+                case (.missedCall, .terminating(reason: .canceled), _ ):
+                    // the caller canceled the call
+                    fallthrough
```
vytis: You can just write one `case` above another, no need for `fallthrough`

PR #578, opened 2017-09-29
Don't start loading loading the new user session until the UI has been torn down

typfel:
### Problem
we were crashing on switching between accounts
https://rink.hockeyapp.net/manage/apps/65633/app_versions/1150/crash_reasons/188662498

### Cause
when running `select()` to switch active account the following happens:

1. start tearing down UI and display the skeleton
2. load the new session and set `activeUserSession`
3. tearing down the UI finishes and we nil out the `activeUserSession`
4. we start loading the UI but it will crash because `activeUserSession` is nil 

### Solution
Delay loading the new session until we've finished tearing down the UI


typfel:
Did some testing on the simulator and it seems to work.

PR #585, opened 2017-10-05
Possible fix for messages getting stuck

mikeger:
# Issue

For some users frequently it is not possible to send the message because of the timeout.

# Investigation

It comes out that at least one of the issue cases can be reproduced with the following steps:
1. Configure the client to have 2 accounts logged in (A is active, B inactive)
2. Kill the client
2. Enable Wi-Fi
2. Start the client
3. Disable Wi-Fi
4. Send the message from another device to the B (inactive) account
5. Activate B
-> Any message sent on B would not be delivered.

During the investigation it came clear that no request was generated, since the `ApplicationStatus` in `AbstractRequestStrategy` was signaling that `notificationFetchStatus` is still `.inProgress`, which means that the sync engine was thinking that we are still fetching the notification stream due to the push received. This is caused by the fact that before the SE never fetched the notification stream due to the push when the app was in the foreground, whereas with the multi-account support this is the normal behavior for the user session that is started on the background.

After the check it comes out that the Wi-Fi switching in STR is irrelevant.

Possible fix is in `ZMMissingUpdateEventsTranscoder.m`, so `isFetchingStreamForAPNS` is only relying on the `pingbackStatus`'s status, which seems enough at this point.

### DIFF
--- a/Source/Synchronization/ZMSyncStrategy.m
+++ b/Source/Synchronization/ZMSyncStrategy.m
@@ -370,7 +370,7 @@ ZM_EMPTY_ASSERTING_INIT()
         return nil;
     }
     
-    return [self.requestStrategies firstNonNilReturnedFromSelector:@selector(nextRequest)];;
+    return [self.requestStrategies firstNonNilReturnedFromSelector:@selector(nextRequest)];
 }
 
 - (ZMFetchRequestBatch *)fetchRequestBatchForEvents:(NSArray<ZMUpdateEvent *> *)events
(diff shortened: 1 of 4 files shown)


typfel:
Can this scenario be captured in a test?

Source/Synchronization/ZMSyncStrategy.m
```diff
@@ -370,7 +370,7 @@ - (ZMTransportRequest *)nextRequest
         return nil;
     }
     
-    return [self.requestStrategies firstNonNilReturnedFromSelector:@selector(nextRequest)];;
```
vytis: I knew it, all these semicolons just cause us trouble!

mikeger:
@typfel I'll test the particular `if` for now

typfel:
OK

PR #586, opened 2017-10-05
Add account limit check to SessionManager

typfel:
### Problem
It was possible to add more than three accounts by triggering a logout and then logging into a new account.

### Solution
We do an additional account limit check before adding the account.


vytis:
CI fails because we already expect iOS11 simulator. Will be fixed when https://github.com/wireapp/wire-ios-sync-engine/pull/583 is merged

PR #589, opened 2017-10-08
Local Notification Refactorization

johnxnguyen:
## Why?
We would like to simplify and streamline the way notifications are created and managed by the `LocalNotificationDispatcher`. Currently, there exists multiple notification classes for various types of notification (messages, system messages, calling , etc) and the logic often overlaps. 

The aim is to create a single notification object type `ZMLocalNotification`, that encapsulates all the necessary information in a configured state for all types of notifications. Furthermore, the notification can be used to generate a `UILocalNotification` or `UNNotification` object, as well as instantiate a `ChatHeadView` for in-app notifications.

## Progress
- [x] Port `ZMLocalNotification` to swift
- [x] Base implementation for message notifications
- [x] Implement call notifications
- [x] Implement event notifications 
- [x] Update `LocalNotificationDispatcher`
- [x] Update tests

## Structural Overview
`LocalNotificationDispatcher` has three entry points to generate different categories of notifications, these are *event*, *calling* and *message* notifications. Within each category there may be further subdivisions, for example, incoming or missed calls, reaction or connection events, etc. Each of these types of notifications may require different types of information when calculating the content of the notification, or if it should be created at all. As a result it is quite difficult to create a single notification class to handle such a variety of notification types.

The `ZMLocalNotification` class bundles the most essential and shared information regarding notifications, such as the text for `title` and `body`, the `soundName` and `category`  and so forth. The actual construction of the notification is delegated to helper "Builder" classes. These builders are provided with the necessary data to generate and configure notifications for a specific case. The `ZMLocalNotification` class simply queries on of these builders to set it's own properties.

The builder objects are created automatically by `ZMLocalNotification` through the call of one of it's initialisers. Thus, initialising a `ZMLocalNotification` with an `ZMUpdateEvent` will identify what kind of event it is, and create the appropriate builder accordingly.

Lastly, each `ZMLocalNotification` holds a reference to a corresponding `UILocalNotification`, which can be scheduled and/or cancelled with the application.

- [x] depends on: https://github.com/wireapp/wire-ios-data-model/pull/380


Source/Notifications/Push notifications/Notification Types/ZMLocalNote.swift
```diff
@@ -0,0 +1,124 @@
...
+let MessageNonceIDStringKey = "messageNonceString"
+let ConversationIDStringKey = "conversationIDString"
+let EventTimeKey = "eventTime"
+
+/// This class encapsulates all the data necessary to produce a local
+/// notification. It configures and formats the textual content for
+/// various notification types (message, calling, etc.) and includes
+/// information regarding the conversation, sender, and team name.
+///
+open class ZMLocalNote: NSObject {
```
typfel: We don't need to prefix with `ZM` when working in Swift.

typfel:
About the `NotificationConstructor` and it's implementations. This seems to be a variant of the builder pattern (https://en.wikipedia.org/wiki/Builder_pattern) so it might be more clear to name them `NotificationBuilder`,  `ReactionEventNotificationBuilder` and so on.

Source/Notifications/Push notifications/Notification Types/ZMLocalNote.swift
johnxnguyen: cool, it's a temporary name anyway. I'll give it a more suitable name when we're ready to remove `ZMLocalNotification` and the related protocols

Source/Notifications/Push notifications/LocalNotificationDispatcher+Calling.swift
```diff
@@ -22,28 +22,20 @@ public extension LocalNotificationDispatcher {
...
-        let note =  notification(for: conversation, sender: sender)
+        let note = ZMLocalNote(callState: callState, conversation: conversation, sender: sender)
+        note.apply(scheduleLocalNotification)
         
-        callingNotifications.cancelNotifications(conversation)
-        
-        note.update(forCallState: callState)
-        scheduleNotification(note)
+        // TODO: cancelling other call notes for this conversation?
+//        callingNotifications.cancelNotifications(conversation)
```
mikeger: Commented code (just something not to forget)

johnxnguyen: yeah, there's a few reminders, I'll get to them :)

Source/UserSession/ZMUserSession+Push.swift
```diff
@@ -137,9 +137,9 @@ extension ZMUserSession: PushDispatcherOptionalClient {
 
 extension ZMUserSession: ForegroundNotificationsDelegate {
     
-    public func didReceieveLocalMessage(notification: UILocalNotification, application: ZMApplication) {
+    public func didReceieveLocal(notification: ZMLocalNotification, application: ZMApplication) {
         DispatchQueue.main.performAsync {
```
typfel: Why are we doing a `DispatchQueue.main.performAsync` and not a `managedObjetctContext.performGroupedBlock`

johnxnguyen: we just want to switch to the UI thread, is it better to perform a block on the ui moc?

typfel: with `performGroupedBlock` we can wait for it to complete in the tests. In general we should never do `DispatchQueue.main.performAsync` since it make the tests interfere with each other.

johnxnguyen: ok good to know, I'll fix it now 😄

mikeger:
@johnxnguyen should we merge it now?

PR #593, opened 2017-10-12
Fix/zios 9069 xcode9 fix issues with  method called on background thread messages

billypchan:
run cancelLocalNotification & get UIApplication.applicationState form main thread


Source/Notifications/Push notifications/Helpers/ZMLocalNotificationSet.swift
```diff
@@ -104,7 +104,11 @@ import WireTransport
         notifications.forEach{
             if($0.conversationID == conversation.remoteIdentifier) {
                 toRemove.insert($0)
-                $0.uiNotifications.forEach{ application?.cancelLocalNotification($0) }
+                $0.uiNotifications.forEach{ notification in
+                    DispatchQueue.main.async {
```
typfel: We rarely use `DispatchQueue.main.async` because it makes tests hard to write. I would suggest to to instead of doing the switch to the main queue here do it further up in the call chain(`LocalNotificationDispatcher`). It will have reference to `NSManagedObjectContext` on which you can run `performGroupedBlock`which allows us to wait for things to complete in the tests.

Source/UserSession/BackgroundAPNSConfirmationStatus.swift
```diff
@@ -31,8 +31,23 @@ import UIKit
...
+        }
+
+        let _ = group.wait()
+        
+        return applicationState! ///FIXME: !!
+    }
+    
     open var needsToSyncMessages : Bool {
-        return messageNonces.count > 0 && application.applicationState == .background
+        return messageNonces.count > 0 && applicationState == .background
```
typfel: Here I would change this to query `OperationStatus` which lives inside the ApplicationStatusDirectory.

billypchan:
I will branch form develop again and restart the task, thanks!

PR #594, opened 2017-10-13
Fix using wrong UUID when reporting call connection state of call

typfel:
Not correctly informing Callkit that the call had connected would trigger and error when the remote party ended the call.

https://wearezeta.atlassian.net/browse/ZIOS-9093


mikeger:
~~Will it blend~~ can it be tested?

typfel:
Yes, should be testable. I'll check

PR #595, opened 2017-10-13
Fix/zios 9069 (take2) xcode9 fix issues with  method called on background thread messages

billypchan:
tested 3 cases to trigger the run time issues:

1. start the app
2. receive a message
3. go to background and receive a message

Reviewer: please check LocalNotificationDispatcher+Messages.swift and LocalNotificationDispatcher.swift to confirm the background cases should include .backgroundCall or not.


Source/Notifications/Push notifications/LocalNotificationDispatcher.swift
```diff
@@ -81,7 +84,11 @@ public class LocalNotificationDispatcher: NSObject {
 extension LocalNotificationDispatcher: ZMEventConsumer {
     
     public func processEvents(_ events: [ZMUpdateEvent], liveEvents: Bool, prefetchResult: ZMFetchRequestBatchResult?) {
-        if self.application.applicationState != .background {
+        ///FIXME: reviewer: should we return in these cases?
```
typfel: We should treat (background, backgroundCall, backgroundFetch and backgroundTask) the same in this case.

Source/Notifications/Push notifications/Helpers/ZMLocalNotificationSet.swift
```diff
@@ -104,7 +107,11 @@ import WireTransport
         notifications.forEach{
             if($0.conversationID == conversation.remoteIdentifier) {
                 toRemove.insert($0)
-                $0.uiNotifications.forEach{ application?.cancelLocalNotification($0) }
+                $0.uiNotifications.forEach{ notification in
+                    userSession.performChanges {
```
typfel: I would not inject the user session into the `ZMLocalNotificationSet.swift` but instead do the `userSession.performChanges` from the `LocalNotificationDispatcher`. 

Also related. `userSession.performChanges` will do a `performGroupedBlockAndWait` and `saveOrRollbackChanges`. This is not necessary here because we are not changing a core data entity. The correct way to just execute something on a context is to do `usersSession.managedObjectcontext.performGroupedBlock`

billypchan:
In LocalNotificationDispatcher replacing all userSession.performChanges with self.syncMOC.zm_userInterface.performGroupedBlock would cause crashes. I would wait @johnxnguyen finish pull request and apply the fix and test again.

billypchan:
@typfel I rewrite the code and treating cancelLocalNotification(), scheduleLocalNotification() like UIKit functions and let ZMUserSession call them in main thread.

typfel:
@billypchan I would put this on hold until @johnxnguyen changes are in.

billypchan:
This PR is closed. The story will be restarted base on latest develop branch.

PR #596, opened 2017-10-16
don’t require non active app state to handle call action

johnxnguyen:
## Problem
Since iOS11, tapping on incoming call notification for the inactive account would not switch to that account to display the call overlay.

## Reason & Solution
On iOS 11, tapping a call notification would invoke this method with an active app state, meaning we would not create and process the pending notification and thus not handle the action. By not requiring the app to be `inactive` or in `background`, the notification action is always handled. Allowing an active state is acceptable because this method is only called when tapping a system notification, which can only be done when the app is not in the foreground. If a call comes for the inactive account while the app is in the foreground, the `LocalNotificationDispatcher` will funnel the notification to the `ChatHeadsViewController`, which will handle the tap action separately.

### DIFF
--- a/Source/UserSession/ZMUserSession+Push.swift
+++ b/Source/UserSession/ZMUserSession+Push.swift
@@ -131,12 +131,12 @@ extension ZMUserSession: ForegroundNotificationsDelegate {
     }
 
     public func didReceiveLocal(notification: UILocalNotification, application: ZMApplication) {
-        if application.applicationState == .inactive || application.applicationState == .background {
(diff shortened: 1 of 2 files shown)


johnxnguyen:
Also, I have manually tested this for both iOS 10 and 11 without any fails.

mikeger:
@johnxnguyen some tests failed :(

johnxnguyen:
@mikeger yeah I'm looking at them now, seems like removing the if condition wasn't such a good idea.

johnxnguyen:
actually, it seems that this incorrect app state is an iOS 11 bug: https://openradar.appspot.com/34332569

PR #597, opened 2017-10-16
Update after removing unused calling events

typfel:
depends on:
- [x] https://github.com/wireapp/wire-ios-transport/pull/67
- [x] https://github.com/wireapp/wire-ios-data-model/pull/379
- [x] https://github.com/wireapp/wire-ios-message-strategy/pull/101


Source/Notifications/Push notifications/Helpers/ZMLocalNotificationSet.swift
```diff
@@ -146,17 +146,6 @@ public extension ZMLocalNotificationSet {
         return nil
     }
     
-    public func cancelNotificationForIncomingCall(_ conversation: ZMConversation) {
```
mikeger: Would conflict with John's changes most likely

typfel: Yep, but shouldn't be too difficult to resolve.

typfel:
Tests passed locally

PR #598, opened 2017-10-17
Fix/first push

johnxnguyen:
## Problem
If the app receives a push (message, calling etc) while in a terminated state, the app is woken but does not display a notification. Only after receiving a second push do both the first and second notifications get fired.

## Reason
When the `PKPushRegistryDelegate` invokes the `pushRegistry:didReceiveIncomingPush:forType:` method after receiving the initial push, it asks `BackgroundActivityFactory` for a background activity session in order to process the payload. However the factory class was not beginning the background activity  because it required it's `application` and `mainGroupQueue` properties to be non nil (which are being set each time we load a `ZMUserSession` and is a separate issue to address).

## Solution
Set these properties before we initialise the `PushDispatcher`.

### DIFF
--- a/Source/SessionManager/SessionManager+Push.swift
+++ b/Source/SessionManager/SessionManager+Push.swift
@@ -155,17 +155,7 @@ extension SessionManager: PushDispatcherClient {
             return
         }
         
-        guard let userInfoData = payload[PushChannelDataKey] as? [String: Any] else {
-            log.debug("No data dictionary in notification userInfo payload");
-            return
-        }
-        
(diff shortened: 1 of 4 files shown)


johnxnguyen:
@mikeger in the end we didn't need to extract the account id from the payload and forward the push to the corresponding user session, but I left the refactoring in, maybe it will be useful later on.

mikeger:
@johnxnguyen how about a small test to verify that after creating the SessionManager it is possible to create the background task?

johnxnguyen:
@mikeger sure!

PR #599, opened 2017-10-18
Add support for new CBR API

typfel:
AVS as re-factored their CBR support and the API has changed. This PR update the wire call center for these changes.


Source/Calling/WireCallCenterV3+Notifications.swift
```diff
@@ -275,6 +274,22 @@ extension WireCallCenterV3 {
...
+    /// Returns a token which needs to be retained as long as the observer should be active.
+    public class func addConstantBitRateObserver(observer: ConstantBitRateAudioObserver, userSession: ZMUserSession) -> Any {
+        return addConstantBitRateObserver(observer: observer, context: userSession.managedObjectContext)
+    }
+    
+    /// Register observer when constant audio bit rate is enabled/disabled
+    /// Returns a token which needs to be retained as long as the observer should be active.
+    internal class func addConstantBitRateObserver(observer: ConstantBitRateAudioObserver, context: NSManagedObjectContext) -> Any {
+        return NotificationInContext.addObserver(name: WireCallCenterCBRNotification.notificationName, context: context.notificationContext, queue: .main) { [weak observer] note in
+            if let note = note.userInfo[WireCallCenterCBRNotification.userInfoKey] as? WireCallCenterCBRNotification {
```
mikeger: Would the notification arrive on the background thread?

typfel: No it will always arrive on the main queue because we specify `queue: .main`

PR #600, opened 2017-10-19
Make sure it is not possible to load the multiple sessions

mikeger:
- Issue happens that the app is started from the push notification.
- Added the precondition check to make sure the issue is not happening any more.

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -396,17 +396,10 @@ public protocol LocalMessageNotificationResponder : class {
             return
         }
 
-        LocalStoreProvider.createStack(
-            applicationContainer: sharedContainerURL,
-            userIdentifier: account.userIdentifier,
-            dispatchGroup: dispatchGroup,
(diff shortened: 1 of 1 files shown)


Source/SessionManager/SessionManager.swift
```diff
@@ -396,17 +396,10 @@ public protocol LocalMessageNotificationResponder : class {
...
-            dispatchGroup: dispatchGroup,
-            migration: { [weak self] in self?.delegate?.sessionManagerWillStartMigratingLocalStore() },
-            completion: { [weak self] provider in
-                self?.activateSession(for: account, with: provider) { session in
-                    self?.registerSessionForRemoteNotificationsIfNeeded(session)
-                    completion(session)
-                }}
-        )
+        self.activateSession(for: account) { session in
+            self.registerSessionForRemoteNotificationsIfNeeded(session)
```
typfel: This shouldn't be done in the `configureSession:`?

typfel: Ah or this only needs be done once per launch.

PR #601, opened 2017-10-19
Fix/remove duplicated users

marcoconti83:
Remove duplicated users caused by concurrent user session


Source/Synchronization/ZMHotFixDirectory+Swift.swift
```diff
@@ -42,6 +42,53 @@ extension UserClient {
...
+        // NOTE:
+        // we are not merging clients since they are re-created on demand
+        
+        self.connection = self.connection ?? user.connection
+        self.addressBookEntry = self.addressBookEntry ?? user.addressBookEntry
+        
+        let mutableLastServerSyncedActiveConversations = self.mutableOrderedSetValue(forKey: "lastServerSyncedActiveConversations")
+        mutableLastServerSyncedActiveConversations.union(user.value(forKey: "lastServerSyncedActiveConversations") as! NSOrderedSet)
+        
+        let mutableConversationsCreated = self.mutableSetValue(forKey: "conversationsCreated")
```
marcoconti83: Eventually, we will need to replace the hardcoded strings by exposing those properties!

marcoconti83:
This should be closed in favor of: https://github.com/wireapp/wire-ios-sync-engine/pull/602

PR #603, opened 2017-10-20
Fix to the typing indicator

nicorsm:
Fixed an issue that is causing a flickering to the typing indicator.

Main fixes:
- added controls in `TypingStrategy` that allow notifications to be fired only if `isTyping` or `clearIsTyping` flags actually changes. On the previous version, the `add()` method was performed even if there weren't any changes to perform, due to the lack of controls and the presence of the fallback instruction (`?? false`).
- added `isLocal` parameter that avoids local changes to affect the visibility of the view.

### DIFF
--- a/Source/UserSession/NotificationInContext+UserSession.swift
+++ b/Source/UserSession/NotificationInContext+UserSession.swift
@@ -95,7 +95,7 @@ public extension ZMConversation {
             guard let `self` = self else { return }
             
             let users = note.userInfo[typingNotificationUsersKey] as? Set<ZMUser> ?? Set()
-            let local = note.userInfo["isLocal"] as? Bool ?? false
+            let local = note.userInfo[IsLocalKey] as? Bool ?? false
             observer?.typingDidChange(conversation: self, typingUsers: users, isLocal: local)
         }
     }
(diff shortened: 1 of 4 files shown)


Source/UserSession/NotificationInContext+UserSession.swift
```diff
@@ -94,7 +94,11 @@ public extension ZMConversation {
             [weak observer, weak self] note in
             guard let `self` = self else { return }
             
-            observer?.typingDidChange(conversation: self, typingUsers: note.userInfo[typingNotificationUsersKey] as? Set<ZMUser> ?? Set())
+            let users = note.userInfo[typingNotificationUsersKey] as? Set<ZMUser> ?? Set()
+            let local = note.userInfo[IsLocalKey] as? Bool ?? false
+            if !local {
```
typfel: Why send the notification in the first place if we will ignore it here? Is something else also listening to these notifications?

nicorsm: The observer is currently referenced from:
1. `notifyTyping()` (NotificationInContext+UserSession.swift:106)
2. `notifyTranscoderThatUser(isTyping:inConversation:)` (TypingStrategy:245)
3. `clearTranscoderStateForTyping(isTyping:inConversation:)` (TypingStrategy:245)
I'll check when the methods should not be called.

nicorsm: @typfel when local = `true`, `addConversationForNextRequest(note:)` (TypingStrategy:153) is listening for the same notification

nicorsm: We could try to separate the two behaviours, but this will create two separate notification types that should be instantiated and observed.

PR #604, opened 2017-10-23
Typing indicator fix

nicorsm:
I accidentally deleted the branch for this PR: https://github.com/wireapp/wire-ios-sync-engine/pull/603.
The two new commits include:
- a refactor that deletes the "local" flag on notification (and uses the already available `ZMTypingChangeNotification` type)
- a restore to the proper use of "clear typing" notification.

### DIFF
--- a/Source/Synchronization/Strategies/TypingStrategy.swift
+++ b/Source/Synchronization/Strategies/TypingStrategy.swift
@@ -228,7 +228,6 @@ extension TypingStrategy : ZMEventConsumer {
     func processIsTypingUpdateEvent(for user: ZMUser, in conversation: ZMConversation, with status: String) {
         let startedTyping = (status == StartedKey)
         let stoppedTyping = (status == StoppedKey)
-
         if (startedTyping || stoppedTyping) {
             typing.setIs(startedTyping, for: user, in: conversation)
         }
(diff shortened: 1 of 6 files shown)


Source/Synchronization/Strategies/TypingStrategy.swift
```diff
@@ -153,10 +160,11 @@ public class TypingStrategy : AbstractRequestStrategy {
         guard let conversation = note.object as? ZMConversation, conversation.remoteIdentifier != nil
         else { return }
         
-        let isTyping = (note.userInfo[IsTypingKey] as? NSNumber)?.boolValue ?? false
-        let clearIsTyping = (note.userInfo[ClearIsTypingKey] as? NSNumber)?.boolValue ?? false
+        let isTyping = (note.userInfo[IsTypingKey] as? NSNumber)?.boolValue
         
-        add(conversation:conversation, isTyping:isTyping, clearIsTyping:clearIsTyping)
+        if isTyping != nil {
+            add(conversation:conversation, isTyping: isTyping!, clearIsTyping: false)
```
vytis: A nitpick, but it maybe would be better to do `if let` instead of checking for nil and then force-unwrapping

nicorsm:
Need to take a look at tests 😕

mikeger:
@nicorsm this is good to be merged now from your side?

nicorsm:
@mikeger yes, I'm done with it.

PR #605, opened 2017-10-24
Fix for double system message

typfel:
ZMConversationTranscoder has taken over the task of inserting system messages.

*NOTE:* also noticed that processDownloadedEvents was never called anywhere and
that therefore all events are considered "live".

---

## Problem
When a user changes a conversation name or add/remove a participant we currently will insert
two system message for this event. This happens because the change event arrives twice, first
in the response of POST response and second in the notification stream.

## Solution:
Add a check before inserting these system messages for if the change has already been applied.

### DIFF
--- a/Tests/Source/Synchronization/Transcoders/ZMConversationTranscoderSystemMessageTests.swift
+++ b/Tests/Source/Synchronization/Transcoders/ZMConversationTranscoderSystemMessageTests.swift
@@ -53,6 +53,9 @@ class ZMConversationTranscoderSystemMessageTests: ObjectTranscoderTests {
         self.sut = nil
         self.localNotificationDispatcher = nil
         self.conversation = nil
+        self.user = nil
+        self.mockSyncStatus = nil
+        
         super.tearDown()
     }
     
--- a/Tests/Source/Synchronization/Transcoders/ZMConversationTranscoderSystemMessageTests.swift
+++ b/Tests/Source/Synchronization/Transcoders/ZMConversationTranscoderSystemMessageTests.swift
@@ -1,0 +1,275 @@
+//
+// Wire
+// Copyright (C) 2017 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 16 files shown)


Tests/Source/Synchronization/Transcoders/ZMConversationTranscoderSystemMessageTests.swift
```diff
@@ -0,0 +1,275 @@
...
+            self.user.remoteIdentifier = UUID.create()
+            
+            self.syncMOC.saveOrRollback()
+        }
+    }
+    
+    override func tearDown() {
+        self.sut = nil
+        self.localNotificationDispatcher = nil
+        self.conversation = nil
```
vytis: Need to nil `user` and `mockSyncStatus`

Tests/Source/Synchronization/Transcoders/ZMConversationTranscoderTests.m
```diff
@@ -59,23 +59,13 @@ - (void)setUp
...
-        [downloadedEvents addObjectsFromArray:events];
-        return YES;
-    }]];
-    
-    self.downloadedEvents = downloadedEvents;
     self.syncStateDelegate = [OCMockObject niceMockForProtocol:@protocol(ZMSyncStateDelegate)];
     self.mockSyncStatus = [[MockSyncStatus alloc] initWithManagedObjectContext:self.syncMOC syncStateDelegate:self.syncStateDelegate];
     self.mockSyncStatus.mockPhase = SyncPhaseDone;
     self.mockApplicationStatus.mockSynchronizationState = ZMSynchronizationStateEventProcessing;
+    self.mockLocalNotificationDispatcher = [OCMockObject niceMockForProtocol:@protocol(PushMessageHandler)];
```
vytis: Important to remember to do `stopMocking` on all `nice` mocks or they will leak everything. The best would be to avoid adding any new code with OCMock, in this case a simple class would do.

typfel: I'll replace with a regular mock which already exists.

PR #609, opened 2017-10-27
Fix leave system message not appear for the self user

typfel:
### Problem
The previous fix for double system message were checking if the removed/added user change has already been applied to the `mutableLastServerSyncedActiveParticipants` set. This doesn't work when the self user is removed since by definition the self user is never part of  the `mutableLastServerSyncedActiveParticipants`.

### Solution
Compare against the `activeParticipants` set instead which does contain the self user. This set will however also contain local change so we must check the `modifiedKeys` set to know if the change is not yet synced. 

depends on:
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/59


mikeger:
@typfel testThatWhenLeavingAConversationWeSetAndSynchronizeTheLastReadServerTimestamp got broken somehow.

PR #612, opened 2017-10-30
Notify when the AVS flow manager is created

mikeger:
- Necessary to know when we can configure the flow manager for uploading the call stats


typfel:
Not related to these changes but I think we need to re-check only setting the delegate once with the multiple `UserSession`:s running in parallel. But I guess the flow manager will be gone soon anyway.

mikeger:
@typfel yeah, I noticed that every new user session is resetting the delegate, could be the source of bugs potentially. I will create the ticket to sort it out.

PR #613, opened 2017-10-30
Fix/zios 9147 get app status form user session.operation status.operation state instead of application.application state

billypchan:
get app state from:
userSession.operationStatus.operationState 
instead of:
application.applicationState


Source/UserSession/ZMUserSession+Private.h
```diff
@@ -47,7 +47,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, readonly) ZMTransportSession *transportSession;
 @property (nonatomic, readonly) NSManagedObjectContext *searchManagedObjectContext;
-@property (nonatomic, readonly) OperationStatus *operationStatus;
+@property (nonatomic, readonly, nullable) OperationStatus *operationStatus;
```
typfel: `OperationStatus` cannot be nil it's created and destroyed together with the `ZMUserSession`

Source/Notifications/Push notifications/LocalNotificationDispatcher.swift
```diff
@@ -35,24 +35,27 @@ public class LocalNotificationDispatcher: NSObject {
...
+    unowned let application: ZMApplication
+    unowned let userSession: ZMUserSession
     let syncMOC: NSManagedObjectContext
     private(set) var isTornDown: Bool
     private var observers: [Any] = []
     
     var localNotificationBuffer = [ZMLocalNotification]()
     
-    @objc(initWithManagedObjectContext:foregroundNotificationDelegate:application:)
+    @objc(initWithManagedObjectContext:foregroundNotificationDelegate:application:userSession:)
```
typfel: Why not inject only the `OperationStatus` since that's the only thing we are accessing on the user session?

Source/UserSession/ZMUserSession+Private.h
typfel: Did you mark this as nullable since some other object in the chain is nullable?

billypchan: In normal case it should not be null. But when running tests it would be null. One of the following items in the chain is null:

- (OperationStatus *)operationStatus
{
    return self.operationLoop.syncStrategy.applicationStatusDirectory.operationStatus;
}

typfel: Ok, then I think we should fix the tests if they need it to be there. How many tests does this affect?

Source/Notifications/Push notifications/LocalNotificationDispatcher.swift
billypchan: Thanks, I am rewiting with OperationStatus as parameter.

Source/UserSession/ZMUserSession+Private.h
billypchan: @typfel These tests are updated with the new mocking background status:
Tests/Source/Calling/CallStateObserverTests.swift
Tests/Source/Notifications/PushNotifications/LocalNotificationDispatcherCallingTests.swift
Tests/Source/Notifications/PushNotifications/LocalNotificationDispatcherTests.swift

code changed:
-        self.application.applicationState = .background
+        self.mockUserSession.operationStatus?.isInBackground = true

typfel: I expected you to remove the `nullable` from above.

Tests/Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoderTests.m
```diff
@@ -57,7 +57,7 @@ - (void)setUp {
     self.mockSyncStatus = [[MockSyncStatus alloc] initWithManagedObjectContext:self.syncMOC syncStateDelegate:self.syncStateDelegate];
     self.mockSyncStatus.mockPhase = SyncPhaseDone;
     self.mockOperationStatus = [[OperationStatus alloc] init];
-    self.mockOperationStatus.isInBackground = NO;
+    self.mockOperationStatus.isInBackground = YES;
```
typfel: Why did this change from NO to YES?

Tests/Source/MessagingTest.h
```diff
@@ -68,6 +68,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) NSString *groupIdentifier;
 @property (nonatomic, readonly) NSUUID *userIdentifier;
 @property (nonatomic, readonly) NSURL *sharedContainerURL;
+@property (nonatomic, readonly) SyncEngineOperationState operationState;
```
typfel: This doesn't seem to be referenced anywhere afaik. You could remove it.

PR #618, opened 2017-11-03
Remove usage of legacy fields

typfel:
Remove all usage of the following fields: 

remove `status`, `status_ref`,` status_time` from the self member object
remove `status` from the other member object (it is always 0)
remove `last_event` and `last_event_time` from the conversationMeta

in addition to that the following fields has also been removed:
`archived`, `clearedEventID`, `muted`, `mutedTime`, `lastRead`

depends on:
- [x] https://github.com/wireapp/wire-ios-data-model/pull/392
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/60


typfel:
One thing which is still missing is to remove the self user from conversations which return 404 and conversations no longer returned by the `/conversations` endpoint. But I'll do that as a separate PR since it's a bigger change.

Tests/Source/Integration/ConversationsTests.m
```diff
@@ -730,7 +717,7 @@ - (void)testThatItSendsAConversationWindowChangeNotificationsIfAConversationIsCh
             [NSThread sleepForTimeInterval:0.002]; // SE has milisecond precision
         }
         
-        self.groupConversation.lastRead = ((MockEvent *)self.groupConversation.events.lastObject).identifier;
+//        self.groupConversation.lastRead = ((MockEvent *)self.groupConversation.events.lastObject).identifier;
     }];
```
mikeger: Comment -- can be removed?

typfel: Yep

PR #621, opened 2017-11-07
Respect "hide notification preview" setting

johnxnguyen:
When this setting is enabled, then the APNS notification should have no title and the body should say *New message*.

Note: since this setting affects only APNS notifications, the logic only affects the `ZMLocalNotification`s `uiNotification` properties.

### DIFF
--- a/Source/Notifications/Push notifications/Notification Types/ZMLocalNotification+Events.swift
+++ b/Source/Notifications/Push notifications/Notification Types/ZMLocalNotification+Events.swift
@@ -46,6 +46,7 @@ extension ZMLocalNotification {
         }
         
         self.init(conversation: conversation, type: .event(event.type), builder: builder!)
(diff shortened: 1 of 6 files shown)


Source/Notifications/Push notifications/Notification Types/ZMLocalNotification+Events.swift
```diff
@@ -46,6 +46,7 @@ extension ZMLocalNotification {
         }
         
         self.init(conversation: conversation, type: .event(event.type), builder: builder!)
+        self.shouldHideContent = (builder as? EventNotificationBuilder)?.shouldHideContent ?? false
```
typfel: This setting can't be changed without restarting the app?

typfel: Ah this is the `ZMLocalNotification` init I saw now.

PR #624, opened 2017-11-08
Fixed an error that was causing the typing indicator to disappear on both sides when sending a message

nicorsm:
Related to ZIOS-7488. Special thanks to @marcoconti83!


johnxnguyen:
@nicorsm could you describe what the problem was?

nicorsm:
@johnxnguyen Sure :) 
When typing something in a conversation (1:1 or group), sending the message was causing the typing indicator to disappear for all the participants in that conversation. 

The problem was, when sending a message from A to B, two notifications were fired: `conversation.typing = stopped`, which was decrementing the number of typing users, and the proper `conversation.otr-message-add` with the message sent (and no informations about the number of currently typing users.

But the delivery confirmations for both notifications (typing and message-add) are sent back from B to A, causing the message-add to update the  number of typing users to 0, since it doesn't have information of this kind.

This is solved by allowing only those types of notifications to update the typing indicator:
- send text
- edit text
- send text (ephemeral)

So, excluding other types like delivery notifications, images, video, etc.

 https://github.com/wireapp/wire-ios-sync-engine/blob/7865a1a4fb6c3a2f4fc28a3f41bb85ea9d186629/Source/Synchronization/Strategies/TypingStrategy.swift#L231-L234

mikeger:
Some issue with tests on CI, I restarted them.

nicorsm:
...still some weird behaviours with CI 😕

PR #626, opened 2017-11-08
extend SessionManagerObserver

johnxnguyen:
## Why?
The implementation of multiple account notification sounds (in the UI) needs to know when a `ZMUserSession` has been created or destroyed. We extend the `SessionManagerObserver` with the optional method `sessionManagerDestroyedUserSession(for accountID:)` so that existing observers still conform to the protocol.

## Edit
I've renamed the `SessionManagerDelegate` method `sessionManagerCreated(userSession:)` to `sessionManagerActivated(userSession:)` because this method is only called explicitly when a session is activated, not necessarily when a new user session is created.

Also, the session created and destroyed notifications have been split into two separate observers. The new session notification is posted when brand new user session is created (within `startBackgroundSession(...)`) and the destroyed session notification is posted when a session is torn down or the session is logged out.


Source/SessionManager/SessionManager.swift
```diff
@@ -781,21 +782,34 @@ extension SessionManager : PreLoginAuthenticationObserver {
...
 @objc public protocol SessionManagerObserver: class {
     func sessionManagerCreated(userSession : ZMUserSession)
+    @objc optional func sessionManagerDestroyedUserSession(for accountId : UUID)
 }
 
 private let sessionManagerObserverNotificationName = Notification.Name(rawValue: "ZMSessionManagerObserverNotification")
 
 extension SessionManager: NotificationContext {
     
+    private var createdKey: String { return "created" }
```
mikeger: Can be replaced with `let createdKey = "created"`

Source/SessionManager/SessionManager.swift
```diff
@@ -781,21 +782,34 @@ extension SessionManager : PreLoginAuthenticationObserver {
 // MARK: - Session manager observer
 @objc public protocol SessionManagerObserver: class {
     func sessionManagerCreated(userSession : ZMUserSession)
+    @objc optional func sessionManagerDestroyedUserSession(for accountId : UUID)
 }
 
 private let sessionManagerObserverNotificationName = Notification.Name(rawValue: "ZMSessionManagerObserverNotification")
```
mikeger: Should it be two distinctive notifications, one for creation and one for destruction?

johnxnguyen:
@mikeger do you think it's bad to notify for new user sessions when loading a background session?

Source/SessionManager/SessionManager.swift
johnxnguyen: yeah, it might be neater to have the separation

mikeger:
@johnxnguyen it should be fine, may be we can add a comment that would describe the circumstances when the session can be created and destroyed.

Source/SessionManager/SessionManager.swift
johnxnguyen: unfortunately we're in an extension here

mikeger: Eh, may be it can be static then?

johnxnguyen: oh yeah.... 😄

johnxnguyen:
@mikeger I've added some tests 😄

mikeger:
@johnxnguyen can this be merged?

johnxnguyen:
@mikeger yes we're good to go. If you can trigger a build then I can bump components for the UI

PR #627, opened 2017-11-08
Duplicated users crash

vytis:
This PR simply merges any duplicate users we encounter when handling response from the server.

Depends on and more details about the reasoning is in https://github.com/wireapp/wire-ios-data-model/pull/394


mikeger:
@vytis some issue with CI, may be because of the branch reference in Cartfile?

mikeger:
Data Model PR was merged, so I guess we can update this one to be mergeable 👍

vytis:
Releasing data model and will update 👍

PR #628, opened 2017-11-09
Update self user group participation status on slow sync

typfel:
## Problem
If a client hasn't been active for long time we will not be able to sync the entire notifications stream when it comes back online. In this case we will perform "slow sync" which fetches the state of all the conversations and users among other things, in order to have an up-to-date state to apply new notification stream events on.

It can happen that while the client was offline that the self user was removed from conversations. We didn't check for this during the slow sync.

## Solution
During the slow sync we fetch a list of all conversations the self user is currently active in. We can then remove the self user from all the conversations not in that list.

If we for some reason are removed from a conversation on the BE but we didn't do a slow sync and we try to fetch a conversation we will receive a 404 response, in which we should also remove the self user from that conversation.


typfel:
@vytis tests were failing for a good reason. It only makes sense to remove the self user on slow sync from group conversations.

PR #631, opened 2017-11-13
ZIOS-9239: Create EmailVerificationStrategy for Team creation feature

billypchan:
# Task
Create a new EmailVerificationStrategy and RegistrationStatus for the new feature "Team creation with email address, then activate with the code from the email".


Source/Synchronization/Strategies/EmailVerificationStrategy.swift
```diff
@@ -0,0 +1,91 @@
...
+        switch (currentStatus.phase) {
+        case let .sendActivationCode(email: email):
+            path = "/activate/send"
+            payload = ["email": email,
+                       "locale": NSLocale.formattedLocaleIdentifier()!]
+        case let .checkActivationCode(email: email, code: code):
+            path = "/activate"
+            payload = ["email": email,
+                       "code": code,
+                       "dryrun": true]
```
mikeger: What is the meaning of "dryrun" in this request?

billypchan: dryrun (boolean, optional): Whether to perform a dryrun, i.e. to only check whether activation would succeed. Dry-runs never issue access cookies or tokens on success but failures still count towards the maximum failure count.,

mikeger: So this means that we are only checking the activation code, but actual activation would come next?

PR #632, opened 2017-11-13
Add default sounds to ZMSound

typfel:
In preparation for supporting `none` sound effect and ringtones I've added the the default sounds as enum cases and create two groups: `soundEffects` and `ringtones`.

### DIFF
--- a/Source/Utility/ZMSound.swift
+++ b/Source/Utility/ZMSound.swift
@@ -118,6 +118,10 @@ public enum ZMSound: String, CustomStringConvertible {
     }
     
     public var description: String {
+        return self.rawValue.capitalized
+    }
+    
+    public var descriptionLocalizationKey: String {
         get {
             switch self {
             case .None:
--- a/Source/Utility/ZMSound.swift
+++ b/Source/Utility/ZMSound.swift
@@ -19,8 +19,10 @@
 
 import Foundation
 import AudioToolbox
+import avs
 
 public enum ZMSound: String, CustomStringConvertible {
+    case None       = "none"
     case Bell       = "bell"
     case Calipso    = "calipso"
     case Chime      = "chime"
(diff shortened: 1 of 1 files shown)


Source/Utility/ZMSound.swift
```diff
@@ -34,13 +36,15 @@ public enum ZMSound: String, CustomStringConvertible {
...
-    
     case Harp       = "harp"
     case Marimba    = "marimba"
     case OldPhone   = "old-phone"
     case Opening    = "opening"
-    
-    public static let allValues = [
+    case WireCall   = "ringing_from_them"
+    case WirePing   = "ping_from_them"
+    case WireText   = "new_message"
```
johnxnguyen: these default sounds aren't included in the `soundEffects` or `ringtones` arrays?

johnxnguyen: ah I see in the UI, the arrays hold only custom sounds.

typfel: No and yes, We combine them into a array in https://github.com/wireapp/wire-ios/pull/1415 but since we have different configurations we don't have include them here.

Source/Utility/ZMSound.swift
```diff
@@ -112,11 +119,26 @@ public enum ZMSound: String, CustomStringConvertible {
     
     public var description: String {
         get {
-            return self.rawValue.capitalized
+            switch self {
+            case .None:
+                return "self.settings.sound_menu.sounds.none"
```
mikeger: We call .localized one level up on those values, right?

mikeger: May be it would be better to be called "descriptionLocalizationKey" then, since currently it's not entirely clear that you have to extra cell .localized.

PR #635, opened 2017-11-14
Create team strategy

vytis:
## What's in this PR?

We are adding a new flow for creating team (together with admin account). The current registration is handled by `ZMAuthenticationStatus` which is pretty complicated. The plan is to slowly transition to a new Swiftified `RegistrationStatus` which encodes states into an enum with associated values.

### New functionality

* Added a new `TeamRegistrationStrategy` which handles team creation. It is very similar to `ZMRegistrationTranscoder`
* Added new method for starting team registration which expects `TeamToRegister` object. This should be created by UI when all needed information is gathered.

### DIFF
--- a/Source/Synchronization/Strategies/RegistrationStatus.swift
+++ b/Source/Synchronization/Strategies/RegistrationStatus.swift
@@ -78,6 +78,7 @@ final public class RegistrationStatus {
     /// - Parameter team: team object containing all information needed
     public func create(team: TeamToRegister) {
         phase = .createTeam(team: team)
+        RequestAvailableNotification.notifyNewRequestsAvailable(nil)
     }
 
     func handleError(_ error: Error) {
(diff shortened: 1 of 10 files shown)


Source/Synchronization/Strategies/RegistrationStatus.swift
```diff
@@ -33,12 +39,22 @@ public protocol RegistrationStatusDelegate: class {
...
+    func emailActivationCodeValidationFailed(with error: Error) {}
+}
+
 final public class RegistrationStatus {
     var phase : Phase = .none
 
     public weak var delegate: RegistrationStatusDelegate?
 
-    /// Used for start email verificiation process by sending an email with verification
+    /// Used to start email verificiation process by sending an email with verification
```
typfel: typo: `verificiation` > `verification`

PR #639, opened 2017-11-16
Assign sender to the call initiator when a call terminates

typfel:
### Issues

When a group ends the push notification says the self user called

### Causes

the user id associated with the terminating event is the self user.

### Solutions

The user id associated with the event is not relevant. We should instead use the userId associated with the incoming call event. This PR changes so that userId associated with the terminating event is always the call initiator.


mikeger:
@typfel test failed: `testThatTheClosedCallHandlerPostsTheRightNotification`

PR #640, opened 2017-11-16
Make call center API more clear

typfel:
### Issues

When a group ends the push notification says the self user called

### Causes

the user id associated with the terminating event is the self user.

### Solutions

The user id associated with the event is not relevant. We should instead use the userId associated with the incoming call event.

The sender parameter on the calling update methods has always been a bit unclear to which user it actually refers. This PR replaces it with a `caller` parameter which is always the user which initiated the call.


mikeger:
_this PR description is by accident partially boilerplate_

typfel:
I'll update PR description. Just wanted to run the test suite.

PR #643, opened 2017-11-28
Fix/localization info

johnxnguyen:
### Issues
If the user was added to a group conversation while their client was in the background, any future message APNS notifications would be incorrectly formatted.

### Causes
When building the push string key, we only really care about missing conversation names of group conversations. However, we were explicitly checking if the `conversationType` is equal to `group`, then adding the conversation name to the push string arguments array, or appending the `noconversationname` component to the push key. The problem is that if you are added to a new group conversation and the app is backgrounded, the client hasn't fetched the conversation from BE and so the `conversationType` will be `invalid`. And so the `noconversationname` isn't appended to the push key, thus generating the wrong push string with an incorrect number of arguments.

### Solutions
Improve the `if` conditions so that we consider missing the `invalid` conversation type.


mikeger:
@johnxnguyen CI failed, I restarted it.

PR #645, opened 2017-11-30
Added error localization

vytis:
### Issues

There are some errors that we show to the user and we want them to be localized.

### Causes

There was a method called `LocalizedDescriptionStringFromZMUserSessionErrorCode` which actually produced a description that is not localized and only usable for debugging purposes.

### Solutions

Added strings for all errors that we currently show in team registration flow.


Resources/ZMLocalizable.strings
```diff
@@ -19,3 +19,11 @@
 
 
 "conversation.username.groupingSeparator" = ", ";
+
+"user_session.error.balcklisted-email" = "E-mail address has been blacklisted";
```
typfel: `balcklisted` > `blacklisted`

PR #646, opened 2017-11-30
Fix call timer getting on every time someone joins a call 

typfel:
### Issues

Call timer gets reset when someone joins a call.

### Causes

The `establishedDate` gets updated when the established-call-handler is called. In group calls that once per call participants which has the effect that timer gets reset when some joins the call.

### Solutions

Work around this behaviour by no updating `establishedDate` if the call is already established.


vytis:
Any tests that could be added for this?

typfel:
Yes let's add a test for this.

PR #653, opened 2017-12-13
CircleCI 2.0 support

vytis:
## What's new in this PR?

We currently get only 1 concurrent build on CircleCI. According to their support we should move to 2.0 infrastructure to solve the issues.

## Solutions

The config was reworked to use [workflows](https://circleci.com/docs/2.0/workflows/) that were introduced with CircleCI 2.0. It basically allows splitting work into several tasks and define dependencies between them. Each build step runs on a fresh VM which means we have to make sure build artefacts are transferred between them.

We currently have split the workload into two steps:

* `build` step. This basically does `xcodebuild build-for-testing`.
* `test` step. This does `xcodebuild test-without-building`, stores test results and uploads test coverage to codecov.io.

### Caching

CircleCI 2.0 introduced a very fine-grained caching. It can archive certain files/directories and assign those files some cache key. The key can be derived in various ways, but in our case we checksum some important files. More specifically:

* `Gemfile.lock`. This contains information about all ruby gems we need to run the test suite. When this file is updated we need to drop the cache and redownload all the gems.
* `Cartfile.lock`. This contains all Carthage dependencies. The cache will also needs to be dropped when we have some updates to this file.

### Moving build artefacts between steps

This is done using [workspaces](https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs). It can save specified files from current directory and then load them back in the next step. We currently save DerivedData folder so that all build artefacts are transferred to the next step. Also build logs are saved from `build` directory.

## Dependencies

We had to update the build scripts slightly to accommodate different way the config is structured.

- [ ] https://github.com/wireapp/wire-ios-shared-resources/pull/15

## Notes

Currently the config points to a temporary branch of the build scripts - `wire-ios-shared-resources/circle2`. We currently have many PRs open so maybe it's best to merge this PR to develop as is and then update the config to point to `master` branch when we merge the scripts. This way we will not break building on current PRs which have an old CircleCI config and point to `wire-ios-shared-resources/master`.


mikeger:
Looks like it's working!

PR #654, opened 2017-12-13
Refactor ZMUserSession initialization

typfel:
### Issues
The `LocalNotificationDispatcher` was initialized with a nil `OperationStatus`.

### Causes
There is a dependency cycle between the `LocalNotificationDispatcher` and the `ZMOperationLoop`, where `ZMOperationLoop` depends on  dispatcher and the dispatcher depends on the operationStatus which is owned by the operation loop.

### Solutions
Break the cycle by moving the operation status (`ApplicationStatusDirectory `) to the top level (`ZMUserSession`).


Tests/Source/Synchronization/ZMSyncStrategyTests.m
```diff
@@ -931,59 +951,60 @@ - (void)testThatItNotifiesTheOperationLoopOfNewOperationWhenEnteringForeground
...
-
-    // tearDown
-    token = nil;
-
-    [self performIgnoringZMLogError:^{
-        WaitForAllGroupsToBeEmpty(0.5);
-    }];
-    [mockObserver stopMocking];
-}
+//@implementation ZMSyncStrategyTests (SyncStateDelegate)
```
vytis: Commented out code

PR #658, opened 2017-12-28
Initial bots support

mikeger:
Allows querying bots from the search and adding bots to the conversation.

Depends on:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/411
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/65
- [x] More data model changes coming
- [x] https://github.com/wireapp/wire-ios-transport/pull/74


Source/UserSession/Search/SearchTask.swift
```diff
@@ -275,3 +279,51 @@ extension SearchTask {
...
+            
+            self.session.transportSession.enqueueSearch(request)
+        }
+    }
+    
+    func servicesSearchRequest(query: String, fetchLimit: Int = 10) -> ZMTransportRequest {
+        let url = NSURLComponents()
+        url.path = "/services"
+        
+        url.queryItems = [URLQueryItem(name: "tags", value: "tutorial"), URLQueryItem(name: "start", value: query)]
```
vytis: So now we can only search for bots tagged with `tutorial`? If we don't provide any tags then server doesn't return anything?

mikeger: Hey, according to the docs, the tags parameter is mandatory. For the current version it's fine to pass `tutorial`, I am not sure if we would use this parameter in the future.

Source/UserSession/Search/SearchResult.swift
```diff
@@ -58,15 +75,30 @@ extension SearchResult {
...
     }
     
     func union(withLocalResult result: SearchResult) -> SearchResult {
-        return SearchResult(contacts: result.contacts,teamMembers: result.teamMembers, addressBook: result.addressBook, directory: directory, conversations: result.conversations)
+        return SearchResult(contacts: result.contacts, teamMembers: result.teamMembers, addressBook: result.addressBook, directory: directory, conversations: result.conversations, services: services)
     }
     
-    func union(withRemoteResult result: SearchResult) -> SearchResult {
-        return SearchResult(contacts: contacts, teamMembers: teamMembers, addressBook: addressBook, directory: result.directory, conversations: conversations)
+    func union(withRemoteResult result: SearchResult, serviceUsers: Bool) -> SearchResult {
```
typfel: I think method name is misleading. I suggest splitting it into to methods:

`union(withDirectoryResult:)`
`union(withServicesResult:)`

typfel: I guess all previous mentions of remoteResult or remoteSearch can be replaced with directoryResult.

mikeger:
Tests are green locally.

PR #660, opened 2018-01-02
Mark conversation as read or unread

mikeger:
This feature is useful to mark some conversation as the one requiring attention when you already finished reading it or marking a non-important conversation as read.

Not every conversation can be marked as unread, since it must contain at least one message with the timestamp.

I also wanted to support marking the whole account's conversations as read, and marking all accounts conversations as read. This is useful for the app icon menu, where we can add the button to mark all conversations as read.

# Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/413


vytis:
Few simple tests would be good here

Source/SessionManager/SessionManager.swift
```diff
@@ -926,3 +926,24 @@ extension SessionManager: NotificationContext {
...
+        let group = DispatchGroup()
+        
+        self.accountManager.accounts.forEach { account in
+            group.enter()
+            self.withSession(for: account) { userSession in
+                userSession.performChanges {
+                    userSession.markAllConversationsAsRead()
+                }
+                
+                group.leave()
```
vytis: Wouldn't this be called immediately after the a block is enqueued  with`userSession.performChanges`, not actually after it gets executed?

mikeger: AFAIK `performChanges` is a synchronous block.

PR #663, opened 2018-01-09
Callkit with multiple accounts

typfel:
## What's new in this PR?

This PR adds support for using Callkit with multiple accounts

The major changes:
- Move `CallKitDelegate` to the session manager
- Add global call observers for observing call state across all sessions
- Automatically switch to the session with the active call when entering app from the background.


Source/Calling/CallKitDelegate.swift
```diff
@@ -125,11 +134,13 @@ extension CallKitDelegate {
         guard contacts.count == 1,
               let contact = contacts.first,
               let customIdentifier = contact.customIdentifier,
-              let remoteIdentifier = UUID.init(uuidString: customIdentifier) else {
+              let remoteIdentifier = UUID.init(uuidString: customIdentifier),
+              let managedObjectContext = sessionManager.activeUserSession?.managedObjectContext
```
mikeger: I can see a possible issue if user has 2 user sessions, A and B, and used to have calls in B. Then he activates session A and goes to the iOS phone app, and then select the past call in the list. The call was done with B account, so the app is unable to place the call (session A is active).

typfel: Right, we could search for a matching contact in all sessions I guess.

Source/SessionManager/SessionManager.swift
```diff
@@ -333,12 +346,20 @@ public protocol LocalNotificationResponder : class {
     /// - completion: runs when the user session was loaded
     /// - tearDownCompletion: runs when the UI no longer holds any references to the previous user session.
     public func select(_ account: Account, completion: ((ZMUserSession)->())? = nil, tearDownCompletion: (() -> Void)? = nil) {
+        guard !isSelectingAccount else { return }
```
mikeger: In this case the completion is not going to be called, could this lead to bugs potentially?

Source/Calling/CallKitDelegate.swift
typfel: Ok, think I found a nice solution. I'm now also adding the `accountId` to the CallKit `customIdentifier`. That way I can always load the correct session and look for the conversation there.

Source/SessionManager/SessionManager.swift
typfel: Mm should be fine for how we use it atm but yes these not always running completion handlers are problematic. We should review the SessionManager because this happen in quite a few places.

PR #671, opened 2018-01-18
Improve notifications tracking

vytis:
### Issues

After shipping tracking implemented in #664 and looking through data there were some curious discrepancies. 

### Causes

Calls to increment counter values were called from multiple threads but there were no code to handle this.

### Solutions

Added isolation queue so that we would serialize all the calls for incrementing counters. Also added one more counter that tracks how many times the app didn't have enough time to process the notification and the background task expired.

### DIFF
--- a/Source/Analytics/NotificationsTracker.swift
+++ b/Source/Analytics/NotificationsTracker.swift
@@ -33,6 +33,7 @@ import WireDataModel
             return "notifications_" + rawValue
         }
     }
+    private let isolationQueue = DispatchQueue(label: "NotificationsProcessing")
 
     weak var analytics: AnalyticsType?
     @objc public init(analytics: AnalyticsType) {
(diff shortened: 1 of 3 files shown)


Source/Analytics/NotificationsTracker.swift
```diff
@@ -55,18 +57,26 @@ import WireDataModel
...
+        increment(attribute: .processingExpired)
+    }
+
     private func increment(attribute: Attributes, by amount: Double = 1) {
-        var currentAttributes = analytics?.persistedAttributes(for: eventName) ?? [:]
-        var value = (currentAttributes[attribute.identifier] as? Double) ?? 0
-        value += amount
-        currentAttributes[attribute.identifier] = value as NSObject
-        analytics?.setPersistedAttributes(currentAttributes, for: eventName)
+        isolationQueue.sync {
```
daehn: Why `sync` and not `async`? Doesn't seem like waiting for the execution is necessary, and as this is a serial queue the enqueued blocks will still be executed in the right order.

vytis: I wanted to make sure that background task doesn't expire until we have saved the counters. By using `async` there is a possibility that all is done processing but we still haven't done updating them.

PR #674, opened 2018-01-23
Fix/remove is encrypted

typfel:
## Issue

The `isEncrypted` has been removed from the `ZMMessage` class. So any usage of needs to be removed

## Depends on

- [x] https://github.com/wireapp/wire-ios-data-model/pull/416
- [x] https://github.com/wireapp/wire-ios-message-strategy/pull/107


WireSyncEngine.xcodeproj/xcshareddata/xcschemes/WireSyncEngine.xcscheme
```diff
@@ -120,7 +120,7 @@
          <EnvironmentVariable
             key = "ZMLOG_TAGS"
             value = "MockTransportRequests"
-            isEnabled = "YES">
+            isEnabled = "NO">
```
daehn: Not sure if this change was intended.

typfel: It should be off by default I'd say. Looks like I did set it to YES in https://github.com/wireapp/wire-ios-sync-engine/pull/647

PR #675, opened 2018-01-23
Do not crash on inactive SE push

mikeger:
### Issues

The client seem to crash when receieving the push and try to forward it to one of the user sessions.

### Causes

The user session, that is managed by the session manager, can be deactivated and deallocated in case of memory pressure. The moment when the user session is torn down and deallocated is separated in time.

The push that app receives from APNS is proposed to all existing user sessions that where created by the session manager. The session manager keeps the weak references to all user sessions that are still in memory.

The client seem to crash when receieving the push and try to forward it to the user session that was torn down, but was not deallocated yet.

### Solutions

As a temporary solution to allow users to keep receiving the pushes the client will check if the user session was not torn down yet before trying to process the push.

Long-term solution would involve investigating why the user session stays retained after it was torn down.


typfel:
@mikeger are you creating a ticket for the long term solution?

mikeger:
@typfel on it.

PR #682, opened 2018-01-30
Remove service user from conversation

vytis:
### Issues

Request for removing service user from conversation needs to be done to a different endpoint. The response to this request differs a bit from regular users removal, but this is handled in WireTransport (see related PR).

### Solutions

Checking if the user that we are about to remove is a service user and then using slightly different path. 

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/79


Tests/Source/Integration/IntegrationTest.h
```diff
@@ -55,6 +55,8 @@
 @property (nonatomic, null_unspecified) MockUser *user3; // not connected, with profile picture, in a common group conversation
 @property (nonatomic, null_unspecified) MockUser *user4; // not connected, with profile picture, no shared conversations
 @property (nonatomic, null_unspecified) MockUser *user5; // not connected, no shared conversation
+@property (nonatomic, null_unspecified) MockUser *serviceUser; // service user
+@property (nonatomic, null_unspecified) MockConversation *groupConversationWithServiceUser;
```
mikeger: Should we set those to nil to avoid the memory leaks on `deinit`?

vytis: Indeed!

PR #684, opened 2018-02-01
ZIOS-9527: Added reachability check when logging with phone.

nicorsm:
### Issues

There was no control over login requests made without an internet connection, so the app was going into a requests loop because of the lack of connection.

### Solutions

I've added the same reachability checks that are made with email login. I've also added a new test.

### DIFF
--- a/Source/UnauthenticatedSession/UnauthenticatedSession+Login.swift
+++ b/Source/UnauthenticatedSession/UnauthenticatedSession+Login.swift
@@ -56,8 +56,13 @@ extension UnauthenticatedSession {
         } catch {
             return false
         }
+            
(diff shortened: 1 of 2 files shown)


daehn:
Shouldn't there be some kind of reachability check in `wire-ios-transport-session` in place? Maybe only for the authenticated session, though.

nicorsm:
@daehn ...you mean `wire-ios-transport`? 😄

mikeger:
@daehn as far as I remember sometimes even if reachability says that the client is offline when the request is fired iOS tries to establish the connection, so completely pausing requests in transport when offline is not the ultimate solution.

PR #686, opened 2018-02-01
Improve notifications tracking

vytis:
### Issues

After shipping tracking of push notifications processing (#664) we have noticed several discrepancies in numbers.

### Causes

It is not very obvious which code path for fetching notification stream only happens when processing push notifications. There is another code path (that might not be used anymore) which handles notifications coming from AppDelegate instead of through PushKit framework.

### Solutions

Registering start of fetching stream when we create a `ZMTransportRequest` and registering finish when the completion handler is called.
Also added tracking to the codepath that is going through the AppDelegate. I assume it is not used anymore (all VoIP notifications go through PushKit), but just for completeness sake it would be good to have it there as well.

### DIFF
--- a/Source/SessionManager/PushDispatcher.swift
+++ b/Source/SessionManager/PushDispatcher.swift
@@ -177,13 +177,8 @@ public final class PushDispatcher: NSObject {
     // -application:didReceiveRemoteNotification:completionHandler:
     public func didReceiveRemoteNotification(_ payload: [AnyHashable: Any],
                                              fetchCompletionHandler: @escaping (UIBackgroundFetchResult)->()) {
-        self.notificationsTracker?.registerReceivedPush()
-        let handler: (UIBackgroundFetchResult)->() = {
-            self.notificationsTracker?.registerNotificationProcessingCompleted()
-            fetchCompletionHandler($0)
-        }
         self.remoteNotificationHandler.didReceiveRemoteNotification(payload,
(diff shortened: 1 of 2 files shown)


Source/SessionManager/PushDispatcher.swift
```diff
@@ -177,8 +177,13 @@ public final class PushDispatcher: NSObject {
     // -application:didReceiveRemoteNotification:completionHandler:
     public func didReceiveRemoteNotification(_ payload: [AnyHashable: Any],
                                              fetchCompletionHandler: @escaping (UIBackgroundFetchResult)->()) {
+        self.notificationsTracker?.registerReceivedPush()
+        let handler: (UIBackgroundFetchResult)->() = {
+            self.notificationsTracker?.registerNotificationProcessingCompleted()
```
typfel: Is this a new `registerNotificationProcessingCompleted` event? Should it have been removed from somewhere else?

vytis: This is for the flow when notifications come through app delegate. I think it was only used for fallback notifications actually, so not sure if it's even worth adding any tracking in here. I did however miss adding a call to `registerNotificationProcessingStarted()` here!

vytis: Actually I will revert this, it should not be used anyway...

PR #687, opened 2018-02-05
Improve usage of format strings

typfel:
### Issues

We have some usage of format strings where it is possible to construct incorrect format strings if great care is not taken when using the localization methods.

### Solutions

*Issue 1:* `ZMUserSession+UserNotificationCategories.m`
Here we are passing format strings as variables down a chain of methods which can be risky if the key is not used as expected. We can fix this by resolving the localized string at the top level and pass it down instead.

*issue 2:*  `ZMLocalNotificationLocalization.h`
Here we expose several localization API method  as an extension on `NSString`. These methods however only work with a certain set of localization strings defined in `Push.strings`. Knowing which methods work together with which localization strings requires careful inspection of the code and is easy to get wrong.

We will solve this replacing these methods with an enum representing every case of possible push notifications. This enum will then expose methods for creating the necessary localized strings which will always return a valid result.


Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Localization.swift
```diff
@@ -0,0 +1,270 @@
...
+        case 2:
+            return String.localizedStringWithFormat(format, arguments[0], arguments[1])
+        case 3:
+            return String.localizedStringWithFormat(format, arguments[0], arguments[1], arguments[2])
+        case 4:
+            return String.localizedStringWithFormat(format, arguments[0], arguments[1], arguments[2], arguments[3])
+        case 0:
+            fallthrough
+        default:
+            return String.localizedStringWithFormat(format)
```
vytis: We still are passing dynamic string to a format here, we should use regular `localizedString(forKey:` method instead

Tests/Source/Notifications/PushNotifications/LocalNotificationDispatcherTests.swift
```diff
@@ -304,6 +304,7 @@ extension LocalNotificationDispatcherTests {
         XCTAssertEqual(self.application.scheduledLocalNotifications.count, 1)
         XCTAssertEqual(self.notificationDelegate.receivedLocalNotifications.count, 0)
         guard let notification = self.application.scheduledLocalNotifications.first else { return XCTFail() }
+        print(notification.alertBody!)
```
vytis: Probably needs to go away:)

PR #688, opened 2018-02-07
Include callkit logs into voice log

typfel:
### Issues

The logs for CallKit weren't ending up in the voice log.

### Causes

 The CallKitDelegate was using a method for logging on the flow manager which had been de-activated.

## Notes

- Deleted the deactivated method from the FlowManager type.
- Cleaned up a few places were the flow manager was used for logging.
- Moved the AVS logging methods from the ZMUserSession to the SessionManager.

### DIFF
--- a/Source/SessionManager/SessionManager+AVS.swift
+++ b/Source/SessionManager/SessionManager+AVS.swift
@@ -1,0 +1,46 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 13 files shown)


Source/SessionManager/SessionManager+AVS.swift
```diff
@@ -0,0 +1,46 @@
...
+    
+    @objc(logMessage:)
+    func log(message : String)
+    
+}
+
+public extension SessionManager {
+    
+    @objc
+    static func addLogger(_ logger : AVSLogger) -> Any? {
```
mikeger: I think `SelfUnregisteringNotificationCenterToken` is actually not optional, so ? can be removed.

PR #693, opened 2018-02-22
ZIOS-9140: Added tracking on notifications for text and call replies

nicorsm:
### Issues

Replies sent through notifications and calls joined from notifications were not tracked via Mixpanel under the `contribution` event.

### Solutions

I've implemented an extension of `AnalyticsType` for reporting these events to Mixpanel (`func tagActionOnPushNotification(conversation: ZMConversation?, action: ConversationMediaAction)`). The properties are the same of the `contributed` event tracked on `wire-ios`.
I've moved some extension methods to `wire-ios-sync-engine`.


Source/UserSession/ZMUserSession+Actions.swift
```diff
@@ -107,10 +107,13 @@ extension ZMUserSession {
...
             self.messageReplyObserver = nil
             self.syncManagedObjectContext.performGroupedBlock {
+                
+                let conversationOnSyncContext = notification.conversation(in: self.syncManagedObjectContext)
                 if result == .failed {
                     zmLog.warn("failed to reply via push notification action")
-                    let conversationOnSyncContext = notification.conversation(in: self.syncManagedObjectContext)
                     self.localNotificationDispatcher.didFailToSendMessage(in: conversationOnSyncContext!)
+                } else {
+                    self.syncManagedObjectContext.analytics?.tagActionOnPushNotification(conversation: conversationOnSyncContext, action: .text)
```
mikeger: This could cause the threading issue, since we are here on the sync context queue, which is mot likely running on the background thread. I don't think that Mixpanel API is thread-safe, so I would suggest to switch to UI context queue that is running on the main thread.

mikeger: Same applies for other calls.

nicorsm: ...but in the `NSManagedObjectContext` extension we are requiring `Analytics` to be called from the sync context, otherwise it will throw a precondition failure:
https://github.com/wireapp/wire-ios-data-model/blob/ae73a4f7be820279ffe10238bd0381bae9d5bcda/Source/Model/Analytics/NSManagedObjectContext%2BAnalytics.swift#L27

mikeger: You are right, it can be only accessed on the sync context, please check with debugger if the call to analytics is synced somehow to the main thread, if not then we have to fix it for existing code as well.

nicorsm: Checked now, all generic calls to `tagEvent:attributes:` (including the real Mixpanel implementation) are made on the main thread:
https://github.com/wireapp/wire-ios/blob/7e5dc8dd66dfe2235b70c789e128ead186342e61/Wire-iOS/Sources/Analytics/Analytics.m#L164

PR #696, opened 2018-02-26
Handling access update event

vytis:
### Issues

When conversation access mode is update we get an update event and need to handle it.

### Solutions

Unpacking the payload and setting new access mode and role for the conversation


vytis:
Yes, the payload contains both items and they are set through the same `conversations/<conv_id>/access` endpoint

PR #697, opened 2018-02-26
Updated for the new API

mikeger:
Also added the test to verify the request generated.


Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -84,13 +84,14 @@ extension ZMConversation {
         userSession.transportSession.enqueueOneTime(request)
     }
     
-    func setMode(_ mode: ConversationAccessMode, in userSession: ZMUserSession, _ completion: @escaping (VoidResult) -> Void) {
-        guard accessMode != mode else { return completion(.success) }
-        let request = WirelessRequestFactory.setMode(for: self, mode: mode)
+    func setAllowGuests(_ allowGuests: Bool, in userSession: ZMUserSession, _ completion: @escaping (VoidResult) -> Void) {
+        guard self.allowGuests != allowGuests else { return completion(.success) }
```
vytis: Maybe it's still worth allowing the request to go through just to make sure we have the same state locally as in the server. Or it's an overkill?

mikeger: Good point.

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -84,13 +84,14 @@ extension ZMConversation {
...
+        let request = WirelessRequestFactory.set(allowGuests: allowGuests, for: self)
         request.add(ZMCompletionHandler(on: managedObjectContext!) { response in
             if let payload = response.payload,
                 let event = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil) {
                 // Process `conversation.access-update` event
+                self.allowGuests = allowGuests
```
vytis: Which thread does the completion handler run on? Can we modify the conversation safely here?

daehn: On the `managedObjectContext`'s  queue that is passed to it, so it should be fine.

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -121,11 +122,16 @@ fileprivate struct WirelessRequestFactory {
         return .init(path: "/conversations/\(identifier)/code", method: .methodPOST, payload: nil)
     }
     
-    static func setMode(for conversation: ZMConversation, mode: ConversationAccessMode) -> ZMTransportRequest {
+    static func set(allowGuests: Bool, for conversation: ZMConversation) -> ZMTransportRequest {
```
daehn: I guess this one should be marked public so it can be used from the UI.

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -84,13 +84,14 @@ extension ZMConversation {
         userSession.transportSession.enqueueOneTime(request)
     }
     
-    func setMode(_ mode: ConversationAccessMode, in userSession: ZMUserSession, _ completion: @escaping (VoidResult) -> Void) {
-        guard accessMode != mode else { return completion(.success) }
-        let request = WirelessRequestFactory.setMode(for: self, mode: mode)
+    func setAllowGuests(_ allowGuests: Bool, in userSession: ZMUserSession, _ completion: @escaping (VoidResult) -> Void) {
```
daehn: I guess this one should be marked public to be accessible from the UI.

mikeger:
@daehn @vytis updated

daehn:
@mikeger is this one still `do not merge`?

mikeger:
Can be merged!

PR #698, opened 2018-02-27
Allow guests: sync engine part

mikeger:
- New request to set the `access` and `access_role`.
- Reefetching of all conversations that must have an access mode.


Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -0,0 +1,95 @@
...
+private let zmLog = ZMSLog(tag: "ConversationLink")
+
+fileprivate extension ZMConversation {
+    struct TransportKey {
+        static let data = "data"
+        static let uri = "uri"
+    }
+}
+
+public enum WirelessLinkError: Error {
```
vytis: I guess this could be renamed/split into several as well

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -0,0 +1,95 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
```
vytis: Some of the things in this file are not really used for allow guests option, maybe worth moving to the other PR.

Source/Synchronization/ZMHotFixDirectory.m
```diff
@@ -115,6 +115,14 @@ - (NSArray *)patches
                      patchCode:^(__unused NSManagedObjectContext *context) {
                          [ZMHotFixDirectory restartSlowSync:context];
                      }],
+
+                    /// We need to refetch all team conversations to get data about access levels that were
+                    /// introduced after implementing wireless users funtionality.
```
vytis: typo - funtionality

Source/Data Model/Conversation+AccessMode.swift
mikeger: You are right, will do!

mikeger:
@vytis should be fixed now

PR #699, opened 2018-02-27
Fetch, create, delete the conversation link

mikeger:
- Code necessary for fetching the conversation links.


Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -24,7 +24,85 @@ public enum SetAllowGuestsError: Error {
...
+        case (403, "invalid-op"?): self = .invalidOperation
+        case (404, "no-conversation-code"?): self = .noCode
+        case (400..<499, _): self = .unknown
+        default: return nil
+        }
+    }
+}
+
 extension ZMConversation {
+    public func fetchWirelessLink(in userSession: ZMUserSession, _ completion: @escaping (Result<String>) -> Void) {
```
daehn: I guess this should be `Result<String?>` to indicate a successful request but no link present on the backend.

daehn: And the response parsing should be adjusted accordingly then.

PR #701, opened 2018-02-28
Incude access role and modes in conversation creation payload

daehn:
## What's new in this PR?

* Incude access role and modes in conversation creation payload.


daehn:
@vytis updated!

PR #703, opened 2018-03-07
New getter to determine if the conversation access can be managed

mikeger:
New property on `ZMConversation` to unify the logic for understanding if the current self user can change the access mode of the conversation.


mikeger:
Tests are failing.

PR #706, opened 2018-03-09
Wireless users expiration

mikeger:
- New simple class `UserExpirationObserver` is created.
- When user is opening the conversation we check if we have any wireless users, if some of them already expired we are marking them to be refetched from the backend. The refetch is triggering the user removal from the conversation.
- If we user is not yet expired, we are starting the timer that should fire at the expiration time.


Source/UserSession/OpenConversationObserver/UserExpirationObserver.swift
```diff
@@ -0,0 +1,46 @@
...
+    func check(users: Set<ZMUser>) {
+        let allWireless = Set(users.filter { $0.isWirelessUser }).subtracting(expiringUsers)
+        
+        let expired = Set(allWireless.filter { $0.isExpired })
+        expiringUsers.subtract(expired)
+        let notExpired = allWireless.subtracting(expired)
+        
+        expired.forEach { $0.needsToBeUpdatedFromBackend = true }
+        notExpired.forEach {
+            self.perform(#selector(expire(_:)), with: $0, afterDelay: $0.expiresAfter)
```
typfel: Have you thought about any possible retain issues here? Like if a not yet expired wireless user would retain this class instance and then the user switches account?

mikeger: Good point, this retains both the class and the target 😨

Source/UserSession/OpenConversationObserver/ZMUserSession+OpenConversation.swift
```diff
@@ -0,0 +1,34 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public protocol OpenConversationObserver {
+    func didOpen(conversation: ZMConversation)
```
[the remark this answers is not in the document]
mikeger: I am researching it, not 100% clear, since the message window is created in the data model. I am hesitant to move this class to data model, since something has to retain it, but:
1. Message window is not a good candidate since it's life-cycle.
2. I don't like putting the stuff into MOC.

mikeger:
@typfel please check the ZMTimer solution.

Source/UserSession/OpenConversationObserver/ZMUserSession+OpenConversation.swift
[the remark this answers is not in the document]
typfel: Ok, then let's not do it.

Source/UserSession/OpenConversationObserver/UserExpirationObserver.swift
typfel: Looks good 👍

Source/UserSession/OpenConversationObserver/UserExpirationObserver.swift
```diff
@@ -30,17 +35,22 @@ import Foundation
...
     }
     
-    func check(usersIn conversation: ZMConversation) {
-        check(users: Set(conversation.activeParticipants.array as! [ZMUser]))
+    public func timerDidFire(_ timer: ZMTimer) {
+        guard let user = timerForUser[timer] else {
+            fatal("Unknown timer: \(timer)")
+        }
+        
+        user.needsToBeUpdatedFromBackend = true
```
daehn: I guess the user can be removed completely from the dict after the timer fired.

mikeger: Actually happens some lines after, I pushed it later, may be you've seen the old diff.

daehn: Ah yeah, I checked the previous diff indeed!

PR #707, opened 2018-03-09
Added simple block for offline check

vytis:
### Issues

If we have airplane mode turned on during registration the requests are being retried continuously. When logging in we check for reachability and show error instead.

### Solutions

Unify checks for reachability in both login and registration.

### DIFF
--- a/Source/UnauthenticatedSession/UnauthenticatedSession.swift
+++ b/Source/UnauthenticatedSession/UnauthenticatedSession.swift
@@ -79,7 +79,7 @@ public class UnauthenticatedSession: NSObject {
         tornDown = true
     }
 
-    func whenReachable(_ block: () -> ()) {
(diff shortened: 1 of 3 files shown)


Source/UnauthenticatedSession/UnauthenticatedSession.swift
```diff
@@ -79,6 +79,13 @@ public class UnauthenticatedSession: NSObject {
         tornDown = true
     }
 
+    func whenReachable(_ block: () -> ()) {
```
mikeger: I think someone can read `whenReachable` as perform when reachable, may be would be more clear if named `onlyWhenReachable` or something similar?

vytis:
@mikeger renamed to `authenticationErrorIfNotReachable`

PR #708, opened 2018-03-09
Added password field to LoginCredentials, corrected handling of errors with duplicate mails

nicorsm:
## What's new in this PR?

Password was missing in the `LoginCredentials` object. This field should be passed when handling the registration of an already registered mail from iPad.
I've also handled the failing of registration with duplicate email addresses.


nicorsm:
@vytis tried to do a fresh clone of `develop`, `carthage update`, run tests on `ConversationTests+Guests.swift` and tests are failing 😔

PR #709, opened 2018-03-12
Remove fallback cancelation

typfel:
## What's new in this PR?

The BE does no longer sends notification fallbacks so it's a waste of resources to attempt cancel them when we receive a push notification.

## What changed

The `BackgroundAPNSPingBackStatus` has turned into `PushNotificationStatus` which now keep track of events which we have been notified about through push notifications. The `ZMMissingUpdateEventsTranscoder` will attempt to fetch notification stream until all the events we've been notified about have been fetched.

## Notes

I've marked a few `TODO:s` where I need investigate more. Also we might need to make the `PushNotificationStatus` a bit more resilient to out of order pushes because there's a potential for request loops if we receive a push with an event id which we have already fetched.


Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
```diff
@@ -365,7 +332,7 @@ - (NSUUID *)nextUUIDFromResponse:(ZMTransportResponse *)response forListPaginato
         // The fetch of the notification stream was initiated after the push channel was established
         // so we must restart the fetching to be sure that we haven't missed any notifications.
         if (syncStatus.pushChannelEstablishedDate.timeIntervalSinceReferenceDate < self.listPaginator.lastResetFetchDate.timeIntervalSinceReferenceDate) {
-            [syncStatus finishCurrentSyncPhaseWithPhase:self.expectedSyncPhase];
+            [syncStatus finishCurrentSyncPhaseWithPhase:self.expectedSyncPhase]; // TODO jacob: I think we should always call this method if we are done. SyncStatus keep track if we need to restart or not.
```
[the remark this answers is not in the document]
typfel: After looking more closely at this, it is not wrong. When the `if statement` is false we will attempt to fetch fetch the notification stream one more time.

Source/Synchronization/ZMOperationLoop+Background.m
```diff
@@ -22,131 +22,66 @@
...
+- (void)fetchEventsFromPushChannelPayload:(NSDictionary *)payload completionHandler:(ZMPushResultHandler)completionHandler source:(ZMPushNotficationType)source
 {
     ZMLogDebug(@"----> Received push notification payload: %@, source: %lu", payload, (unsigned long)source);
     
     [self.syncMOC performGroupedBlock:^{
         if (![self notificationIsForCurrentUser:payload]) {
             ZMLogDebug(@"Push is for the different user: skipping");
             if (nil != completionHandler) {
-                completionHandler(ZMPushPayloadResultNoData);
+                completionHandler(ZMPushPayloadResultNoData); // TODO jacob: should we really call the completion handler? That would end the background activity
```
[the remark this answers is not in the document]
typfel: Confirmed that this code path is not called anymore can be removed.

Source/UserSession/PushNotificationStatus.swift
```diff
@@ -0,0 +1,108 @@
...
+    /// - parameter eventId: UUID of the event to fetch
+    /// - parameter completionHandler: The completion handler will be run when event has been downloaded and when there's no more events to fetch
+    @objc(fetchEventId:completionHandler:)
+    public func fetch(eventId: UUID, completionHandler: @escaping (ZMPushPayloadResult) -> Void) {
+        guard eventId.isType1UUID else {
+            return zmLog.error("Attempt to fetch event id not conforming to UUID type1: \(eventId)")
+        }
+        // add eventId to list of events to fetch
+        
+        if let order = managedObjectContext.zm_lastNotificationID?.compare(withType1: eventId), order == .orderedDescending || order == .orderedSame {
```
vytis: Would it be an option to either subclass NSUUID (UUID is a struct AFAIK?) and add `Comparable` to it? Or even to regular UUID? Or would it then break down when we have regular-style UUIDs?

Either way this code seems to be duplicated in quite a few places.

typfel: It's duplicated in one place in this file. I can extract the logic  it into a private function.

typfel: After discussion: Yes, it would cleaner to create a separate type for UUID type 1 but it's bigger task which I wouldn't do in the scope of this PR.

typfel: After discussion: Yes, it would cleaner to create a separate type for UUID type 1 but it's bigger task which I wouldn't do in the scope of this PR.

PR #710, opened 2018-03-13
Fixed integration tests for conversation with guests

nicorsm:
## What's new in this PR?

We were using the `.allowGuests` property for checking if the conversation can accept guests from outside. Since the purpose of this flag has changed during time, I'm now checking for the correct values of the `accessMode` and `accessRole` properties.


Tests/Source/Integration/ConversationTests+Guests.swift
```diff
@@ -82,7 +82,8 @@ class ConversationTests_Guests : TeamTests {
         let conversation = self.conversation(for: mockConversation)!
 
         XCTAssert(waitForAllGroupsToBeEmpty(withTimeout: 0.1))
-        XCTAssertTrue(conversation.allowGuests)
+        XCTAssertTrue(conversation.accessMode == [.code, .invite])
```
vytis: Better to use `XCTAssertEqual` in this case, produces more readable failure messages

PR #712, opened 2018-03-15
Move isGroup and isConstantBitRate into CallSnapshot

typfel:
## What's new in this PR?

Move some state properties into the `CallSnapshot`: `isGroup` and `isConstantBitRate`. This simplifies the interface and makes sure we don't keep any state around after the call is over.


mikeger:
@typfel should be merged? Or some things to fix in the tests?

typfel:
@mikeger I'll need to check the tests. This is not urgent it's not a fix for anything.

mikeger:
Sure, just afraid that we'll have to merge it later, but in fact there is no hurry.

mikeger:
@typfel ;)

typfel:
@mikeger yes, let me update the tests :)

typfel:
@mikeger ok finally updated this :)

PR #713, opened 2018-03-15
Adjust http status code when wireless link already exists remotely

daehn:
## What's new in this PR?

 * Adjust http status code when wireless link already exists remotely.

### This PR depends on:

- [x] A new release of `wire-ios-mock-transport` after https://github.com/wireapp/wire-ios-mocktransport/pull/77 has been merged.


Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -110,7 +110,7 @@ extension ZMConversation {
                     }
                 }
             }
-            else if response.httpStatus == 204,
+            else if response.httpStatus == 200,
```
mikeger: I believe that the mock transport needs to be adjusted as well.

mikeger: Oops, missed your other commit 🗡

PR #714, opened 2018-03-16
Dispatch to the correct thread

mikeger:
It appears that ZMTimer is calling it's callback on the background thread. Now the call is dispatched to the correct thread.


daehn:
@mikeger tests failed to compile!

PR #715, opened 2018-03-21
Add strategy and status to remotely delete v3 assets

daehn:
## What's new in this PR?

* Add strategy and status to remotely delete v3 assets.

### This PR depends on:

- [x] A new release of `wire-is-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/450 has been merged.


Source/Synchronization/Strategies/Asset Deletion/AssetDeletionRequestStrategy.swift
```diff
@@ -0,0 +1,71 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+fileprivate struct RequestFactory {
+    static func request(for identifier: String, on queue: ZMSGroupQueue, block: @escaping ZMCompletionHandlerBlock) -> ZMTransportRequest {
```
vytis: At the moment all asset requests code is in Request Strategy, so maybe for consistency sake it's worth putting it there?

Source/Synchronization/Strategies/Asset Deletion/AssetDeletionRequestStrategy.swift
```diff
@@ -0,0 +1,71 @@
...
+        super.init(withManagedObjectContext: context, applicationStatus: applicationStatus)
+        requestSync = ZMSingleRequestSync(singleRequestTranscoder: self, groupQueue: context)
+        
+    }
+    
+    private func handle(response: ZMTransportResponse, for identifier: String) {
+        switch response.result {
+        case .success: identifierProvider.didDelete(identifier: identifier)
+        case .permanentError: identifierProvider.didFailToDelete(identifier: identifier)
+        default: break
```
vytis: So all other errors should hopefully not reach us and transport will retry until it's `.success` or `.permanentError`?

Source/Synchronization/Strategies/Asset Deletion/AssetDeletionRequestStrategy.swift
daehn: I thought about this, but I don't really want to put the associated status in there as well which would then split the two in separate frameworks. A another reason to keep it in SE would be as it is only needed in the main UI project and not in any extensions.

Source/Synchronization/Strategies/Asset Deletion/AssetDeletionRequestStrategy.swift
daehn: Yeah, the sync object should already retry on temporary errors.

Source/Synchronization/Strategies/Asset Deletion/AssetDeletionRequestStrategy.swift
daehn: Basically there aren't really any status objects in message strategy, but I agree and I guess that shouldn't necessarily be obstacle and we could move these files there to have all asset related code in one place.

vytis: I was thinking of maybe moving only this function into `AssetRequestFactory`. Status is fine to be here I think.

daehn: Ah sorry totally misread your comment in that case 😂, thought you meant moving the strategy into the message-strategy framework.

vytis: When I read it again, I actually meant the way you understood and answered :D Basically I agree with the fact that we don't need it in share extension and it might as well live in SE. So all is good from my side :)

PR #717, opened 2018-03-22
Proposed API for backup restoring

mikeger:
## What's new in this PR?

The new API expects to receive the URL to the archived backup to be restored. On successful restore the callback would be called. The callback would receive the instance of Account that would represent the newly restored account or an error.

### DIFF
--- a/Source/SessionManager/SessionManager+Restore.swift
+++ b/Source/SessionManager/SessionManager+Restore.swift
@@ -26,9 +26,9 @@ public enum BackupRestoreError: Error {
 
 extension SessionManager {
     /// Restores the account database from the Wire iOS database back up file.
-    /// @param callback called when the restoration is ended. If success, Result.success with the new restored account
+    /// @param completion called when the restoration is ended. If success, Result.success with the new restored account
     /// is called.
-    public func restoreFromBackup(at location: URL, callback: (Result<Account>)->()) {
-        callback(Result.failure(BackupRestoreError.fileError))
+    public func restoreFromBackup(at location: URL, completion: (Result<Account>)->()) {
+        completion(Result.failure(BackupRestoreError.fileError))
     }
 }
--- a/Source/SessionManager/SessionManager+Restore.swift
+++ b/Source/SessionManager/SessionManager+Restore.swift
@@ -25,9 +25,10 @@ public enum BackupRestoreError: Error {
 }
 
 extension SessionManager {
-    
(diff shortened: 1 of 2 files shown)


Source/SessionManager/SessionManager+Restore.swift
```diff
@@ -0,0 +1,33 @@
...
+
+public enum BackupRestoreError: Error {
+    case versionTooNew
+    case versionTooOld
+    case fileError
+}
+
+extension SessionManager {
+    
+    public func restoreFromBackup(at location: URL, callback: (Result<Account>)->()) {
```
vytis: It will probably have to be URL to the archive file, would be good if that reflected in the parameter more explicitly

Source/SessionManager/SessionManager+Restore.swift
```diff
@@ -0,0 +1,34 @@
...
+    case versionTooNew
+    case versionTooOld
+    case fileError
+}
+
+extension SessionManager {
+    /// Restores the account database from the Wire iOS database back up file.
+    /// @param callback called when the restoration is ended. If success, Result.success with the new restored account
+    /// is called.
+    public func restoreFromBackup(at location: URL, callback: (Result<Account>)->()) {
```
typfel: What about renaming `callback` to `completion` for consistency?

Source/SessionManager/SessionManager+Restore.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public enum BackupRestoreError: Error {
+    case versionTooNew
+    case versionTooOld
```
daehn: Not sure in which case the `versionToOld` error will be throws, currently the metadata verification can throw the following errors:

* `BackupMetadata.VerificationError.userMismatch`
* `BackupMetadata.VerificationError.backupFromNewerAppVersion`

mikeger: I introduced this before thinking in detail about it, I guess if we have this Error enum from data model we don't need this error enum anyways: we can just rethrow what we get from the data model?

vytis: at least on https://github.com/wireapp/wire-ios-sync-engine/pull/718 I am doing the same - only adding one additional error that could happen at SE level.

PR #721, opened 2018-03-27
Unzip history backup file on import

daehn:
## What's new in this PR?

* Unzip history backup file on import.


Source/SessionManager/SessionManager+Backup.swift
```diff
@@ -48,16 +53,50 @@ extension SessionManager {
...
             }
         }
     }
     
+    // MARK: - Import
+    
+    /// Restores the account database from the Wire iOS database back up file.
+    /// @param completion called when the restoration is ended. If success, Result.success with the new restored account
+    /// is called.
+    public func restoreFromBackup(at location: URL, with userId: UUID, completion: @escaping RestoreResultClosure) {
```
vytis: So is this the API that UI will use? It doesn't really know about user IDs, so we probably need to wrap it somehow to get the user ID of the active account. But also we need to create the account right after login response is received, but before the cookie is extracted.
I think that we can keep it as is for now and update later when we have the authentication changes completed.

daehn: @vytis yeah this API isn't final yet, depends a lot on what we have available at this point.

PR #731, opened 2018-04-04
Correctly handle the import error

mikeger:
The fact that import error was ignored was overlooked during the merge.


Source/SessionManager/SessionManager+Backup.swift
```diff
@@ -88,7 +88,15 @@ extension SessionManager {
                 from: url,
                 applicationContainer: self.sharedContainerURL,
                 dispatchGroup: self.dispatchGroup,
-                completion: { _ in complete(.success) }
+                completion: { result in
```
typfel: @daehn @mikeger  there's is no shorthand for result forwarding?

daehn: If they are of the same type (which I think they are in this case) then it can be forwarded directly.

daehn: Ah never mind, one is a VoidResult and the other one a Result<T>.

vytis: So maybe worth adding to WireUtilities - makes sense if we care only that it doesn't return error. What other result implementations do in this case?

vytis: Nevermind, @daehn is faster to make a PR than I can type - https://github.com/wireapp/wire-ios-utilities/pull/38

daehn: I added a helper property to `Result` to make this transformation easier: https://github.com/wireapp/wire-ios-utilities/pull/38.

mikeger: Oh cool, thanks for the helper.

PR #733, opened 2018-04-05
Include user handle in the backup name

mikeger:
The name specified as `Wire-<user_handle>-Backup_Date.ios_wbu`.


typfel:
It's not possible to lookup the handle and throw the error further down the stack to avoid passing it as a argument?

mikeger:
@typfel it's all static methods unfortunately, I don't feel like accessing the global state from them.

typfel:
Ok 👍

PR #740, opened 2018-04-11
Refactor removing/adding participants

typfel:
## What's new in this PR?

Re-factor adding an removing participants to based on a single request which either succeeds or fails.

### Issues

The existing code for adding/removing participants is based on maintaining two sets of participants, local and remote. This enabled adding and removing people in offline, unfortunately it's also slightly complicated and brittle which has caused clients to erroneously remove all participants from conversations.

### Causes

The cause for this bug has proven hard to pin point is likely trigged by some edge case in core data.

### Solutions

We've decided to simplify this code and not maintain a "unsynced" state of the participants in a conversation.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/471


Source/Data Model/Conversation+Participants.swift
```diff
@@ -0,0 +1,148 @@
...
+                if let payload = response.payload, let event = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil) {
+                    userSession.syncManagedObjectContext.performGroupedBlock {
+                        ZMConversation(remoteID: conversationId, createIfNeeded: false, in: userSession.syncManagedObjectContext)?.updateLastRead(fromPostPayloadEvent: event)
+                        userSession.operationLoop.syncStrategy.processUpdateEvents([event], ignoreBuffer: true)
+                    }
+                }
+                
+                completion(.success)
+            }
+            else if  response.httpStatus == 204 {
```
daehn: Additional whitespace after the `if␣␣`

Tests/Source/MessagingTest.m
```diff
@@ -142,11 +145,30 @@ - (void)setUp;
...
     self.mockOperationStatus = [[OperationStatus alloc] init];
     self.mockOperationStatus.isInBackground = NO;
+    self.mockOperationLoop = [OCMockObject niceMockForClass:ZMOperationLoop.class];
+    self.mockSyncStrategy = [OCMockObject niceMockForClass:ZMSyncStrategy.class];
+    
+    [[[self.mockOperationLoop stub] andReturn:self.mockSyncStrategy] syncStrategy];
+    
+    ZM_WEAK(self);
+    void(^recordProcessedEvents)(NSInvocation *) = ^(NSInvocation * invocation) {
+//        NOT_USED(invocation);
```
daehn: Can be removed?

daehn:
@typfel is there a corresponding UI PR yet?

typfel:
@daehn haven't pushed it yet but will do it now.

mikeger:
@typfel can be merged?

typfel:
@mikeger need to bump data model but Jenkins is not co-operating

daehn:
@mikeger do we have a UI PR for this yet? I wouldn't merge it before that one is ready to be merged as well.

mikeger:
Jacob prepared something, I am checking https://github.com/wireapp/wire-ios/tree/fix/participant-update-ZIOS-9849

PR #741, opened 2018-04-16
Fix broken tests on CI

typfel:
### Issues

The newly added backup integration tests were failing

### Causes

The backup tests were relying on on-disk persistence which didn't correctly configure the dispatch groups with the core data contexts. This was causing the initial sync not to finish before the tests tried to perform a backup.

### Solutions

- Correctly configure the dispatch group for on-disk persistance
- Remove workaround for creating a self client.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/474


typfel:
Some other tests are also flaky but we'll tackle those in another PR.

PR #744, opened 2018-04-17
Fix: app icon badge number does not show notification

vytis:
### Issues

The app icon was not showing unread conversation for the first notification you get. Only after second one it shows up.

### Causes

Unread conversations were counted when notification stream is done fetching. This was done in UI context, but at that point the changes are not yet saved and merged into UI context. So in this situation unread count on UI context is still 0, while on sync context there are unread conversations

### Solutions

Calculate the unread count in sync context and pass the number back to UI for updating the badge.

### DIFF
--- a/Source/Public/ZMUserSession.h
+++ b/Source/Public/ZMUserSession.h
@@ -47,6 +47,7 @@
 @protocol LocalStoreProviderProtocol;
 @protocol FlowManagerType;
 @protocol SessionManagerType;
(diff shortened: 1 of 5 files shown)


vytis:
Will try to add some tests for this

vytis:
I give up, everything is mocked there so it would need extensive changes to whole ZMUserSessionTests set up to make it work.

PR #746, opened 2018-04-20
Update due to API changes

typfel:
## What's new in this PR?

Update due to API changes

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-cryptobox/pull/20


Source/SessionManager/SessionManager+Backup.swift
```diff
@@ -100,10 +101,11 @@ extension SessionManager {
             guard let `self` = self else { return }
             let decryptedURL = SessionManager.temporaryURL(for: location)
             do {
-                try SessionManager.decrypt(from: location, to: decryptedURL, password: password)
+                try SessionManager.decrypt(from: location, to: decryptedURL, password: password, accountId: userId)
             } catch {
                 switch error {
                 case ChaCha20Encryption.EncryptionError.decryptionFailed: return complete(.failure(BackupError.decryptionError))
+                case ChaCha20Encryption.EncryptionError.keyGenerationFailed: return complete(.failure(BackupError.keyCreationFailed))
```
vytis: In UI we handle some of these errors, no changes needed there?

typfel: When I looked at the UI code it only cares if it's a `.decryptionFailed` or not. This line I added for keeping same behaviour is actually unnecessary.

PR #753, opened 2018-05-02
ZIOS-9942: Added shouldSwitchAccounts to check if there's an ongoing call

nicorsm:
## What's new in this PR?

With this PR I'm saving the current `CallState` object provided by the `WireCallCenterCallStateObserver` APIs. When UI asks for switching accounts, I'm checking if the state is not `.estabilished` (which should be the state for ongoing calls).


Source/SessionManager/SessionManager.swift
```diff
@@ -947,3 +949,12 @@ extension SessionManager {
         }
     }
 }
+
+
+extension SessionManager {
+    
+    public var shouldSwitchAccounts: Bool {
```
daehn: There could also be a new delegate with a method with a completion closure that the UI can provide, this way we would only have to implement this once and not think about adding checks on the UI for each new case in which we have to be careful with switching accounts.

E.g. something like `confirmSwitchingAccount(completion: (Bool) -> Void)` that maybe only get's called when there is a call ongoing. Then on the UI side the delegate providing this method could call `SelfProfileViewController.displaySwitchAccountAlertIfNeeded` from within that method.

nicorsm: We've decided to keep the current solution, since currently we're handling two cases - adding and switching accounts - and the latter one is a little bit complicated: the UI is dismissing the view controller before switching the account, so the logic cannot be included there.

Source/SessionManager/SessionManager.swift
```diff
@@ -947,3 +949,12 @@ extension SessionManager {
...
+        return currentCallState != .established
```
typfel: I think it would be safer to query the current current call state of the active user session like this:

`return callCenter.activeCallConversations(in: activeUserSession).isEmpty`

Source/SessionManager/SessionManager.swift
typfel: The current solution won't work a account switch is triggered by tapping on a push notification or a chat head. Maybe we should discuss how can make the delegate approach work.

Source/SessionManager/SessionManager.swift
```diff
@@ -947,3 +966,25 @@ extension SessionManager {
...
+
+
+extension SessionManager {
+    
+    public func askCallingConfirmationIfNeeded(completion: @escaping ()->Void) {
+        if shouldSwitchAccounts {
+            completion()
+        } else {
+            self.switchingDelegate?.confirmSwitchingAccount(completion: { (completed) in
+                if completed {
```
mikeger: I would rename it to `confirmed`, so it's more clear what is going on.

daehn:
@typfel looks good to me, but seems like one test is flaky there?

nicorsm:
@typfel thanks for taking charge of the PR!

daehn:
@typfel @nicorsm I kicked off a new minor release.

typfel:
@daehn I think fastline failed for some reason, all the tests passed.

PR #758, opened 2018-05-18
BE API for user consent

mikeger:
NB: this API is only available on staging backend.

Adds the support for getting and setting the user marketing consent state.

Tests would come as the second PR. Was tested against the real backend.

### DIFF
--- a/Source/Data Model/ZMUser+Consent.swift
+++ b/Source/Data Model/ZMUser+Consent.swift
@@ -36,31 +36,32 @@ extension ZMUser {
         fetchConsent(for: .marketing, in: userSession, completion: completion)
     }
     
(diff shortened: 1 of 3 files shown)


typfel:
I understand that might not be feasible for the hot fix but would it make sense in the feature to make this part of the regular user strategy, slow sync and so on?

mikeger:
@typfel according to the current plan, we are going to fetch the value whenever we would like to display it for the user, like the conversation link, since the other client might change it.

typfel:
@mikeger there's no `user.update` events trigged when it changes?

mikeger:
@typfel I don't know, I follow the architecture guide.

PR #764, opened 2018-06-01
Unarchive cleared conversations when a call is incoming

daehn:
## What's new in this PR?

* Unarchive cleared conversations when a call is incoming.

### DIFF
--- a/Source/UserSession/CallStateObserver.swift
+++ b/Source/UserSession/CallStateObserver.swift
@@ -164,7 +164,7 @@ extension CallStateObserver : WireCallCenterCallStateObserver, WireCallCenterMis
     
     private func updateConversation(_ conversation: ZMConversation, with callState: CallState) {
         switch callState {
-        case (.incoming(_, shouldRing: true, degraded: _), true): conversation.isArchived = false
+        case .incoming(_, shouldRing: true, degraded: _): conversation.isArchived = false
         default: break
         }
     }
--- a/Source/UserSession/CallStateObserver.swift
+++ b/Source/UserSession/CallStateObserver.swift
@@ -163,7 +163,7 @@ extension CallStateObserver : WireCallCenterCallStateObserver, WireCallCenterMis
     }
     
     private func updateConversation(_ conversation: ZMConversation, with callState: CallState) {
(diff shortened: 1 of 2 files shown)


Source/UserSession/CallStateObserver.swift
```diff
@@ -158,5 +161,12 @@ extension CallStateObserver : WireCallCenterCallStateObserver, WireCallCenterMis
             self.syncManagedObjectContext.enqueueDelayedSave()
         }
     }
+    
+    private func updateConversation(_ conversation: ZMConversation, with callState: CallState) {
+        switch (callState, nil != conversation.clearedTimeStamp) {
+        case (.incoming(_, shouldRing: true, degraded: _), true): conversation.isArchived = false
```
typfel: This is going to sync the archived state to other clients, maybe that's the right thing to do we just don't do it when you unarchive a conversation due to a new message for example.

Source/UserSession/CallStateObserver.swift
```diff
@@ -158,5 +161,12 @@ extension CallStateObserver : WireCallCenterCallStateObserver, WireCallCenterMis
             self.syncManagedObjectContext.enqueueDelayedSave()
         }
     }
+    
+    private func updateConversation(_ conversation: ZMConversation, with callState: CallState) {
+        switch (callState, nil != conversation.clearedTimeStamp) {
```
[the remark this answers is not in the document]
daehn: Yeah I wasn't sure about that, but yeah might be more consistent.

daehn:
@mikeger that's why I added the `More testing` label 😉

mikeger:
Yeah I am blind, saw it later 👓

Source/UserSession/CallStateObserver.swift
daehn: @typfel but syncing it is good or not? We don't want the state to get out of sync on different clients I assume.

typfel: @daehn yes, I think in other cases it was an optimization since all clients will receive the new message and unarchive on their own. But AVS discards incoming calls if they are received late so syncing is probably good here I agree.

PR #772, opened 2018-06-18
Refactor push token registration

typfel:
## What's new in this PR?

We currently only rely on voIP push notification but we still had code paths for registering regular 
push notifications. This PR removes those code paths and removes some abstraction around push token registrations which are no longer necessary since we now only work with the PushKit framework.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios/pull/2286

## Notes

Additionally:

* The push notification category setup has been moved from the user session to the session manager and converted to swift (`PushNotificationCategory.swift`)


Source/SessionManager/SessionManager+Push.swift
```diff
@@ -17,44 +17,32 @@
...
-    public func registerSessionForRemoteNotificationsIfNeeded(_ session: ZMUserSession) {
+    
+    public func updatePushToken(for session: ZMUserSession) {
         session.managedObjectContext.performGroupedBlock {
             // Refresh the tokens if needed
-            self.pushDispatcher.lastKnownPushTokens.forEach { type, actualToken in
-                switch type {
-                case .voip:
-                    if actualToken != session.managedObjectContext.pushKitToken?.deviceToken {
-                        session.managedObjectContext.pushKitToken = nil
```
typfel: @mikeger  I no longer reset `pushKitToken` to `nil` before setting the new one and I removed the check for if the deviceToken hasn't changed. I did this because from what I could see this also checked in `- (void)setPushKitToken:(NSData *)deviceToken` (ZMUserSession.m:507).

Did I miss something here?

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -131,20 +119,60 @@ extension SessionManager: ZMRequestsToOpenViewsDelegate {
...
+            let account = self.accountManager.account(with: accountId),
+            let activity = BackgroundActivityFactory.sharedInstance().backgroundActivity(withName: "Process PushKit payload", expirationHandler: { [weak self] in
+                log.debug("Processing push payload expired")
+                self?.notificationsTracker?.registerProcessingExpired()
+            }) else {
+            return completion()
         }
-            
+        
+        notificationsTracker?.registerReceivedPush()
```
vytis: I think this should be the first thing to be called after payload is received. Not sure what to do if any of the checks above fail and it's not a valid notification. Maybe worth adding another step in tracking? We then could see how many notifications arrive for wrong account, not .voip etc.

Tests/Source/Integration/APNSTests.m
```diff
@@ -127,33 +116,13 @@ - (void)testThatItUpdatesNewTokensIfNeeded
...
-    id mockPushRegistrant = [OCMockObject niceMockForClass:ZMPushRegistrant.class];
-    [(ZMPushRegistrant *)[[mockPushRegistrant expect] andReturn:newToken] pushToken];
-
-    [[[mockPushRegistrant stub] andReturn:mockPushRegistrant] alloc];
-    (void)[[[mockPushRegistrant stub] andReturn:mockPushRegistrant] initWithDidUpdateCredentials:OCMOCK_ANY
-                                                                               didReceivePayload:OCMOCK_ANY
-                                                                              didInvalidateToken:OCMOCK_ANY];
-
-    // when
+    // when]
```
vytis: Typo :)

Source/SessionManager/SessionManager+Push.swift
typfel: So apparently the PushKit is also used to communicate with the watch complications  and file management. They have nothing to do with push notifications. But for catching a drop off for not finding the account or creating the background task I agree that it might we worth adding another tracking step.

Source/UserSession/ZMUserSession+Actions.swift
```diff
@@ -172,7 +187,10 @@ private let zmLog = ZMSLog(tag: "Push")
         }
         
         enqueueChanges {
-            guard let reaction = ZMMessage.addReaction(.like, toMessage: message) else { return }
+            guard let reaction = ZMMessage.addReaction(.like, toMessage: message) else {
+                print("AAAaaaahahhhhh")
```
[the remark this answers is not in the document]
typfel: 😬

Source/SessionManager/SessionManager.swift
```diff
@@ -331,22 +335,26 @@ public protocol SessionManagerSwitchingDelegate: class {
...
         } else {
             self.notificationsTracker = nil
         }
-
+        
         super.init()
         
-        self.pushDispatcher.fallbackClient = self
+        // Configure push notification categories and register for voIP push notifications
+        self.application.registerUserNotificationSettings(UIUserNotificationSettings(types: [.alert, .badge, .sound], categories: PushNotificationCategory.allCategories))
```
typfel: This needs to move into function is called by the UI when we want ask the user about notification permissions.

Source/SessionManager/SessionManager+Push.swift
mikeger: It does not seems to be necessary indeed, the setter handles the logic.

Source/UserSession/ZMUserSession+Actions.swift
```diff
@@ -22,12 +22,27 @@ private let zmLog = ZMSLog(tag: "Push")
 
 @objc extension ZMUserSession {
     
+    public func handleCategoryNotification(_ note: ZMStoredLocalNotification) {
```
mikeger: I think the method name must not include the word "Category", before we had several methods, like handleThisCategory or handleThatCategory, when we are not stating the category name in the method name it looks like it is not necessary to say "Categoty" any more.

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -17,44 +17,37 @@
...
 @objc extension SessionManager {
-    public func registerSessionForRemoteNotificationsIfNeeded(_ session: ZMUserSession) {
+    
+    @objc public func configureUserNotifications() {
+        // Configure push notification categories
+        self.application.registerUserNotificationSettings(UIUserNotificationSettings(types: [.alert, .badge, .sound], categories: PushNotificationCategory.allCategories))
+    }
+    
+    public func updatePushToken(for session: ZMUserSession) {
         session.managedObjectContext.performGroupedBlock {
```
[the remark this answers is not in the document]
typfel: But I still need to ensure that I access the `managedObjectContext` from the right thread, which setting pushKitToken will do.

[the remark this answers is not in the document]
typfel: But yes, where we call this method we are sure that we are on the right thread.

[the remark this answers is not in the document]
typfel: So I think it's safest to keep the `performGroupedBlock` since this is a public method.

PR #777, opened 2018-06-22
Fixed the hotfix tests

mikeger:
### Issues

Original migration test was not taking in account the new migration.


mikeger:
@zenkins release major

PR #781, opened 2018-06-25
Added logic to persist the timer system message on datamodel

nicorsm:
## What's new in this PR?

Added call to `appendMessageTimerUpdateMessage()` method of `ZMConversation`, so in this way we can persist locally the system message for timer changes.

## Needs release with

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/502


Source/Synchronization/Transcoders/ZMConversationTranscoder.swift
```diff
@@ -34,13 +34,21 @@ extension ZMConversationTranscoder {
...
         if let timeoutIntegerValue = payload["message_timer"] as? Int {
             let timeoutValue = MessageDestructionTimeoutValue(rawValue: TimeInterval(timeoutIntegerValue))
             conversation.messageDestructionTimeout = .synced(timeoutValue)
         } else {
             conversation.messageDestructionTimeout = nil
         }
+        
+        if let user = ZMUser(remoteID: senderUUID, createIfNeeded: false, in: managedObjectContext),
+            let timestamp = event.timeStamp(),
+            !user.isSelfUser {
```
daehn: We don't want to show the system message if the self user changes the value?

PR #784, opened 2018-06-26
Fix/timed message system message

johnxnguyen:
### Issues

If a local timeout has been set (to say, 5 minutes), then setting a global timer to the value `none` generates a system message stating "you set the timed messages to 5 minutes" instead of "you set the timed messages to Off".

### Causes

The system message is created when `ZMConversationTranscoder` processes the `conversationMessageTimerUpdate` event. After having set the `messageDestructionTimeout` to `.synced(0)`, the timer value for the system message is read from `messageDestructionTimeoutValue`. This returns (as specified in the [DataModel](https://github.com/wireapp/wire-ios-data-model/blob/cc158917ee9d10c1895f57e95807f1364e0abaf0/Source/Model/Conversation/ZMConversation%2BEphemeral.swift#L176)) first the nonzero synced timeout value, else the nonzero local timeout value, or finally 0. Although we have cleared the sync timeout, the local one still exists and its value is passed on to the system message.

### Solutions

Instead of getting the first available timeout value, check that `messageDestructionTimeout` is a synced timeout and pass its value (or 0 if the timeout is local) to the system message.

### DIFF
--- a/Source/Synchronization/Transcoders/ZMConversationTranscoder.swift
+++ b/Source/Synchronization/Transcoders/ZMConversationTranscoder.swift
@@ -46,7 +46,17 @@ extension ZMConversationTranscoder {
         
         if let user = ZMUser(remoteID: senderUUID, createIfNeeded: false, in: managedObjectContext),
             let timestamp = event.timeStamp() {
-            let timer = conversation.hasSyncedMessageDestructionTimeout ? conversation.messageDestructionTimeoutValue : 0
+            
+            let timer: TimeInterval
+            
+            // system message timer should reflect the synced value, not local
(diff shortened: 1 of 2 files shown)


zenkins:

johnxnguyen:
The tests are passing for me locally, these failing tests seem to be unrelated. Anyone else having issues with them? 

- testThatItDoesNotProcessTheSameEventsTwiceWhenCalledSuccessively() - EventDecoderTest
- testThatOnlyGroupConversationsAreUpdated() - ZMHotFixDirectoryTests

mikeger:
@johnxnguyen please release with the CI. Thanks!

johnxnguyen:
@zenkins release patch

PR #785, opened 2018-06-26
Avoid double timer system messages

nicorsm:
## What's new in this PR?

When a user changes the timer of a conversation, the system message is appended twice. This is because SE is considering the change sent from `ZMTransportSession`. I'm discarding the message if there are changes and if they're made by the current user.

**Problem**: this solution doesn't work with _Off_ - in this case, the change is not ignored because here `.none` is different than `nil`:
https://github.com/wireapp/wire-ios-sync-engine/blob/4bef033b9ee20872d9c69b03c9e6682b550b6d10/Source/Synchronization/Transcoders/ZMConversationTranscoder.swift#L50-L52

### DIFF
--- a/Tests/Source/Synchronization/Transcoders/ZMConversationTranscoderTests.swift
+++ b/Tests/Source/Synchronization/Transcoders/ZMConversationTranscoderTests.swift
@@ -465,7 +465,7 @@ extension ZMConversationTranscoderTests_Swift {
         }
     }
     
-    func testThatItDiscardsDoubleSystemMessageWhenSyncedTumeoutChanges_Value() {
+    func testThatItDiscardsDoubleSystemMessageWhenSyncedTimeoutChanges_Value() {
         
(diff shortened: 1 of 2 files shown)


mikeger:
@nicorsm I think adding some tests for the cases would be great.

Tests/Source/Synchronization/Transcoders/ZMConversationTranscoderTests.swift
```diff
@@ -464,6 +464,99 @@ extension ZMConversationTranscoderTests_Swift {
             XCTAssertEqual(self.localNotificationDispatcher.processedMessages.last, message)
         }
     }
+    
+    func testThatItDiscardsDoubleSystemMessageWhenSyncedTumeoutChanges_Value() {
```
vytis: typo - Tumeout

nicorsm:
@zenkins release patch

PR #787, opened 2018-06-28
Fix/turning off global timer

johnxnguyen:
### Issues

Turning off the synced timeout while a local timeout already existed produced duplicate system messages and also turned off the local timeout.

### Causes

Changes from [this PR](https://github.com/wireapp/wire-ios-sync-engine/pull/786) resulted in getting a nil value from the event payload, which was subsequently written to the conversations timeout property. This meant that any existing local timeout was nullified.

The duplicate system message checks by @nicorsm seemed to be missing the case when a local timer existed, and the synced timer was turned off. The first update event was processed correctly, whereby the conversation timeout was set to `.synced(.none)`, but upon the second event, reading the conversations timeout property would return the local timeout, and therefore the duplication check passed.

### Solutions

Upon unpacking the event payload, if the timeout value is nil, set it to 0. In the duplication checks, return early if the new timeout value is `.synced(.none)` and no synced timeout currently exists.

## Depends On
https://github.com/wireapp/wire-ios-data-model/pull/509


zenkins:

johnxnguyen:
The single failing test is a flaky hotfix test, merging anyway

PR #790, opened 2018-06-29
Add german translations

daehn:
## What's in this PR?

* Add german translations.


mikeger:
@zenkins release patch

PR #793, opened 2018-07-05
Nil profile picture identifiers on 4xx responses

daehn:
## What's in this PR?

* We were not resetting the local `previewProfileAssetIdentifier` or `completeProfileAssetIdentifier` of a user when receiving a non-success response.


daehn:
@zenkins release patch

PR #795, opened 2018-07-09
Ensure no subsequent requests are fired when user image response is 4xx

daehn:
## What's new in this PR?

* We now set the `smallProfileRemoteIdentifier` / `mediumRemoteIdentifier` to nil when receiving a 4xx response when fetching old-style profile image assets.


daehn:
@zenkins release patch

PR #797, opened 2018-07-12
Simplified push token registration and enabled verifying with backend

vytis:
### Issues

Users have complained about missing push notifications, which we suspect is because the push token is not registered with backend correctly. This aims to simplify the code surrounding this and adds a way to check what is the currently registered token on backend and update it if necessary.

### Changes

1. Created a new class `PushToken` that supersedes `ZMPushToken` and is moved to data model (https://github.com/wireapp/wire-ios-data-model/pull/516)
2. Storing `PushToken` as a property on `UserClient` instead of on `NSManagedObject`
3. Updated `PushTokenStrategy.swift` to listed to updates on `UserClient`
4. Added functionality to get current push token from backend and reset the local one if we find it doesn't match
5. ~~Migrating current token to a new place in a Hotfix. This also triggers the validation of current token.~~ **EDIT**: after discussing with backend it was decided to skip the migration part because re-registering all push tokens is not a problem. This allows us to remove more code and avoid testing the migration scenario
6. Cleaned up tests, there were multiple places where we were checking if push token is registered correctly.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/516
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/88

## Notes

This should be merged when `GET /push/tokens` is available in production backend.


Source/Synchronization/Strategies/PushTokenStrategy.swift
```diff
@@ -26,161 +27,189 @@ private let zmLog = ZMSLog(tag: "Push")
...
+                // We should remove the local token
+                client.pushToken = nil
+
+                // Make sure UI tries to get re-register a new one
+                NotificationInContext(name: ZMUserSession.resetPushTokenNotificationName,
+                                      context: managedObjectContext.notificationContext,
+                                      object: nil,
+                                      userInfo: nil).post()
+
+                return false
```
mikeger: Should there be any case when this method is returning `true`?

Source/Synchronization/Strategies/PushTokenStrategy.swift
```diff
@@ -26,161 +27,189 @@ private let zmLog = ZMSLog(tag: "Push")
...
     public func processEvents(_ events: [ZMUpdateEvent], liveEvents: Bool, prefetchResult: ZMFetchRequestBatchResult?) {
         guard liveEvents else { return }
-        
-        events.forEach{process(updateEvent:$0)}
+
+        events.forEach{ process(updateEvent:$0) }
     }
-    
+
     func process(updateEvent event: ZMUpdateEvent) {
```
mikeger: In which circumstances we are receiving this event?

Source/Synchronization/Strategies/PushTokenStrategy.swift
```diff
@@ -26,161 +27,189 @@ private let zmLog = ZMSLog(tag: "Push")
...
+            if current.count == 1 && // We found one token
+                current[0].token == pushToken.deviceTokenString // It matches what we have locally
+            {
+                // Clear the flags and we are done
+                client.pushToken = pushToken.resetFlags()
+                return false
+            } else {
+                // There is something wrong, local token doesn't match the remotely registered
+
+                // We should remove the local token
```
mikeger: I think it's more likely that the local token is correct, since we can only receive it from iOS?

Source/UserSession/ZMUserSession+Push.swift
```diff
@@ -41,6 +41,55 @@ extension Dictionary {
...
+        }
+    }
+
+    @objc func resetPushToken() {
+        managedObjectContext.performGroupedBlock {
+            self.sessionManager.updatePushToken(for: self)
+        }
+    }
+
+    public func validatePushToken() {
```
mikeger: This would be called from the UI code?

WireSyncEngine.xcodeproj/project.pbxproj
```diff
@@ -3036,6 +3032,7 @@
 			baseConfigurationReference = 0994E1DE1B835C4900A51721 /* ios-test-target.xcconfig */;
 			buildSettings = {
 				BUNDLE_LOADER = "$(TEST_HOST)";
+				GCC_TREAT_WARNINGS_AS_ERRORS = NO;
```
mikeger: Not sure if we should set the settings in the project file if we have the xcconfigs setup.

Source/Synchronization/Strategies/PushTokenStrategy.swift
vytis: This happens when backend tries to send a push to this token and it fails validation on Apple side.

Source/Synchronization/Strategies/PushTokenStrategy.swift
vytis: Yes, but to play it safe we clear the cached version and immediately fire notification to get the push token from the system and sync it to the server.

WireSyncEngine.xcodeproj/project.pbxproj
vytis: No, it's a leftover from testing, I removed it from main target, but not from test target

Source/UserSession/ZMUserSession+Push.swift
vytis: Yes, I will make a PR soon with the changes needed on UI. Also added a comment since it's the only public method.

vytis:
@zenkins release major

PR #798, opened 2018-07-13
Removed tracking of the handle picking

mikeger:
As the part of the tracking cleanup.


mikeger:
@zenkins release major

PR #799, opened 2018-07-13
Align video calling terms with other platforms

infotexture:


mikeger:
@zenkins release patch

PR #800, opened 2018-07-16
Fix: crash in ZMConversationTranscoder - processDestructionTimerUpdate

billypchan:
### Issues

App crashes when a ZMConversation object is nil when setting its messageDestructionTimeout property.

### Causes

Missing nil checking of the ZMConversation object.

### Solutions

Change the signature of the the method processDestructionTimerUpdate, mark the conversation argument as optional. 

## Notes

- [x] Check why the upstream caller conversationFromEventPayload returns a nil conversation object.


billypchan:
@zenkins release patch

PR #801, opened 2018-07-16
fix: testIncomingVideoCall

billypchan:
## What's new in this PR?

update testIncomingVideoCall after localized string is updated.


billypchan:
@zenkins release patch

PR #802, opened 2018-07-17
ZIOS-10297: Initial Shared Identity Session Infrastructure

alexaubry:
### Issues

To begin implementation of session creation with shared identity (SSO), we need to have:

- An object that detects the login code from the pasteboard
- An object that validates the login code and returns the session cookie data

### Solutions

**Code detection**

We introduce the `SharedIdentitySessionRequestDetector`. You create this object with an object that conforms to the `Pasteboard` protocol (adopted by `UIPasteboard` and mock, for tests).

You call the `detectCopiedRequestCode` method to get the UUID stored in the pasteboard (must be in the format: `wire-<uuid>`). We do the processing in the background to keep the app responsive, as the string might be in the universal clipboard, and it might take time to retrieve it.

The completion handler is called on the main queue, with the extracted UUID, or `nil` if the contents of the pasteboard could not be validated.

You can also call `detectRequestCode(in string:)` to validate user input directly.

**Session Authenticator**

To get the session cookie from the code, we will be using an object that conforms to the `SharedIdentitySessionRequester` protocol. When specs are more details on the BE side, we will implement a concrete requester that makes the actual network request.

This protocol has a method named `requestIdentity(for token: String, _ completion:)`, that takes the login code, processes it, and returns a `SharedIdentitySessionResponse`.

The response can either be:

- success, with the cookie data
- pending additional information (e.g. user needs to put name), with the cookie data
- failure, with the error


Source/Registration/SharedIdentity/SharedIdentitySessionRequestDetector.swift
```diff
@@ -0,0 +1,89 @@
...
+    @objc public func detectCopiedRequestCode(_ completionHandler: @escaping (UUID?) -> Void) {
+        func complete(_ code: UUID?) {
+            DispatchQueue.main.async {
+                completionHandler(code)
+            }
+        }
+
+        processQueue.async {
+            if #available(iOS 10, *) {
+                guard self.pasteboard.hasString else {
```
daehn: It looks like `hasString` on the `Pasteboard` protocol is available on all iOS versions and the availability check here might not be needed.

Source/Registration/SharedIdentity/SharedIdentitySessionRequester.swift
```diff
@@ -0,0 +1,45 @@
...
+/// The result of a shared identity session request.
+public enum SharedIdentitySessionResponse {
+
+    /// A user session was created.
+    case success(Data)
+
+    /// A session was created, but the user still needs to provide information.
+    case pendingAdditionalInformation(Data)
+
+    /// A failure occured when
```
daehn: I think the `when` could be removed from the comment.

Source/Registration/SharedIdentity/SharedIdentitySessionRequestDetector.swift
alexaubry: Good call. I removed the `hasString` variable from `Pasteboard`, and moved the check to the `UIPasteboard` conformance directly, as checking for `hasString` and for `string != nil` was a bit redundant here.

alexaubry:
@zenkins release minor

PR #803, opened 2018-07-17
Revert to string-based clipboard detection to include original string

alexaubry:
### Issues

On UI, we use the SSO request detector to pre-fill to code alert. Thus we need to have the full string, including the`wire-prefix` returned by the detector, and not the UUID.

### Solutions

`detectCopiedRequestCode` not provides a String and not a UUID


alexaubry:
@zenkins release minor

PR #804, opened 2018-07-17
Find the SSO code when it's nested inside a larger string

alexaubry:
### Issues

We identified that some users could mis-copy the code from the e-mail (ex: if iOS also copies a newline after the code), which would prevent automatic code detection from working.

### Solutions

In the method that validates the code text, we:

- look for the `wire-` prefix anywhere in the string
- load the 36 next characters after this (UUID = 32 bytes + 4 dashes)
    - if there are not enough characters, we return `nil`
- validate the loaded characters as a UUID
- return the valid code part of the clipboard when detecting for auto-fill

This is an alternative to using regex, which would be much slower.


alexaubry:
@zenkins release patch

PR #805, opened 2018-07-18
Include isNew field in shared identity detector result

daehn:
## What's new in this PR?

* Include isNew field in shared identity detector result.
* Remove Objective-C compatibility (the `DetectorResult` is Swift only and we only use this class from Swift anyways).

### DIFF
--- a/Tests/Source/Registration/SharedIdentitySessionRequestDetectorTests.swift
+++ b/Tests/Source/Registration/SharedIdentitySessionRequestDetectorTests.swift
@@ -45,7 +45,7 @@ class SharedIdentitySessionRequestDetectorTests: XCTestCase {
         let detectionExpectation = expectation(description: "Detector returns a result")
 
         detector.detectCopiedRequestCode {
-            detectedCode = $0
+            detectedCode = $0?.code
             detectionExpectation.fulfill()
         }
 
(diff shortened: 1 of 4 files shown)


Tests/Source/Registration/SharedIdentitySessionRequestDetectorTests.swift
```diff
@@ -165,5 +165,98 @@ class SharedIdentitySessionRequestDetectorTests: XCTestCase {
...
+            
+            detector.detectCopiedRequestCode {
+                XCTAssertEqual($0?.isNew, true)
+                XCTAssertEqual($0?.code, code)
+                detectionExpectation.fulfill()
+            }
+            
+            waitForExpectations(timeout: 1, handler: nil)
+        }
+
```
alexaubry: Why are you wrapping this in `do` blocks? Didn't know you could do this 😄

mikeger: Would be cool to have an alternative, `dont_do`

daehn: @alexaubry just so I can use the same variable names multiple times, and structure the code a bit.

alexaubry: Very nice!

daehn:
@zenkins release major

PR #807, opened 2018-07-19
Implement standard company login requestor

alexaubry:
## What's new in this PR?

In this pull request, we implement the `CompanyLoginRequester`. Given the current build type and backend environment, it will generate the URL that will be opened by the browser. We notify when the URL is ready to be opened with the `CompanyLoginRequesterDelegate`.

We also add the URL actions that will handle the call-backs from the company login flow.

*Note* : The format of the URL and of the callback is subject to change.


Source/UserSession/ZMBackendEnvironment+setup.swift
```diff
@@ -50,6 +60,15 @@ enum BuildType {
         }
         
     }
+
+    var urlScheme: String {
```
daehn: We have to ensure these schemes are actually registered, we might have to adjust the xcconfig files for this. Also we should ensure using different schemes doesn't break anything with the connect bot scheme that is already in place. Maybe @mikeger can help us out here and knows more?

Source/Registration/Company/CompanyLoginRequester.swift
```diff
@@ -0,0 +1,76 @@
...
+        guard let buildType = BuildType(bundleID: bundleIdentifier) else {
+            fatalError("This unofficial build of Wire doesn't support company login.")
+        }
+
+        var urlComponents = URLComponents()
+        urlComponents.scheme = "https"
+        urlComponents.host = environment.type.backendHost
+        urlComponents.path = "/sso/initiate-login/\(token.uuidString)"
+
+        let successCallback = "\(buildType.urlScheme)://login/success/$cookie"
```
daehn: Maybe better to use url query parameters instead of path parameters for this, e.g. `/success?cookie=$cookie` and `/failure?label=$label`.

mikeger: I think it's better to use the `URLComponents` to construct the URL from string.

Source/UserSession/ZMBackendEnvironment+setup.swift
daehn: screen shot 2018-07-19 at 15 02 24

mikeger: Ideally to decide the URL scheme on the compile time. We have same sync engine that we use in all sorts of builds, so compile-time for SE would not work. I think one of the possible ways is to define the URL scheme in UI project's xcconfig files and then nicely apply it to Wire-Info.plist and pass the correct value to the sync engine.

alexaubry:
@mikeger @daehn I updated the PR. Thanks for the feedback!

Source/Registration/Company/CompanyLoginRequester.swift
```diff
@@ -0,0 +1,112 @@
...
+public protocol CompanyLoginRequesterDelegate: class {
+
+    /**
+     * The login requester asks the user to verify their identity on the given website.
+     *
+     * - parameter requester: The requester asking for validation.
+     * - parameter url: The URL where the user should be taken to perform validation.
+     */
+
+    func companyLoginSRequester(_ requester: CompanyLoginRequester, didRequestIdentityValidationAtURL url: URL)
```
daehn: Typo `companyLoginSRequester` → `companyLoginRequester`

alexaubry:
@daehn Some updates: We remove `RawURLAction`. Since the URL handling is done in Swift, we don't need this intermediate representation anymore. Passing the actual `URLAction` also allows us to present the alert for SSO errors in UI directly.

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -96,18 +90,7 @@ extension URLAction {
             return nil
         }
     }
-    
-    var rawAction: RawURLAction {
```
daehn: I guess this was for Objective-C interoperability, not needed anymore?

alexaubry: That's what I thought too. Already removed it

PR #809, opened 2018-07-20
Add logic to verify auth using url scheme using verification token

daehn:
## What's in this PR?

* Add logic to verify auth using url scheme using verification token.

### Still missing:
- [x] The verification call still has to be inserted. When receiving a company url auth callback we need to call `CompanyLoginVerificationToken.current.matches(identifier: tokenReceivedInTheUrlScheme)`.
- [x] We need to call `CompanyLoginVerificationToken.flushIfNeeded()` when opening the app.
- [x] We need to call `CompanyLoginVerificationToken.flush()` after using a token to verify a login operation.


vytis:
Maybe some more tests would be good for parts missing according to https://codecov.io/gh/wireapp/wire-ios-sync-engine/pull/809/diff

daehn:
@vytis merged in develop, seems like the token code is tested quite well and there were some tests of the url scheme handler in dev I didn't have locally.

daehn:
@zenkins release minor

PR #811, opened 2018-07-23
ZIOS-10243: Fix "Declined call" state

nicorsm:
### Issues

Calls rejected by the recipient should not be counted towards the amount of missed calls inside the conversations list.

### Solutions

While posting new notifications on the `WireCallCenterCallStateNotification`, I'm now passing the `previousCallState` parameter, so we can know if the call was previously rejected by the recipient and currently closed by the sender. I've also added the support for the `isRelevant` flag.

### Needs release with

- [x] https://github.com/wireapp/wire-ios-data-model/pull/523


Source/Calling/SystemMessageCallObserver.swift
```diff
@@ -60,7 +60,14 @@ final class CallSystemMessageGenerator: NSObject {
                 systemMessage =  conversation.appendPerformedCallMessage(with: 0, caller: caller)
             } else if reason == .canceled || reason == .timeout || reason == .normal {
                 log.info("Appending missed call message: \(caller.displayName), \"\(conversation.displayName)\"")
-                systemMessage = conversation.appendMissedCallMessage(fromUser: caller, at: timestamp ?? Date())
+                
+                var isRelevant = true
+                if case .incoming(video: _, shouldRing: false, degraded: _)? = previousCallState {
```
typfel: `shouldRing: false` seems wrong here, when `shouldRing` is false  the receiver doesn't get the option to decline the call.

typfel: Ok, I now get the intension behind `shouldRing: false` is that that's intention that the call has been declined by the user but that's not always true. A call could also be  `shouldRing: false` if  the app was re-started after the call begun. Maybe that's not relevant.

nicorsm:
@zenkins release major

PR #812, opened 2018-07-24
Do not fail silently when SSO success URL callback is not valid

alexaubry:
### Issues

Before this PR, we were dismissing the URL action if the callback was not valid. This means that when the users were redirected to the app from a seemingly valid URL, we would not show an alert to tell them something went wrong. 

This is a problem for troubleshooting issues, as we would have no context.

### Solutions

When decoding the `URLAction`, we return a company login error when the company login success could not be processed.

The error codes were borrowed from actual error codes from Apple to make sure we avoid conflicts, and are as follows:

| Wire | Apple Equivalent | Raw Value |
|------|------------------|-----------|
| SessionManagerURLHandlerError.missingRequiredParameter | CarbonCore.missingRequiredParameterErr | -2063 |
| SessionManagerURLHandlerError.invalidCookie | Security. errSecInvalidAccessCredentials | -67700 |

### DIFF
--- a/Source/SessionManager/SessionManagerURLHandler.swift
+++ b/Source/SessionManager/SessionManagerURLHandler.swift
@@ -70,16 +70,19 @@ extension URLAction {
             switch pathComponents[1] {
             case "success":
                 guard let cookieString = components.query(for: "cookie") else {
-                    return nil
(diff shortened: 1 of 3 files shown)


mikeger:
For the reference, what is CarbonCore https://en.wikipedia.org/wiki/Carbon_(API)

alexaubry:
@zenkins release minor

PR #814, opened 2018-07-25
Fix push token registration

vytis:
### Issues

The push token was not registered during fresh login and it was not possible to recover by resetting the push token from Settings.

### Causes

We were getting the push token before user client was registered and since we could not save it anywhere it was discarded. Trying to reset token from UI would also fail because it expected to have a token on user client.

### Solutions

When client is registered during login we register push token again which will trigger the request to register it with backend. When validating push token we also try to register it if it's not there yet.


vytis:
@zenkins release minor

PR #819, opened 2018-07-27
Add first implementation of company token validation logic

daehn:
## What's new in this PR?

* Add first implementation of company token validation logic.


daehn:
Just realized we won't have a `ZMUserSession` at this point (to use as `RequestEnqueueProvider`), will adjust the code accordingly.

daehn:
@alexaubry @mikeger updated!

PR #820, opened 2018-07-27
Use new call config endpoint

typfel:
## What's new in this PR?

Use new call config endpoint which has better IPv6 support.


mikeger:
@zenkins release patch

PR #822, opened 2018-07-27
Forward status code when validating company login token

daehn:
## What's in this PR?

* Forward status code when validating company login token.


daehn:
@zenkins release minor

PR #825, opened 2018-07-30
Actually perform a HEAD request when validating a SSO token

daehn:
## What's in this PR?

* Actually perform a HEAD request when validating a SSO token.

### DIFF
--- a/Source/Registration/Company/CompanyLoginRequester.swift
+++ b/Source/Registration/Company/CompanyLoginRequester.swift
@@ -47,8 +47,8 @@ extension URLQueryItem {
 }
 
 @objc public protocol URLSessionProtocol: class {
-    @objc(dataTaskWithURL:completionHandler:)
-    func dataTask(with url: URL, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask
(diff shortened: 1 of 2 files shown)


Source/Registration/Company/CompanyLoginRequester.swift
```diff
@@ -47,8 +47,8 @@ extension URLQueryItem {
 }
 
 @objc public protocol URLSessionProtocol: class {
-    @objc(dataTaskWithURL:completionHandler:)
-    func dataTask(with url: URL, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask
+    @objc(dataTaskWithRequest:completionHandler:)
+    func dataTask(with request: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Swift.Void) -> URLSessionDataTask
```
alexaubry: Why do we need `Swift.Void`?

daehn: We don't, just copied and pasted from the `URLSession.swift` header.

daehn: @alexaubry updated!

daehn:
@zenkins release patch

PR #827, opened 2018-08-01
Revert back to call config v1

mikeger:
## What's new in this PR?

Back to the future.


mikeger:
@zenkins release patch

PR #829, opened 2018-08-01
ZIOS-10369: Add information about SSO user in login error

alexaubry:
### Issues

We allowed SSO users with an e-mail and password to start the SSO login flow again from the login with e-mail page. This resulted in a loop (SSO flow -> e-mail sign in screen -> SSO flow ...)

### Solutions

We add information about whether the current self user was created from SSO in the error that is passed to the UI to display the sign-in screen.

This will be followed by a UI PR that disables the "For Companies" login button when it detects the  `ZMUserUsesCompanyLoginCredentialKey` key in the error.


alexaubry:
@zenkins release minor

PR #830, opened 2018-08-08
Updated services endpoint

vytis:
### Issues

The way we fetch services changed slightly by moving from `/services` to `/teams/<team_id>/services/whitelisted`.


Source/UserSession/Search/SearchTask.swift
```diff
@@ -318,11 +321,14 @@ extension SearchTask {
         }
     }
     
-    static func servicesSearchRequest(query: String) -> ZMTransportRequest {
+    static func servicesSearchRequest(teamIdentifier: UUID, query: String) -> ZMTransportRequest {
         let url = NSURLComponents()
-        url.path = "/services"
-        
-        url.queryItems = [URLQueryItem(name: "tags", value: "integration"), URLQueryItem(name: "start", value: query.trimmingCharacters(in: .whitespacesAndNewlines))]
+        url.path = "/teams/\(teamIdentifier.transportString())/services/whitelisted"
```
alexaubry: Could we use the Swift `URLComponents` here? Or is there a specific feature we need from `NSURLComponents`?

Source/UserSession/Search/SearchTask.swift
```diff
@@ -318,11 +321,14 @@ extension SearchTask {
...
+
+        let trimmedQuery = query.trimmingCharacters(in: .whitespacesAndNewlines)
+        if !trimmedQuery.isEmpty {
+            url.queryItems = [URLQueryItem(name: "prefix", value: trimmedQuery)]
+        }
         let urlStr = url.string?.replacingOccurrences(of: "+", with: "%2B") ?? ""
         return ZMTransportRequest(getFromPath: urlStr)
```
alexaubry: Theoretically, `URLComponents` escapes the returned URL, so we don't need to do it ourselves for the `+` character and the whitespace and newlines.

Might be worth adding test cases for this, and check if it works without the manual escaping.

Source/UserSession/Search/SearchTask.swift
vytis: Yeah, not sure what was the original reason to use `NSURLComponents`, maybe Swiftified version was not yet there when the code was written

Source/UserSession/Search/SearchTask.swift
vytis: It seems to not percent-encode `+`. It is a valid character in the URL but means something else, when you want a literal `+` you need to do it yourself.

vytis:
It seems the test was flaky on CI but always fail on local machine. `git bisect` told me that it started happening after changing threading behaviour of MockTransport (https://github.com/wireapp/wire-ios-mocktransport/pull/90). The test in question was checking what happens when access token is revoked but was also counting the number of requests made during the test. After threading changes one more request would be sent off before realising that we are unauthenticated. It was not really a test meant for checking this behaviour so I just removed the offending line.

vytis:
@zenkins release minor

PR #833, opened 2018-08-16
Feature/notification center

johnxnguyen:
## What's new in this PR?

This PR drops iOS 9 support and switches to the `UserNotification` framework for handling remote and local push notifications.

### Notification Content
`ZMLocalNotification` is the existing abstraction for generating notification content (e.g. title/body text, user info), from which a `UILocalNotification` object was generated. Now, it is used to generate a `UNNotificationContent` object, which is in turn used to generate a `UNNotificationRequest` that can be scheduled to the OS. 

Most notable change here is the `NotificationUserInfo` data type (thanks @alexaubry), which wraps the typical primitive notification `userInfo` dictionary and contains several necessary items of data including the identifiers for the conversation, sender, message and self user.

### Scheduling, Managing, Receiving
The `UIApplication` class is **no longer responsible for scheduling and managing notifications**. As a result, all notification related methods of the `Application` protocol abstraction has been removed. 

**Scheduling** and **managing** notifications can easily be done through the shared `UNUserNotificationCenter` object from any location. 

**Receiving** (local) notifications and action responses is the responsibility of the `UNUserNotificationCenterDelegate`, which is implemented by both `SessionManager` and `ZMUserSession`. The session manager is the designated delegate (it is called by the OS), whereby it routes the method call to the appropriate user session. The user session then handles the notification/actions as usual.

## Tests
Since `UNNotification` and `UNNotificationResponse` objects have no public initialisers, it is impossible to invoke the delegate methods implemented by `SessionManager` and `ZMUserSession` directly. Therefore, the logic in these methods has been refactored out increase testability.

The `UserNotificationCenter` protocol abstracts the Notification Center object, which allows us to mock it for testing. One pain though is to remember to assign this mock to all objects (potentially `LocalNotificationDispatcher`, `ZMLocalNotificationSet`, `ZMUserSession`, `SessionManager`) that will interact with the Notification Center object, a task that is not always obvious. Perhaps there is a way to be able to set the mock once and have it used everywhere?

## Notes
There are still some leftover tasks which I will do in a follow up PR. These are:
- Remove all iOS 10 annotations and iOS 9 compatibility code (mostly related to CallKit).


johnxnguyen:
@alexaubry thanks for doing the ground work.

typfel:
@zenkins release major

typfel:
@zenkins release major

PR #834, opened 2018-08-16
Only open websocket in the background when a call is active

typfel:
### Issues

The web socket was misbehaving during calls preventing call events to be processed.

### Causes

The web socket will fail if the app is in the background and the user is not connected to the active call since sockets will eventually close in the background. VoIP sockets are not working since iOS 10.

### Solutions

Don't allow the web socket to be open unless the user is participating in the call.

### DIFF
--- a/Source/Calling/WireCallCenterV3.swift
+++ b/Source/Calling/WireCallCenterV3.swift
@@ -570,7 +570,7 @@ public struct CallEvent {
         callSnapshots.removeValue(forKey: conversationId)
     }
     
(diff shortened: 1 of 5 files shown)


mikeger:
@typfel may be I am missing something, but not sure how the change is affecting the push channel state?

typfel:
callInProgress is updating the OperationState to .backgroundCall which makes the push channel stay open.

typfel:
@zenkins release patch

PR #835, opened 2018-08-20
Support e-mail activation codes in registration

alexaubry:
### Issues

The current registration flow does not send a 6-digit activation code to the e-mail address provided by the user, as expected in the current spec. 

Rather, it sends an e-mail with a link that the user needs to click. We make requests in a loop (ca. every 30s) until the link is marked as clicked by the backend. In addition to being the old implementation, this could cause the app to be killed when the user went to the background to click the link, or the backend sending us a 420 code when we waited too long, which blocked the registration flow.

### Solutions

The e-mail code logic was already implemented by team registration related strategies. Leveraging this current infrastructure, and the changes from data model where we don't expose the unregistered user to Objective-C anymore, we add support to user registration inside the Swift `RegistrationStatus` class:

- We delete request strategies related to user-only registration
- We add a phase enum case in the registration status to handle user registration
- We handle user registration cases in previously team-only registration strategies
- We make the activation steps generic, to support both e-mail and password, instead of just e-mail
- We remove methods related to registration from the unauthenticated session, as we can pass the `RegistrationStatus` to its consumers directly (ex: the new authentication coordinator).

The Objective-C `AuthenticationStatus` only handles login from now on.

## Dependencies

Needs releases with:

- [x] [wire-ios-data-model #542](https://github.com/wireapp/wire-ios-data-model/pull/542) - Modernize unregistered user structure + input normalization
- [x] [wire-ios-request-strategy #69](https://github.com/wireapp/wire-ios-request-strategy/pull/69/) - Use normalization changes from data model


Source/Registration/UnregisteredUser+Payload.swift
```diff
@@ -0,0 +1,58 @@
...
+            payload["phone"] = number
+            payload["phone_code"] = verificationCode
+
+        case .email(let address, let password)?:
+            payload["email"] = address
+            payload["password"] = password
+            payload["email_code"] = verificationCode
+
+        default:
+            break
```
vytis: Maybe it's then better to force-unwrap the `credentials`? If this is reached we have a problem anyway...

Tests/Source/Integration/TeamCreationTests.swift
```diff
@@ -41,72 +53,171 @@ class TeamCreationTests : IntegrationTest {
         super.tearDown()
```
vytis: don't forget to `nil` out all the the things that are created in `setUp()`

alexaubry:
@zenkins release major

PR #836, opened 2018-08-23
Provide a fallback for localized caller name

johnxnguyen:
### Issues

Sometimes it could happen that a user receives an incoming call via CallKit, but the caller name is a cryptic identification string.

### Causes

This happens in situations when the conversation in which the call is being made is unknown to the client. There already exists localisation measures to deal with missing caller/conversation names, however this logic is never reached. It depends on knowing the `conversationType`, which unfortunately is indeterminate.

### Solutions

Provide a fallback string for CallKit in the case that no other suitable string was found. This fallback is "Someone calling in a conversation".


johnxnguyen:
@zenkins release patch

PR #837, opened 2018-08-27
Add logs for tracing the processing of events

typfel:
## What's new in this PR?

This PR adds logs for tracking the event processing. This should make it easier to debug issues were it was previously unclear why events didn't reach the event consumers.

## Notes

- Some files has been converted to Swift in order to use shared `Logging` struct.


typfel:
@zenkins release patch

PR #838, opened 2018-08-28
Tracking calling test coverage

vytis:
### Issues

Most of the calling code is in SyncEngine and we would like to track test coverage better. This PR adds a label to Codecov.io in the same way as with https://github.com/wireapp/wire-ios-data-model/pull/546

## Notes

It's a bit suboptimal to have to list files related to calling in `codecov.yml`, need to think a bit what could be a better approach for this. Maybe simply doing a regexp `*[c|C]alling*` would work just as well.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,14 @@
+coverage:
+  status:
+    project:
+      calling:
+        flags: calling
+
+flags:
+  calling:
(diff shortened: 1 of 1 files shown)


vytis:
Please wait before merging this one, I need to make sure all looks allright on codecov.io

PR #841, opened 2018-08-29
ZIOS-10445: Adjust VoIP networking session service type

alexaubry:
## What's new in this PR?

iOS 10 provides a new session service type for call signaling. We use it instead of the VoIP type when available, as VoIP service type should only be used for voice calls (which go don't go through this session).

### Dependencies

- [ ] https://github.com/wireapp/wire-ios/pull/2606


typfel:
I think we should not switch to background session for downloading notifications or sending calling messages. These are time sensitive tasks which doesn't seem fit use case for a background session. Also they rely on completion handlers which will not be called when executed on background session as I understood i.

alexaubry:
@typfel As discussed, we are only changing the session type of the VoIP session instead of removing it.

alexaubry:
@zenkins release patch

alexaubry:
@zenkins release patch

PR #843, opened 2018-08-29
Don't use CallKit for unknown conversations

johnxnguyen:
### Issues

Answering a CallKit call in an unknown conversation fails immediately.

### Causes

When the call is answered via the CallKit screen, the `join` method is called on the conversation's `VoiceChannel`. This voice channel object is optional and is computed, however it depends on the `conversationType`, which is undetermined for an unknown conversation. Consequently, the call fails.

### Solutions

We cannot simply fetch the conversation because downstream syncing has been disabled in the background. Instead, we can prevent the CallKit call from appearing, and instead alert the user of the incoming call via APNS. This approach forces the user to open the app, which also gives them the opportunity to see who is calling them (since the CallKit screen would have no information about the caller whatsoever).

### DIFF
--- a/Source/Calling/CallKitDelegate.swift
+++ b/Source/Calling/CallKitDelegate.swift
@@ -265,6 +265,10 @@ extension CallKitDelegate {
             return log("Cannot report incoming call: conversation is missing handle")
         }
         
+        guard !conversation.needsToBeUpdatedFromBackend else {
+            return log("Cannot report incoming call: conversation needs to be updated from backend")
+        }
+        
         let update = CXCallUpdate()
         update.supportsHolding = false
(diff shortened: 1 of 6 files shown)


johnxnguyen:
@mikeger I've added the test that CallKit ignores incoming calls for unfetched conversations, but I couldn't figure out how to test that the APNS notification is scheduled instead. 

 This would occur in `CallStateObserverTests` which is a subclass of `MessageTest` which mocks the user session. I know I'd need to make sure the mock user session returns `.callKit` (only for this test) as the call notification type, but the user session is mocked in set up, which is too early to set the notification type. Do you have any ideas?

typfel:
@zenkins release patch

PR #845, opened 2018-08-31
Fix: flow manager not created

alexaubry:
### Issues

After removing `ZMCallFlowRequestStrategy`, the AVS could not be initialized and calls could not be made or received.

### Causes

`ZMCallFlowRequestStrategy` used to set itself as the delegate of the `FlowManager` wrapper, but the previous implementation of `FlowManager` we only created the actual `AVSFlowManager` when a delegate was set.

Because no object set itself as the delegate of the flow manager anymore, the `AVSFlowManager` was never created and the AVS could not be initialized.

### Solutions

We always create the `AVSFlowManager` when initializing the `FlowManager` wrapper. We also remove the `FlowManagerDelegate`, as no one uses it.


typfel:
`FlowManagerDelegate` was still used in the tests, but I guess you saw it.

alexaubry:
@zenkins release patch

PR #846, opened 2018-08-31
Feature/unordered relations masked

mikeger:
### Issues

Adopted the code for the new unordered relation logic.


typfel:
@zenkins release major

PR #847, opened 2018-09-03
Add call rejected elsewhere end reason

alexaubry:
### Issues

The web client sends `WCALL_REASON_REJECTED` call termination reason messages when the user rejects the call on one client so that other clients stop ringing. 

As discussed with @typfel, while we are not going to handle this case yet, we don't want it to appear as an unknown error.


typfel:
I'd suggest searching for `.terminating` in sync-engine to make sure that this change doesn't have any unwanted side effects.

alexaubry:
@typfel I added two cases to handle rejected elsewhere as answered elsewhere.

alexaubry:
@zenkins release minor

PR #848, opened 2018-09-03
Support media stopped callback

typfel:
## What's new in this PR?

Add another call state `.mediaStopped` which is fired as soon as the audio and video devices has been shut down. UI will use this call state to tear down the call UI faster without having to wait for any remaining signaling.

This could also have been implemented as a `CallClosedReason` but that would have introduced multiple terminating call states for a call which didn't seem desirable.


typfel:
@zenkins release major

PR #849, opened 2018-09-10
Don't wait for the sync to finish before we perform notification actions

typfel:
### Issues

When tapping on notification to open a message inside the app, the app waits for the syncing to finish before showing the message. This waiting time can be a really long in bad network conditions.

### Solutions

Don't wait for the sync to finish. The reason we were waiting for the sync to finish was because in the old calling protocol we did need to fetch the call state from the BE before we could react on a call notification.


typfel:
@zenkins release patch

PR #850, opened 2018-09-11
ZIOS-10471: Reorganize calling code

alexaubry:
### Issues

Calling is coordinated by the `WireCallCenterV3` class. As a consequence, the file where this class is defined became quite large and monolithic. 

All the AVS callbacks were defined as global functions in that class (as they needed to be passed in C), which in addition to reduce readability, polluted to global scope of the entire sync engine.

The class also lacked documentation.

#### AVS Wrapper

The AVS callback functions are now defined as properties on the `AVSWrapper`, using `@convention(c)` closures, instead of the global functions.

To reduce the responsibilities of the call center, we also perform converting of C pointer to Swift values in the `AVSWrapper`. To achieve this, and make it more readable, we add a protocol, `AVSValue`. 

This protocol defines the `AVSType` as an associated type, which is the raw C type of the Swift value from AVS (for example: `UnsafePointer<Int8>` for `String`. We then use type inference to convert the value from C to Swift, and pass it to the call center, whose callback logic was moved to a dedicated extension that has local methods that only take Swift values.  

***Downsides***

The downside of this approach is that we need to define a typealias for each avs callback function, and when a new one is added.

The `AVSType` decoding extension on `AVSWrapper` can also seem quite large and repetitive.

### Call Center

To make the call center more readable, we isolated to helper classes (call state, participant state, video state, call event, etc) to different files. Then, we grouped similar functions into extensions, each having a single scope (AVS integration, user actions, state helpers, snapshots, ...).

We also added documentation to make it more understandable for newcomers.

Finally, we removed calling V2 error codes that were not used anymore.

## To-Do

- [x] Update tests with new API


Tests/Source/Calling/WireCallCenterV3Mock.swift
```diff
@@ -139,14 +139,14 @@ public class WireCallCenterV3Mock : WireCallCenterV3 {
...
         
-    public override var nonIdleCalls : [UUID : CallState ] {
-        return mockNonIdleCalls
-    }
-    
-    public override var activeCalls: [UUID : CallState] {
-        return mockActiveCalls
-    }
-    
+//    public override var nonIdleCalls : [UUID : CallState] {
```
typfel: commented out code

Tests/Source/Calling/WireCallCenterV3Mock.swift
```diff
@@ -160,14 +160,14 @@ public class WireCallCenterV3Mock : WireCallCenterV3 {
...
-    public override func callState(conversationId: UUID) -> CallState {
-        return mockCallState
-    }
-
-    var mockInitiator : ZMUser?
-    
-    override public func initiatorForCall(conversationId: UUID) -> UUID? {
-        return mockInitiator?.remoteIdentifier ?? super.initiatorForCall(conversationId: conversationId)
-    }
+//
```
typfel: commented out code

Tests/Source/Calling/WireCallCenterV3Tests.swift
```diff
@@ -548,678 +534,198 @@ class WireCallCenterV3Tests: MessagingTest {
...
-        callBackMemberHandler(conversationIdRef: oneOnOneConversationIDRef, userId: otherUserID, audioEstablished: true, context: context)
+        callBackMemberHandler(conversationId: oneOnOneConversationID, userId: otherUserID, audioEstablished: true)
         XCTAssert(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
-        
+
         // then
         let connectedState = sut.state(forUser: otherUserID, in: oneOnOneConversationID)
         XCTAssertEqual(connectedState, CallParticipantState.connected(videoState: .stopped))
     }
-//    func testThatOtherIncomingCallsAreRejectedWhenWeAnswerCall() {
```
typfel: commented out code

PR #855, opened 2018-09-17
Turning on parallel tests

vytis:
Trying out if it decreases test time


vytis:
Quite a few tests seem to be flaky, maybe will have to try some other time!

PR #861, opened 2018-09-27
Fix/ephemeral bundling

johnxnguyen:
### Issues

Notifications for ephemeral messages were being grouped with non ephemeral messages from the same conversation, making it possible to see who sent the message.

### Causes

Notifications are grouped by conversation.

### Solutions

If the message is ephemeral, don't set the `threadIdentifier` of the `UNNotificationContent` object. This will cause a new group of notifications to appear, containing only ephemeral messages (regardless of sender or conversation).


johnxnguyen:
@zenkins release patch

PR #862, opened 2018-09-28
Align drafts placeholder with input bar copy

infotexture:
See also https://github.com/wireapp/wire-ios/pull/2742.


mikeger:
@infotexture wanted to mention that in this particular case you are going to change the title of the push notification quick action button.

mikeger:
@zenkins release patch

PR #864, opened 2018-10-02
Feature/in app notifications

johnxnguyen:
## What's new in this PR?

This PR removes the existing code to present `ZMLocalNotification` objects as Chat Heads in the UI project and instead presents native `UNNotification` objects while the app is foregrounded.

An Object conforming to the new `ForegroundNotificationResponder` protocol is able to control whether a notification will be presented while the app is in the foreground. This intended to be used in the UI project to prevent notifications appearing when the message for that notification
 is already visible to the user.


johnxnguyen:
@zenkins release major

PR #865, opened 2018-10-02
Remove mute button from notifications quick actions

vytis:
### Issues

There will be more than one type of mute configuration and this action does not make sense anymore.


vytis:
@zenkins release major

PR #866, opened 2018-10-02
Request show mention message when opened via notification

typfel:
## What's new in this PR?

When opening a notification containing a self mention we request the UI to show the first unread mention in that conversation.

## Notes

`ZMRequestsToOpenViewsDelegate` was replaced with `ShowContentDelegate` and part of the logic was moved up to the SessionManager level.


Source/UserSession/ZMUserSession+Actions.swift
```diff
@@ -27,43 +27,50 @@ private let zmLog = ZMSLog(tag: "Push")
...
-        guard let strongDelegate = requestToOpenViewDelegate else { return }
+    
+    func showContent(for userInfo: NotificationUserInfo) {
+        if let conversation = userInfo.conversation(in: managedObjectContext) {
+            
+            if let message = userInfo.message(in: conversation, managedObjectContext: managedObjectContext) as? ZMClientMessage,
+                let textMessageData = message.textMessageData, textMessageData.isMentioningSelf {
+                showConversation(conversation, at: conversation.firstUnreadMessageMentioningSelf)
+            } else {
+                showConversation(conversation)
```
johnxnguyen: Wouldn't we want to scroll show the message in the conversation for non mention notifications too?

typfel: According to the spec we only scroll to the first unread message if the notification contains a self mention, all other notifications will scroll to the first unread message in the conversation.

typfel:
@zenkins release major

Source/UserSession/ZMUserSession+Actions.swift
johnxnguyen: But this won't work if the conversation is already visible.

typfel: Doesn't work how? Because of how we handle this in the UI?

johnxnguyen: Yes, I think in the UI the `ZClientViewController` will only load the conversation if it's not already loaded... makes sense of course. But it means calling this for the current conversation will not make it scroll to the first unread message, as would be the case when you tap on the in app notification for that conversation.

johnxnguyen: But it sounds like the solution lies in the UI rather than here

typfel: Yep that sounds likely but I wouldn't attempt to solve that as part of these PRs. The conversation selection/loading code is in need of some re-factoring.

PR #869, opened 2018-10-04
Fix/foreground notifications

johnxnguyen:
## What's new in this PR?

Just a small update to the `ForegroundNotificationResponder` protocol. Passing just the conversation ID was too limiting, so instead we pass the entire `NotificationUserInfo` object (it contains the message nonce, the recipient ID, etc...).


johnxnguyen:
@zenkins release major

PR #870, opened 2018-10-04
New mute logic support

mikeger:
## What's new in this PR?

- Updated for the new Data Model API for muting the conversations.
- Added migration step to re-fetch all conversations in order to adjust to the potential update happened while this client was not updated.
- Added upstream syncing of the muted flags state.
- Added logic for unarchiving the conversation if it set to only mentions and mention is received.
- Added logic to display the notification if the conversation set to only mentions and mention is received.


mikeger:
@zenkins release major

PR #871, opened 2018-10-04
Wait for call events to finish processing when awaken by push notification

typfel:
### Issues

The app would in some cases get suspended by the system while processing an incoming call event.

### Causes

When the app gets woken up by PushKit in the background the app is supposed to process the notification as fast as possible and then call the completion handler which tells the system it's now fine to put the app back to sleep.

We used to download all the events and forward them to to all events consumers and then call the completion handler. The works fine except for calls events, because they might need to wait for us contacting the turn servers before they can be processed which happens asynchronously. This would result in call events not being processed in a timely manner.

### Solutions

Delay calling the completion handler if we are waiting for call events to be processed.

### Notes

- `ZMOperatonLoop+Background.m` has been converted to Swift.


Tests/Source/Synchronization/ZMOperationLoopTests.m
```diff
@@ -84,7 +86,7 @@ - (void)tearDown;
 {
     WaitForAllGroupsToBeEmpty(0.5);
     self.pushChannelObserverToken = nil;
-    [self.pushNotificationStatus stopMocking];
+//    [self.pushNotificationStatus stopMocking];
```
mikeger: Needs to be uncommented?

typfel: Needs to be deleted because it's not a mock anymore.

typfel:
@zenkins merge patch

typfel:
@zenkins release patch

PR #872, opened 2018-10-08
Fix race condition while waiting for call events

typfel:
### Issues

The app would still get suspended while processing call events in some cases.

### Causes

If `finishedProcessingCallEvent` was called before `waitForCallEventProcessingToComplete` we would not wait for the timeout timer to finish.

### Solutions

Check if we have a timer running in `waitForCallEventProcessingToComplete`.


typfel:
@zenkins release patch

PR #873, opened 2018-10-08
Fix: Team one-to-one show in group section

billypchan:
## What's new in this PR?

This PR is a follow-up of https://github.com/wireapp/wire-ios/pull/2762.

### Issues

1to1 conversation shows up in group section and in the contacts sections

### Causes

1to1 conversations are included in SearchResult.conversations array.

### Solutions

Filter out all 1to1 conversations with TeamMembers.

### DIFF
--- a/Source/UserSession/Search/SearchTask.swift
+++ b/Source/UserSession/Search/SearchTask.swift
@@ -118,11 +118,11 @@ extension SearchTask {
     }
     
     func conversations(matchingQuery query: String) -> [ZMConversation] {
-        let nonTeamUserPredicate = NSPredicate(format: "%K.@count > 2",
+        let non1to1ConversationPredicate = NSPredicate(format: "%K.@count > 2",
             ZMConversationLastServerSyncedActiveParticipantsKey
         )
 
-        let predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [ZMConversation.predicate(forSearchQuery: query), nonTeamUserPredicate])
+        let predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [ZMConversation.predicate(forSearchQuery: query), non1to1ConversationPredicate])
 
         let fetchRequest = ZMConversation.sortedFetchRequest(with: predicate)
         fetchRequest?.sortDescriptors = [NSSortDescriptor(key: ZMNormalizedUserDefinedNameKey, ascending: true)]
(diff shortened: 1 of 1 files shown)


Source/UserSession/Search/SearchTask.swift
```diff
@@ -93,7 +93,21 @@ extension SearchTask {
             let connectedUsers = self.request.searchOptions.contains(.contacts) ? self.connectedUsers(matchingQuery: self.request.query) : []
             let teamMembers = self.request.searchOptions.contains(.teamMembers) ? self.teamMembers(matchingQuery: self.request.query, team: team) : []
             let conversations = self.request.searchOptions.contains(.conversations) ? self.conversations(matchingQuery: self.request.query) : []
```
typfel: This is the code executing the fetch predicate to fetch the matching conversations.

https://github.com/wireapp/wire-ios-data-model/blob/c740d231bdd5efdd9d994e343316004e80a3c50a/Source/Model/Conversation/ZMConversation.m#L889

You can see that it contains a sub predicate to only return conversation with the type `.group`. However since this fetch request is executed on the sqlite store it ignores our additional code for dealing with fake one-to-one conversation in teams.

I wouldn't put this filtering in this method but rather in

https://github.com/wireapp/wire-ios-sync-engine/blob/c63bab52aac7eec91f3520b755893338b9015750/Source/UserSession/Search/SearchTask.swift#L119

Source/UserSession/Search/SearchTask.swift
```diff
@@ -93,7 +93,21 @@ extension SearchTask {
...
-            let result = SearchResult(contacts: connectedUsers, teamMembers: teamMembers, addressBook: [], directory: [], conversations: conversations, services: [])
+
+            // Filter the 1 to 1 conversations which already exists in teamMembers
+            let teamUsers = teamMembers.map { $0.user }
+
+            let conversationsWithoutTeamMembers = conversations.filter{
+                if $0.conversationType != ZMConversationType.oneOnOne {
+                    return true
+                }
+                if let user = $0.activeParticipants.array.first as? ZMUser {
```
typfel: Which case are you solving with this additional if case? Isn't it enough to filter on the `conversationType`?

billypchan: Yes, it may be redundant since all one to one conversations should match the query.

typfel:
The rules about how to to handle team one-to-one can be found here:
https://github.com/wireapp/wire-ios-data-model/blob/14b7f4eb16e16b4c10cf0bad16c45a24fd2e6f99/Source/Model/Conversation/ZMConversation.m#L487

I'd suggest to evaluate if it's feasible to turn those rules into a predicate which can run in a fetch request. That  means the predicate can only rely on properties stored in the data model (no computed properties).

If it's possible to construct such a predicate I'd integrate directly in:
https://github.com/wireapp/wire-ios-data-model/blob/14b7f4eb16e16b4c10cf0bad16c45a24fd2e6f99/Source/Model/Conversation/ZMConversation.m#L889

If it's not possible to construct such a predicate I'd fall back to to doing a in memory filter evaluating that `conversationType == .group` in
https://github.com/wireapp/wire-ios-sync-engine/blob/c63bab52aac7eec91f3520b755893338b9015750/Source/UserSession/Search/SearchTask.swift#L119

billypchan:
Close this ticket and fix this issue in Data model:
https://github.com/wireapp/wire-ios-data-model/pull/572

PR #874, opened 2018-10-08
Allow mute action for notification for non-team users

mikeger:
### Issues

The mute action should be displayed for the non-team users only. Team users must see no action.


mikeger:
@zenkins release patch

PR #876, opened 2018-10-09
Fix completion handler not being executed on the SE moc

typfel:
### Issues

App would crash with 

```
*** Terminating app due to uncaught exception 'NSGenericException', reason: '*** Collection <__NSCFSet: 0x283d55dd0> was mutated while being enumerated.'
```

in `unreadConversationCountInContext:`

### Causes

This is indicating the we are executing a count request from the wrong thread. Looking at the stack trace we confirm that it indeed executing the count request on the sync managed object context from the main thread. The stack trace begins from a timer which was newly added to handle processing of call events.

### Solutions

Switch back to sync thread in the completion handler.


typfel:
@zenkins release patch

PR #877, opened 2018-10-10
Updated strings for 3.22

mikeger:
New hot fresh strings 🥖


mikeger:
@zenkins release patch

PR #878, opened 2018-10-11
We should not expect to always get a userId in the close call handler

typfel:
### Issues

Call UI wouldn't dismiss when an outgoing call times out.

### Causes

We were always expecting a userId in the `call closed` handler which we don't get when a call times out.

### Solutions

Make `userId` optional.


typfel:
@zenkins release patch

PR #879, opened 2018-10-12
Fixed tests, because OTR assets can expire with new changes

nicorsm:
## What's new in this PR?

Fixed tests due to new changes about assets sending on slow networks - see https://github.com/wireapp/wire-ios-data-model/pull/547 for more info


nicorsm:
FYI there are some tests failing about mentions and notifications that are not in the scope of this PR (see CircleCI)

PR #885, opened 2018-10-19
Replace build badge with Azure Pipelines build

kaylangan:
Since you no longer have a .circleci/config.yml, I updated the build badge to use the Azure Pipelines build. Not sure if you want to scope this to a particular branch, but I scoped it to `develop`.


zenkins:

PR #886, opened 2018-10-23
Move isAuthenticated to BackendEnvironmentProvider

mikeger:
### Issues

Moved the isAuthenticated check that depends on the backend URL to `BackendEnvironmentProvider `. Seems to be cleaner solution, but in the end just swapping 2 arguments.


mikeger:
@typfel that is the maximum that I've managed to upgrade.

mikeger:
@zenkins release major

PR #887, opened 2018-10-25
SSO login: do not require email and password for device management

mikeger:
### Issues

Some necessary changes in order to allow SSO to work:

- Do not require adding email when signing in first time. Before we needed to add email in order to register the first client.
- Change the way we delete the clients, try without credentials first, and ask for login/password in case of an error.

Accompanying UI PR is also coming.


mikeger:
@zenkins release major

PR #889, opened 2018-10-29
Being more strict when push notification processing is done

vytis:
### Issues

When processing push notifications we were calling the completion handler earlier than we really finish processing everything. Namely:
1. We enqueue a delayed save when processing events. This would trigger actual save later and by that time we could already have signalled iOS that we are done.
2. Application badge is updated after calling completion handler.
3. After receiving messages we need to send delivery receipts if needed. This was not connected to push processing.

We have quite some crashes because when iOS tries to suspend the app we still have the database file open and locked. 

### Solutions

1. Removed delayed save after push events are finished processing.
2. Updating app badge and then calling completion handler afterwards.
3. Added a completion handler that gets called when all delivery receipts are sent. Only then we signal back to iOS that we are done processing the push notification.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-request-strategy/pull/85


vytis:
Will have to revisit this later, not good enough to merge at this point

PR #891, opened 2018-11-01
Fix/scroll to message

johnxnguyen:
### Issues

Tapping on a notification does not scroll that message into view, if the conversation was already loaded.

### Causes

We were only scrolling to first unread mentions if the notification was for a mention. Otherwise we just show the conversation.

### Solutions

Also scroll to the message for non mention notifications.


typfel:
@johnxnguyen I guess there are related to changes on UI for this work?

johnxnguyen:
@typfel yes and they're already merged: https://github.com/wireapp/wire-ios/pull/2839

PR #893, opened 2018-11-06
Feature/direct replies mute logic

johnxnguyen:
## What's new in this PR?

This PR updates notification processing to use the new DM API:

- occurrences of the `MutedMessageType.nonMentions` has been replaced with `.regular`.
- the logic used to determine if a `ZMLocalNotification` should be created for a message now uses the DM defined `ZMConversationMessage.isSilenced` property ([see here](https://github.com/wireapp/wire-ios-data-model/blob/0d15757dbdfd83ce8036c2b3b927cfdf754b42cd/Source/Model/Conversation/ZMConversation%2BMute.swift#L99)). There is no need to reimplement the logic.

## Needs Release With
- [ ] https://github.com/wireapp/wire-ios-data-model/pull/591


johnxnguyen:
Regarding the failing test, it also fails on the the `feature/direct-replies` branch. @typfel not sure if this relates to your recent changes?

typfel:
@johnxnguyen yep, looks like something I broke. I'll take a look.

PR #895, opened 2018-11-08
Add call quality observing

vytis:
### Issues

AVS provides functionality to track connection quality to each participant in a call. We would like to add support for this in the sync engine.

### Solutions

The quality is reported as being normal, medium or poor depending on multiple factors about the call. It is tracked per participant, so to calculate the call quality we just take the worst quality. The network quality is reported each 5 seconds.


Source/Calling/AVSWrapper.swift
```diff
@@ -90,6 +90,8 @@ public class AVSWrapper: AVSWrapperType {
 
         wcall_set_data_chan_estab_handler(handle, dataChannelEstablishedHandler)
         wcall_set_group_changed_handler(handle, groupMemberHandler, observer)
+        let timerIntervalInSeconds: Int32 = 5
```
alexaubry: Maybe this can be moved out of the initializer?

Source/Calling/CallParticipantSnapshot.swift
```diff
@@ -21,6 +21,13 @@ import Foundation
 class CallParticipantsSnapshot {
     
     public private(set) var members : OrderedSetState<AVSCallMember>
+
+    // We take the worst quality of all the legs
+    public var networkQuality: NetworkQuality {
+        return members.map { $0.networkQuality }
```
alexaubry: You can use they keypath syntax here:

```suggestion
        return members.map(\.networkQuality)
```

Source/Calling/AVSWrapper.swift
vytis: Not sure if it's worth polluting the initializer, it's basically an arbitrary constant how often the callback will fire...

Source/Calling/CallParticipantSnapshot.swift
```diff
@@ -17,10 +17,18 @@
...
 
 class CallParticipantsSnapshot {
     
     public private(set) var members : OrderedSetState<AVSCallMember>
+
+    // We take the worst quality of all the legs
+    public var networkQuality: NetworkQuality {
+        return members.array.map(\.networkQuality)
+            .sorted() { $0.rawValue < $1.rawValue }
+            .last ?? .normal
```
mikeger: Have you discussed the logic here with AVS folks? As I understand the code, if we would have the problem connecting to any of the peers the quality for us would be displayed as bad?

vytis: Yes, that was the plan as far as I can remember.

vytis:
@zenkins release minor

PR #897, opened 2018-11-09
Enable parallel tests

vytis:
### Issues

Xcode 10 enables running tests in parallel, trying out if tests are not more flaky because of this


vytis:
The total time decreased by 2-3 minutes, however there are some problems with Fastlane:
* By default it uses `xcpretty` which cannot parse output from testing on multiple simulators and Azure always sees as 0 tests have been run.
* We could use [trainer](https://github.com/KrauseFx/trainer) to parse test data instead, but it's not pre-installed on Azure. This means we would need to install all Ruby gems each build and it takes around 1-2 minutes, so total time gains would be negligible. 

Closing the PR and will maybe need to revisit this at some point in the future.

PR #898, opened 2018-11-09
Fix: update notification string to "replied to you"

billypchan:
### Issues

Update display string for reply notification for the design update.


billypchan:
@zenkins release patch

PR #899, opened 2018-11-09
Removed unused fastlane config

vytis:


vytis:
@zenkins release patch

PR #901, opened 2018-11-14
Localization for 3.24

mikeger:
_This field was intentionally left blank._


mikeger:
@zenkins release patch

PR #906, opened 2018-11-27
Save immediately after processing events

vytis:
### Issues

We have received user complaints about missing messages. The notification alert would be visible in lock screen, but the message would be missing.

### Causes

We could not reproduce the issue, but looking at crash logs when this happens we see that it crashes because of locked database file after delayed save. There is no good reason to have a delayed save when processing events because we do work in reasonably sized batches. 

### Solutions

Switch to regular save.


vytis:
@johnxnguyen It is an optimization that has been done in the past, but for what reason exactly it's not clear. Maybe it's worth revisiting this and removing it completely.

PR #908, opened 2018-11-28
Updated mock environment to not use HTTPS endpoint

vytis:
### Issues

When looking at logs of failing tests on SyncEngine there were some errors from App Transport Security about invalid certificate. We currently use https://example.com as backend in tests and at the moment a request is made to fetch client blacklist in tests (https://github.com/wireapp/ios-architecture/issues/53). It seems that example.com certificate has expired today and request is failing.

### Causes

It's not clear why tests would crash in this situation and on local machine I could not reproduce this. It does seem however that it's caused by not being able to fetch blacklist. When using non-secure example.com we see tests fail instead of crash.

### Solutions

Use non-secure example.com. A key enabling insecure loads was added to info.plist. Use production endpoint for blacklist fetching. Try to refactor code to avoid doing real network request in the future.

### DIFF
--- a/Tests/Source/Utility/BackendEnvironment+mock.swift
+++ b/Tests/Source/Utility/BackendEnvironment+mock.swift
@@ -20,6 +20,6 @@ import Foundation
 
 public extension BackendEnvironment {
     @objc public static var mockEnvironment: BackendEnvironment {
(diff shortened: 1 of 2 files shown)


vytis:
@mikeger for now yes, we need to refactor tests a bit to stop this...

vytis:
@mikeger @typfel so now tests simply fail because blacklist cannot be reached. Added real backend blacklist URL and it seems to be working again, but retried the tests just in case.

PR #909, opened 2018-11-28
Read receipts user setting sync

mikeger:
### Issues

This adds UserPropertyRequestStrategy loading.

Depends on 

- [ ] https://github.com/wireapp/wire-ios-request-strategy/pull/88


Source/Synchronization/Transcoders/ZMSelfStrategy.m
```diff
@@ -83,13 +83,7 @@ - (instancetype)initWithManagedObjectContext:(NSManagedObjectContext *)moc
         self.clientStatus = clientRegistrationStatus;
         self.syncStatus = syncStatus;
         self.upstreamObjectSync = upstreamObjectSync;
-        if (self.upstreamObjectSync == nil) {
```
typfel: Maybe some test was relying on this?

mikeger: I checked, and the tests are inserting the mock object, may be it was necessary before.

Tests/Source/Utility/KeySetTests.swift
```diff
@@ -37,7 +37,7 @@ class KeySetTests: MessagingTest {
     func testThatItCanIterate() {
         // given
         let sut = KeySet(["foo", "bar"])
-        var result: [WireDataModel.KeyPath] = []
+        var result: [WireDataModel.StringKeyPath] = []
```
typfel: Hmm seems like this test case belongs in data model

typfel: It's already there so it there so this file can be deleted.

mikeger:
@zenkins release major

PR #912, opened 2018-12-03
Allow user properties slow sync

mikeger:
### Issues

Set user properties to be fetched on slow sync. 

NB: the properties would be fetched after the slow sync itself is ended.

## Dependencies

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/622

### DIFF
--- a/Source/UserSession/SyncStatus.swift
+++ b/Source/UserSession/SyncStatus.swift
@@ -124,6 +124,9 @@ extension Notification.Name {
     }
     
     public func forceSlowSync() {
+        // Refetch user settings.
+        ZMUser.selfUser(in: managedObjectContext).needsPropertiesUpdate = true
+        // Set the status.
         currentSyncPhase = SyncPhase.fetchingLastUpdateEventID.nextPhase!
         syncStateDelegate.didStartSync()
     }


Source/UserSession/SyncStatus.swift
```diff
@@ -124,6 +124,9 @@ extension Notification.Name {
     }
     
     public func forceSlowSync() {
+        // Refetch user settings.
+        ZMUser.selfUser(in: managedObjectContext).needsPropertiesUpdate = true
```
typfel: This won't trigger a fetch of the user properties on a fresh login, also It won't guarantee that the user properties has been fetched before we start processing the notification stream. The other slow sync strategies gets the `SyncStatus` injected which is not possible in this case since the properties sync lives in `wire-ios-request-strategy.`

mikeger: On the fresh login it would rely on the default true value. Do we really need to guarantee it before starting the notification stream processing?

mikeger: My idea was that it's just a setting and it would be fetched before you would actually start read anything. In the worst case it would not send some read receipts on the beginning.

mikeger:
@zenkins release minor

PR #913, opened 2018-12-04
Possible fix for syncing issue

vytis:
### Issues

After https://github.com/wireapp/wire-ios/pull/2970 we have noticed that app stays in syncing mode until the app is force-quit.

### Causes

From logs it seems that background tasks are not created when processing.

### Solutions

If some of the notifications processing expire, we should resume `BackgroundActivityFactory` after coming to foreground. Otherwise no background tasks will be created and we will be stuck.


vytis:
@zenkins release patch

PR #914, opened 2018-12-04
Fix parsing of receipt mode payload

typfel:
## What's new in this PR?

receipt mode payload changed from
```
{data: Int}
```

to 

```
{data: receipt_mode:  Int}
```


typfel:
@zenkins release patch

PR #915, opened 2018-12-04
Fix deadlock in tests

vytis:
### Issues

`BackgroundAPNSConfirmationStatusTests` were getting stuck because of threading issues in WireTransport

### Causes

Because we have a single queue inside `BackgroundActivityFactory` the state of tests would not be completely the same after running several tests in a row.

### Solutions

Make sure to expire any outstanding background tasks when tearing down tests.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-transport/pull/116


vytis:
@zenkins release minor

PR #917, opened 2018-12-05
Resuming background task creation after receiving push

vytis:
### Issues

After getting the background task expiration callback we stop creating new background tasks. If the app doesn't get killed, but just suspended, no background tasks could be created anymore.

### Causes

We need to call `BackgroundActivityFactory.resume()` to reset the internal state that we are no longer in "expiration" mode. We do it when app comes to foreground, but we need to it when we get woken up by a push as well.

### DIFF
--- a/Source/SessionManager/SessionManager+Push.swift
+++ b/Source/SessionManager/SessionManager+Push.swift
@@ -69,6 +69,8 @@ extension SessionManager: PKPushRegistryDelegate {
         guard type == .voIP else { return completion() }
         
         log.debug("Received push payload: \(payload.dictionaryPayload)")
+        // We were given some time to run, resume background task creation.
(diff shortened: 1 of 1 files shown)


typfel:
Seems like we more cases where this could happen, background fetch for example.

vytis:
@typfel good point! Any others? Maybe some of the UIApplicationDelegate methods would be better suited for this?

typfel:
Would work if `applicationDidEnterBackground` is called when the app is unsuspended.

typfel:
And about more cases, not sure exactly but worth checking how the app is launched when

- You do an action on a push notification (reply, like)
- Start call via Callkit from the call history

vytis:
@typfel so basically the only place I could find is when liking/replying a message from notification. Starting a call opens the app directly.

PR #919, opened 2018-12-06
Improve background task naming

vytis:
## What's new in this PR?

To be able to read logs more efficiently we need to have descriptive names for background tasks. It's good to have some identifier (message UUID, push payload ID etc.) in the name to distinguish between similar ones.


vytis:
@zenkins release minor

PR #920, opened 2018-12-06
More robust check against duplicate events

typfel:
### Issues

We still had issues with duplicate `receipt-mode` events if toggled back and forth.

### Causes

Filtering out events that doesn't change the value only works the duplicated events come in one by one.

### Solutions

Verify the events have newer timestamp than previously processed events.


typfel:
@zenkins release patch

PR #922, opened 2018-12-07
Background task would not be finished after receiving call messages

vytis:
### Issues

Looking at logs we noticed this situation:
1. User receives calling message. Background tasks get created and destroyed correctly
2. User receives another calling message. Background tasks get created but never finished
3. Push completion handler for calling message no.2 would never get called and the task would expire eventually.

### Causes

When we receive a calling message we start a 2s timer to make sure all the things inside AVS are initialized. For unknown reason the timer would fire once, but not the second time it was created.

### Solutions

NSTimer works by scheduling things in run loop and it's not obvious what happens when it is used on background queue. In addition all this happens while app is running in background. The solution was to replace NSTimer with `ZMTimer` which is a wrapper around GCD timer api.

## Notes

Added more logging around this part of code.

### DIFF
--- a/Source/Data Model/NSManagedObjectContext+LastNotificationID.swift
+++ b/Source/Data Model/NSManagedObjectContext+LastNotificationID.swift
@@ -34,7 +34,7 @@ extension NSManagedObjectContext : ZMLastNotificationIDStore {
                     previousValue.compare(withType1: value) != .orderedAscending {
                 return
             }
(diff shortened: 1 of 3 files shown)


vytis:
P.S. Thanks @typfel for sitting through all this together!

vytis:
@zenkins release minor

PR #923, opened 2018-12-10
Process each event completely before processing the next one

typfel:
### Issues

Read receipt events would not be processed after coming back online.

### Causes

When processing a batch of events we would first process all the `message-add` events and then continue with all the `conversation-update` events. This results in the read receipt status for a conversations being out of sync while processing the message events since the conversation events (read receipt events) haven't been applied yet.

### Solutions

Let all event consumers process an event before continuing to process the next event.

### DIFF
--- a/Source/Synchronization/ZMSyncStrategy+EventProcessing.swift
+++ b/Source/Synchronization/ZMSyncStrategy+EventProcessing.swift
@@ -41,9 +41,11 @@ extension ZMSyncStrategy: ZMUpdateEventConsumer {
             
             Logging.eventProcessing.info("Consuming: [\n\(decryptedUpdateEvents.map({ "\tevent: \(ZMUpdateEvent.eventTypeString(for: $0.type) ?? "Unknown")" }).joined(separator: "\n"))\n]")
             
-            for eventConsumer in self.eventConsumers {
-                autoreleasepool {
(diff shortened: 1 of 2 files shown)


Source/Synchronization/ZMSyncStrategy+EventProcessing.swift
```diff
@@ -41,9 +41,11 @@ extension ZMSyncStrategy: ZMUpdateEventConsumer {
             
             Logging.eventProcessing.info("Consuming: [\n\(decryptedUpdateEvents.map({ "\tevent: \(ZMUpdateEvent.eventTypeString(for: $0.type) ?? "Unknown")" }).joined(separator: "\n"))\n]")
             
-            for eventConsumer in self.eventConsumers {
-                autoreleasepool {
-                    eventConsumer.processEvents(decryptedUpdateEvents, liveEvents: true, prefetchResult: prefetchResult)
+            autoreleasepool {
```
vytis: Not sure if this autoreleasepool is still needed, but should probably be added inside the loop body. Outside of it it doesn't really do much

typfel: I think we can remove it then since it would wrap the processing of 1 single event.

typfel: Updated

typfel:
@zenkins release patch

PR #925, opened 2018-12-11
Remove implicit save when requesting profile picture

typfel:
## What's new in this PR?

Update tests after the removal of implicit save when requesting profile images.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/635


typfel:
@zenkins release patch

typfel:
@zenkins release patch

PR #926, opened 2018-12-12
Recover from being out of sync with BE

typfel:
### Issues

If you for some reason was out of sync with the BE regarding the read receipt setting in a conversation we would not recover when attempting to change it.

### Causes

If you try to change the read receipt setting to the already set value BE will respond with an empty 204 response,  which we would just silently ignore.

### Solutions

Update the local `hasReadReceiptsEnabled` in case of a 204 response.


typfel:
@zenkins release patch

PR #928, opened 2018-12-12
Add a hotfix for syncing read receipts setting after upgrading

typfel:
## What's new in this PR?

We need to slow sync group conversations and the self user read receipt setting after upgrading to a version which supports read receipts. This fixes the edge case were the read receipt setting is turned on in a conversation before the user has had the chance to upgrade Wire.


typfel:
@zenkins release patch

PR #929, opened 2018-12-12
Don't show group created push for team one-to-one conversations

typfel:
### Issues

When team member starts a one-to-one conversation the other side would see a "<Name> created a group conversation".

### Causes

Team one-to-one conversation are group conversation under the hood so they will generate `conversation.create` events.

### Solutions

Add an additional check to verify that the new conversation is actually a group conversation.


typfel:
@zenkins release patch

PR #930, opened 2018-12-13
Only refetch conversations you are still a member of

typfel:
### Issues

When we need to re-fetch group conversations after upgrading it can need quite a lot of requests.

### Solutions

Exclude groups which you are no longer a member of.


typfel:
@zenkins release patch

PR #931, opened 2018-12-13
Fix not finishing background task if read

typfel:
### Issues

Background task would not finish if read

### Causes

Not taking new .read state into account

## Notes

see https://github.com/wireapp/wire-ios-request-strategy/pull/98


typfel:
@zenkins release patch

PR #932, opened 2018-12-17
Don't use deprecated ZMOtrAssetMeta

typfel:
## What's new in this PR?

Don't rely on `ZMOtrAssetMeta` in the tests since it's no longer in use.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/638
- [x] https://github.com/wireapp/wire-ios-images/pull/22


typfel:
@zenkins release major

PR #933, opened 2018-12-17
Changes from transport

vytis:
### Issues

After changing the way certificate pinning works (https://github.com/wireapp/wire-ios-transport/pull/122) there were some small API changes.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/122
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/104


typfel:
@zenkins release major

PR #934, opened 2018-12-18
Use isSent instead of checking against every possible delivery state

typfel:
## What's new in this PR?

Use isSent instead of checking against every possible delivery state

See https://github.com/wireapp/wire-ios-data-model/pull/640

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/640


typfel:
@zenkins release minor

PR #935, opened 2018-12-27
Custom APS certificate name support

mikeger:
### Issues

Moved the APS environment logic from utilities to SE. Also simplified it and used the `BuildType` enum logic to handle the different certificate names.

### Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/643
- [x] https://github.com/wireapp/wire-ios-utilities/pull/70


vytis:
@zenkins release major

PR #936, opened 2019-01-08
ZIOS-11019: Added read receipts when replying and adding like to a notification

nicorsm:
### Issues

Read receipts were not sent after replying and adding a like to a notification.

### Solutions

I've added the logic to send them if needed.

### DIFF
--- a/Source/UserSession/ZMUserSession+Actions.swift
+++ b/Source/UserSession/ZMUserSession+Actions.swift
@@ -136,14 +136,14 @@ private let zmLog = ZMSLog(tag: "Push")
         
         enqueueChanges {
             guard let message = conversation.append(text: message) else { return /* failure */ }
-            self.setReadReceiptIfNeeded(with: userInfo, in: conversation)
+            self.appendReadReceiptIfNeeded(with: userInfo, in: conversation)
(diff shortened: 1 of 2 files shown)


typfel:
I think we should a test for this. Also do did you confirm with product/design that this is the behaviour we want (only confirming the message you responded to)?

Source/UserSession/ZMUserSession+Actions.swift
```diff
@@ -136,12 +136,21 @@ private let zmLog = ZMSLog(tag: "Push")
...
         enqueueChanges {
             guard let message = conversation.append(text: message) else { return /* failure */ }
+            self.setReadReceiptIfNeeded(with: userInfo, in: conversation)
             self.messageReplyObserver = ManagedObjectContextChangeObserver(context: self.managedObjectContext, callback: { [weak self] in
                 self?.updateBackgroundTask(with: message)
             })
         }
     }
     
+    private func setReadReceiptIfNeeded(with userInfo: NotificationUserInfo, in conversation: ZMConversation) {
```
typfel: Maybe `append` is more appropriate verb here, like `appendReadReceiptIfNeeded `.

nicorsm:
@zenkins release minor

PR #937, opened 2019-01-09
Allow new device names

mikeger:
### Issues

Support for new device names is added.


mikeger:
@zenkins release patch

PR #938, opened 2019-01-09
Improve CallKit logging

typfel:
## What's new in this PR?

Add more logging when requesting CallKit actions


mikeger:
@zenkins release patch

PR #940, opened 2019-01-10
Feature/share ext ipc

johnxnguyen:
## What's new in this PR?

This PR contains two minor changes:

1. Expose the `ZMUserSession` method `mergeChangesFromStoredSaveNotificationsIfNeeded()`. We will need to invoke this method from the UI in order to update changes from the Share Extension while the Wire Client is in the foreground.
2. When the above method is invoked, don't process pending changes on the UI context. This solves an issue where although the changes from Share Extension are merged on the UI, the UI still doesn't update to reflect these changes. The cause is subtle; if we process pending changes, then the context is no longer in a dirty state. Then when we save the context (a couple lines later), the save doesn't actually occur because it isn't dirty.

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-utilities/pull/71
- [x] https://github.com/wireapp/wire-ios-share-engine/pull/80


johnxnguyen:
Flaky test, releasing anyway.

johnxnguyen:
@zenkins release minor

PR #942, opened 2019-01-16
Fix notifications background processing

vytis:
### Issues

We have noticed that when a user receives multiple push notifications in one go (e.g. coming back online) the background tasks would not get finished correctly.

### Causes

PushKit expects us to call completion handler when we are done processing push payload. In our case we call this when notification stream has been downloaded and all events were processed. 

When there are multiple pushes coming in one go we get all the needed information by fetching notification stream only once. To make sure we don't miss any events we keep track of what we received from pushes and what was downloaded from notification stream inside `PushNotificationStatus`. The logic was not working correctly because we would save the last notification ID AFTER we do calculations of what events were fetched. This caused a bug that completion handler of only the first push payload would be called, but others would just linger on until background tasks start expiring. 

### Solutions

Instead of relying on `managedObjectContext.zm_lastNotificationID` we pass the last notification identifier after calculating it.

### Note

Also changed the name of push notification background task to make it easier to parse.

### DIFF
--- a/Source/SessionManager/SessionManager+Push.swift
+++ b/Source/SessionManager/SessionManager+Push.swift
@@ -85,7 +85,7 @@ extension SessionManager: PKPushRegistryDelegate {
         
         guard let accountId = payload.dictionaryPayload.accountId(),
               let account = self.accountManager.account(with: accountId),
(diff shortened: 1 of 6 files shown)


vytis:
@zenkins release major

PR #943, opened 2019-01-17
Tear down ephemeral timers when going to background

vytis:
### Issues

Ephemeral timers would still be kept active when app was put to background. This meant that app would stay running in the background trying to wait until they fire which might happen in the far future (e.g. ephemeral with 24h timeout). More detailed explanation is in underlying PR - https://github.com/wireapp/wire-ios-data-model/pull/648

### Causes

When coming to foreground we would restart timers for all ephemeral messages, but would not tear them down when going to background.

### Solutions

Tear down timers when app goes to background.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/648


vytis:
@zenkins release major

PR #944, opened 2019-01-18
Start timers when app is launched or comes to foreground

vytis:
### Issues

After changes introduced by https://github.com/wireapp/wire-ios-data-model/pull/648 we don't start ephemeral timers explicitly. It turned out that after #943 we would never start them. This lead to ephemeral messages not having a countdown at all.

### Solutions

When app is launched or comes to foreground we start the timers explicitly.


vytis:
@zenkins release patch

PR #945, opened 2019-01-21
Fix after transport changes

vytis:
### Issues

There were some API changes in https://github.com/wireapp/wire-ios-request-strategy/pull/104

### Solutions

Most of it because we have only single session, so no need to force to use other kinds of sessions. Some unused methods and their tests have been removed also.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/104
- [x] https://github.com/wireapp/wire-ios-transport/pull/127
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/105 

## Notes

Also removed some test classes that were on disk, but not added to the project since about 2 years.


vytis:
@zenkins release major

PR #946, opened 2019-01-21
ZIOS-11056: Add aggregate logging of events during notification processing

nicorsm:
## What's new in this PR?

Added logging during event processing for:
- [x] time elapsed
- [x] number of events processed
- [x] number of core data objects changes
- [ ] number of saves performed


Source/Synchronization/ZMSyncStrategy+EventProcessing.swift
```diff
@@ -44,12 +46,14 @@ extension ZMSyncStrategy: ZMUpdateEventConsumer {
             for event in decryptedUpdateEvents {
                 for eventConsumer in self.eventConsumers {
                     eventConsumer.processEvents([event], liveEvents: true, prefetchResult: prefetchResult)
+                    self.eventProcessingTracker?.registerEventProcessed()
```
vytis: So this will be a bit misleading, because `self.eventConsumers` are the strategies, so for each event it will loop over the strategies and you will actually get `decryptedUpdateEvents.count * eventConsumers.count` invocations of this method.

Source/Synchronization/ZMSyncStrategy+EventProcessing.swift
```diff
@@ -36,6 +36,8 @@ extension ZMSyncStrategy: ZMUpdateEventConsumer {
         eventDecoder.processEvents(updateEvents) { [weak self] (decryptedUpdateEvents) in
             guard let `self` = self else { return }
             
+            self.eventProcessingTracker?.registerStartedProcessing()
```
vytis: I think this is a bit overkill - we can simply do `let start = Date()` and then at the end do:
```
Logging.eventProcessing.debug("Processed in \(-start.timeIntervalSinceNow), \(self.eventProcessingTracker?.debugDescription ?? "")")
```

This would also simplify the API of the `EventProcessingTracker` because `save(attribute: Attributes, value: Double)` would not be needed.

Source/Synchronization/EventProcessingTracker.swift
```diff
@@ -0,0 +1,134 @@
...
+        increment(attribute: .dataDeletionPerformed)
+    }
+    
+    public func registerFinishedProcessing() {
+        let attribute = Attributes.processingDuration
+        let currentValue = (persistedAttributes(for: eventName)[attribute.identifier] as? Double) ?? 0
+        save(attribute: attribute, value: Date().timeIntervalSince1970 - currentValue)
+    }
+    
+    private func increment(attribute: Attributes, by amount: Double = 1) {
```
vytis: `amount` should probably be Integer, we don't really add anything that's not whole values

nicorsm:
@zenkins release minor

PR #950, opened 2019-01-24
Revert transport changes

vytis:
## What's new in this PR?

Reverting #945 because it caused other issues.


vytis:
@zenkins release major

PR #951, opened 2019-01-24
Fix: Add the missing localization for one case

billypchan:
## What's new in this PR?

Add the missing localization for ping one-time case.


billypchan:
This PR is not correct to add "time" for pinging once. It can be fixed in another way.

PR #953, opened 2019-01-28
Ignore team.conversation-create and team.conversation-delete events

typfel:
### Issues

We were creating invalid conversations.

### Causes

Upon receiving `team.conversation-create` we created empty conversations and attempted to sync them which failed with 404. The reason for this is that we only receive `team.conversation-create` when a team conversation is created which we are not a member of.

### Solutions

There's no reason to create conversations which we are not a member of so we should just ignore these events.


vytis:
Maybe it would be better to tweak tests to make sure we ignore these events and don't create conversations when receiving the event?

typfel:
👍  I'll add/restore one test for testing that we ignore the event.

typfel:
@vytis updated

typfel:
@zenkins release patch

PR #961, opened 2019-02-07
Process user.delete events

typfel:
### Issues

We were seeing that some users kept sending broadcast messages which the backend reported as redundant. This caused our client to think it was out of sync with the backend and therefore started a slow sync.

### Causes

The redundant users were users who had deleted their accounts. The backed does inform us about about deleted accounts through the `user.delete` update event and if we try to fetch a deleted user it will have the "deleted" flag set to `true`. Unfortunately our client ignores this information at moment which is the reason why it kept sending sending redundant messages.

### Solutions

* Flag users which have been deleted in the database when receiving the `user.delete` event or when the `deleted` flag is set when fetching a user.
* Remove deleted users from group conversations.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/672
- [x] https://github.com/wireapp/wire-ios-transport/pull/129
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/107


typfel:
@zenkins release major

PR #962, opened 2019-02-11
ZIOS-11207: Fix unresponsive SSO link

alexaubry:
### Issues

If the user launched the app from scratch after they signed in, tapping an SSO link has no effect.

### Causes

The handling of the link assumed that there was an unauthenticated session, which isn't always the case (for ex, if the user launches the app and never started the login flow).

### Solutions

- Force-create an unauthenticated session when the SSO link is tapped.
- Do not cache the SSO team lookup results (in case the team gets deactivated later).


alexaubry:
Manual build passed: https://dev.azure.com/wireswiss/Wire%20iOS/_build/results?buildId=3157&view=logs&jobId=5264e576-3c6f-51f6-f055-fab409685f20

alexaubry:
@zenkins release minor

PR #963, opened 2019-02-12
ZIOS-11190: Fix dismissing SSO flow if the user is already logged in

alexaubry:
### Issues

When launching the app from scratch with an SSO link, the flow would appear briefly and disappear after a few seconds if there is already an account logged in.

### Causes

When the app launches, we try to select the first available account. This causes a race condition with the URL handling:

- we launch the app and start asynchronously loading the resources on disk
- we handle the URL synchronously and tell the app state controller to go unauthenticated and the SSO flow starts
- the asynchronous loading finishes and tells the app state controller to go authenticated
- the SSO flow is dismissed and the conversation list is displayed

### Solutions

Do not select the initial account if the URL action at launch causes a logout action.


alexaubry:
@zenkins release patch

PR #964, opened 2019-02-14
Fix unwanted unread conversations

typfel:
### Issues

`.newConversation` system messages can now generate unread dots, this created unwanted unread conversation for people how upgrade.

### Causes

We previously tried to solve this by introducing `hasBeenModifiedSinceSlowSync` but that solution didn't cover all the edge cases, a conversation can be modified by for example someone being added or removed causing the same issue appear re-appear.

### Solutions

- Mark conversations created from a slow sync as read after inserting the .newConversation system message
- Add a hot fix which marks all .newConversation system messages as read

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/680
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/108


typfel:
@zenkins release major

PR #966, opened 2019-02-18
Bump components

vytis:
After https://github.com/wireapp/wire-ios-request-strategy/pull/115 and https://github.com/wireapp/wire-ios-mocktransport/pull/109


vytis:
@zenkins release major

PR #968, opened 2019-02-20
Try to overlap background activities

vytis:
### Issues

When app is running in the background it should have an active background task or it can be suspended by iOS at any point. This sometimes happen when firing `ZMRequestAvailableNotification ` - the previous one has already been completed and new one has not been started yet. This leads to app running in the background for longer than needed.

### Causes

The main problem is that when request completes we fire `ZMRequestAvailableNotification`. However this is done on a private queue when all other completion handlers have been called already. This means that all background tasks of running requests have been finished by now and there could potentially be a gap before we start a new one.

### Solutions

Start another background activity just after the request was generated and finish it after we fire the last `ZMRequestAvailableNotification`.

## Notes

Another approach was tried to use the same `BackgroundActivity` that was created in `executeNextOperation` and pass it to the `requestGenerator`. This does not work in practice, because `requestGenerator` can potentially be called multiple times and would then try to finish the same activity multiple times.

### DIFF
--- a/Source/Synchronization/ZMOperationLoop.m
+++ b/Source/Synchronization/ZMOperationLoop.m
@@ -243,6 +243,7 @@ static char* const ZMLogTag ZM_UNUSED = "OperationLoop";
             return nil;
         }
         ZMTransportRequest *request = [self.syncStrategy nextRequest];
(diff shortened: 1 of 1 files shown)


vytis:
@zenkins release patch

PR #969, opened 2019-02-21
Make sure there is an active background task

vytis:
### Issues

Looking at real-live logs after #968 we have noticed that sometimes the background task introduced there is not ended which causes the app to stay in background longer.

### Causes

There is this curious extra notification that we try to do when dispatch group on sync MOC is empty. It seems in some cases this never gets run. This needs further debugging, but since the dispatch group is only useful in tests it's quite low priority.

### Solutions

Remove extra notification and code to wait until group is empty. There is also no need for the extra background task because we should have either delayed save or next request happening at that point already.


vytis:
@zenkins release patch

PR #972, opened 2019-02-27
ZIOS-11250: Add observation mechanism for new unauthenticated sessions

alexaubry:
### Issues

Sometimes, the session manager recreates user sessions. However, the authentication coordinator in the UI sets up observers on the session to detect events. Thus, we need a mechanism to detect new unauthenticated sessions so that we can recreate these observers.


alexaubry:
@zenkins release major

PR #975, opened 2019-03-01
Hotfix path: refetch the self user properties after update

alexaubry:
### Issues

When upgrading from 3.26 to 3.27, some users did not have the `managed_by` flag set in their self profile. It caused an issue with identifying their level of permissions to edit their profile picture and name.

### Solutions

Add a hotfix patch to refetch the self user properties when updating the app.

### DIFF
--- a/Source/Synchronization/ZMHotFixDirectory.m
+++ b/Source/Synchronization/ZMHotFixDirectory.m
@@ -161,6 +161,13 @@ static NSString* ZMLogTag ZM_UNUSED = @"HotFix";
                      patchCode:^(NSManagedObjectContext *context) {
                          [ZMHotFixDirectory markAllNewConversationSystemMessagesAsRead:context];
                      }],
+
+                    /// We need to refetch the managedBy flag of the user after the backend release.
+                    [ZMHotFixPatch
+                     patchWithVersion:@"235.0.1"
+                     patchCode:^(NSManagedObjectContext *context) {
+                         [ZMHotFixDirectory refetchUserProperties:context];
+                     }],
                     ];
     });
     return patches;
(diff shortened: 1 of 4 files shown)


Tests/Source/Synchronization/ZMHotFixTests.swift
```diff
@@ -98,4 +98,34 @@ class ZMHotFixTests_Integration: MessagingTest {
...
+            // WHEN
+            let sut = ZMHotFix(syncMOC: self.syncMOC)
+            self.performIgnoringZMLogError {
+                sut!.applyPatches(forCurrentVersion: "235.0.0")
+            }
+        }
+        XCTAssertTrue(self.waitForAllGroupsToBeEmpty(withTimeout: 0.5))
+
+        syncMOC.performGroupedBlock {
+            self.syncMOC.saveOrRollback()
```
typfel: Is the test not passing without this extra save?

Source/Synchronization/ZMHotFixDirectory.m
```diff
@@ -161,6 +161,13 @@ - (NSArray *)patches
                      patchCode:^(NSManagedObjectContext *context) {
                          [ZMHotFixDirectory markAllNewConversationSystemMessagesAsRead:context];
                      }],
+
+                    /// We need to refetch the managedBy flag of the user after the backend release.
+                    [ZMHotFixPatch
+                     patchWithVersion:@"235.0.1"
+                     patchCode:^(NSManagedObjectContext *context) {
+                         [ZMHotFixDirectory refetchUserProperties:context];
```
typfel: Shouldn't we call `refetchSelfUser:` here?

Tests/Source/Synchronization/ZMHotFixTests.swift
alexaubry: It didn't pass because I was calling the wrong method 😬

alexaubry:
@zenkins release patch

PR #978, opened 2019-03-06
Add hotfix for re-downloading team members

typfel:
## What's new in this PR?

We need to re-download the team members after the `createdBy` and `createdAt` fields were introduced.


typfel:
@zenkins release patch

PR #979, opened 2019-03-07
Long ringtones

typfel:
## What's new in this PR?

- Use long version of sounds for ringtones
- Add method for updating Callkit configuration, this is necessary in order to change the ringtone without restarting the app.


typfel:
@zenkins release minor

PR #980, opened 2019-03-12
Don't rely on order of .established & .establishedDataChannel

typfel:
## What's new in this PR?

Support for AVS using vanilla WebRTC. This has changed the semantics of `.establishedDataChannel` and  `.established`:

Previously `.establishedDataChannel` always happened before `.established`, that's no longer true so we now also need to report Callkit calls as connected on `.established`.


typfel:
@zenkins release patch

PR #981, opened 2019-03-13
Remove V2 profile images

typfel:
## What's new in this PR?

- Remove `UserImageStrategy` which used to download V2 profile images
- Remove code for re-uploading V2 profile images as V3

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/695


Source/Synchronization/Strategies/SearchUserImageStrategy.swift
```diff
@@ -120,11 +120,7 @@ public class SearchUserImageStrategy : AbstractRequestStrategy {
         switch asset {
```
vytis: Since `SearchUserAssetKeys` contains only one case, maybe we could just make it a struct?

Tests/Source/Synchronization/Strategies/SearchUserImageStrategyTests.swift
```diff
@@ -425,8 +327,8 @@ extension SearchUserImageStrategyTests {
         XCTAssertEqual(request2.method, .methodGET);
         XCTAssertTrue(request2.needsAuthentication);
         
-        let expectedPath1 = requestPath(for:assetID1, of:searchUser1.remoteIdentifier!)
-        let expectedPath2 = requestPath(for:assetID2, of:searchUser2.remoteIdentifier!)
+        let expectedPath1 = "/assets/v3/\(assetID1)" //requestPath(for:assetID1, of:searchUser1.remoteIdentifier!)
```
vytis: Is that commented out on purpose?

typfel: No  😔

typfel:
@zenkins release major

PR #982, opened 2019-03-14
Update AVS

typfel:
## What's new in this PR?

Update AVS because they added some new public API


typfel:
@zenkins release patch

PR #983, opened 2019-03-19
Reduce unnecessary saving during the initial sync

typfel:
### Issues

The initial sync on a large account takes a long time to finish.

### Causes

The profiling report for a large account shows that a significant amount of time is spent in core data's save method, which is triggered when inserting a conversation.

### Solutions

Remove a delayed save after inserting a new conversation. This save should not be necessary since we enqueue a save after the request response has been processed.


typfel:
@zenkins release patch

PR #984, opened 2019-03-19
Feature: Go to a specific user profile when clicking on a link outside/inside the app

billypchan:
## What's new in this PR?

Supporting the URL scheme type for opening a user profile or a conversation.

Modify `sessionManagerShouldExecuteURLAction` of protocol `SessionManagerURLHandlerDelegate` to return a flag that the `URLAction` is executed successfully or not.

### DIFF
--- a/Source/DeepLink/DeepLinkError.swift
+++ b/Source/DeepLink/DeepLinkError.swift
@@ -1,0 +1,28 @@
+//
+// Wire
+// Copyright (C) 2019 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/**
+ * Errors that can occur when requesting a user profile or conversation from a link.
+ */
+
+public enum DeepLinkRequestError: Error, Equatable {
+    /// The Deep link provided by the user was invalid.
+    case invalidLink
+}
--- a/Source/DeepLink/DeepLinkError.swift
+++ b/Source/DeepLink/DeepLinkError.swift
@@ -23,7 +23,7 @@ import Foundation
  */
 
 public enum DeepLinkRequestError: Error, Equatable {
-    /// The Deep link provided by the user was invalid.
-    case invalidLink
+    case invalidUserLink
+    case invalidConversationLink
     case notLoggedIn
 }
--- a/Source/DeepLink/DeepLinkError.swift
+++ b/Source/DeepLink/DeepLinkError.swift
@@ -25,4 +25,5 @@ import Foundation
 public enum DeepLinkRequestError: Error, Equatable {
     /// The Deep link provided by the user was invalid.
     case invalidLink
+    case notLoggedIn
 }
--- a/Source/SessionManager/SessionManagerURLHandler.swift
+++ b/Source/SessionManager/SessionManagerURLHandler.swift
@@ -181,6 +181,13 @@ extension URLAction {
 }
 
 public protocol SessionManagerURLHandlerDelegate: class {
+
+    /// sessionManager executes a URLAction
+    ///
+    /// - Parameters:
+    ///   - action: the action to execute
+    ///   - callback: the callback with a bool shouldExecute, it should be called after the action is executed.
+    /// - Returns: return false if the Action is not executed
     func sessionManagerShouldExecuteURLAction(_ action: URLAction, callback: @escaping (Bool) -> Void) -> Bool
 }
 
--- a/Source/SessionManager/SessionManagerURLHandler.swift
+++ b/Source/SessionManager/SessionManagerURLHandler.swift
@@ -18,6 +18,20 @@
 
 import Foundation
 
+public struct User: Equatable {
+    public static func == (lhs: User, rhs: User) -> Bool {
+        return lhs.id == rhs.id
+    }
+
+    private let id: UUID
+    var user: UserType?
+
+    init(id: UUID) {
+        self.id = id
+        user = nil
+    }
+}
+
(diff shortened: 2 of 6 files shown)


Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -178,15 +206,19 @@ public final class SessionManagerURLHandler: NSObject {
             return false
         }
 
-        if action.requiresAuthentication {
+        if action.opensDeepLink {
```
alexaubry: Why not return true for `requiresAuthentication` in this case? Do we need a separate variable?

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -178,15 +206,19 @@ public final class SessionManagerURLHandler: NSObject {
...
+            guard let _ = userSessionSource?.activeUserSession else {
+                pendingAction = action
+                return true
+            }
```
alexaubry: It seems like we don't do anything here?

Source/SessionManager/SessionManagerURLHandler.swift
billypchan: For `openDeepLink` case, no need to call `handle(action: action, in: userSession/unauthenticatedSession)` so I created a workflow for it.

Source/SessionManager/SessionManagerURLHandler.swift
billypchan: The only thing have to do here is storing `pendingAction` if there is no `activeUserSession`.

Source/DeepLink/DeepLinkError.swift
```diff
@@ -0,0 +1,28 @@
...
+
+import Foundation
+
+/**
+ * Errors that can occur when requesting a user profile or conversation from a link.
+ */
+
+public enum DeepLinkRequestError: Error, Equatable {
+    /// The Deep link provided by the user was invalid.
+    case invalidLink
```
vytis: I guess this should have some more cases - otherwise we can just return true/false depending if the link is valid or not.

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -25,7 +25,11 @@ public enum URLAction: Equatable {
 
     case startCompanyLogin(code: UUID)
     case warnInvalidCompanyLogin(error: ConmpanyLoginRequestError)
-    
+
+    case openConversation(id: UUID)
```
typfel: We don't want the UI to work with uuid and the `NSManagedObjecContext`. I would replace this with a `ZMConversation` instance.

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -25,7 +25,11 @@ public enum URLAction: Equatable {
...
+    case openUserProfile(id: UUID)
```
typfel: ... and this with a `UserType/ZMUser.

Source/DeepLink/DeepLinkError.swift
billypchan: Currently, we just have one kind or error case but we may have more error case later.

vytis: Will there be more errors added later in the implementation or maybe some time in the future? If the latter then we can just have `case warnInvalidDeepLink` for now and add the error type when we have at least two errors. If more are coming shortly then just ignore this comment :)

Source/SessionManager/SessionManagerURLHandler.swift
billypchan: @typfel there is a problem related to the user session. Since when we create this URLAction, the UserSession would be nil and we can not obtain the ZMUser/ZMConversation object at this point.

typfel: Can't we create the `URLAction` later, like just before we call the `handle(action:)`. Might be that we need another representation before the user session is loaded.

billypchan: The `URLAction` first created when app delegate opens a URL (before user session and UI are created). I think we can change the associated value's type to a custom struct and assign the ZMUser/ZMConversation object later.

Source/DeepLink/DeepLinkError.swift
billypchan: I added one more case `notLoggedIn` from the updated spec.

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -18,14 +18,51 @@
 
 import Foundation
 
+public struct DeepLinkUser: Equatable {
```
billypchan: @typfel I created this struct `DeepLinkUser` to prevent showing the UUID to the UI. It takes the UUID and stores it privately when `init`. Once the active user session is obtained, we get the `user` object and the UI can access it.

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -25,7 +25,32 @@ public enum URLAction: Equatable {
 
     case startCompanyLogin(code: UUID)
     case warnInvalidCompanyLogin(error: ConmpanyLoginRequestError)
-    
+
+    case openConversation(id: UUID, conversation: ZMConversation?)
+    case openUserProfile(id: UUID, user: ZMUser?)
```
billypchan: @typfel I changed the associated values with optional `ZMConversation/ZMUser`

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -198,28 +246,46 @@ public final class SessionManagerURLHandler: NSObject {
...
+        if let result = delegate?.sessionManagerShouldExecuteURLAction(mutableAction, callback: callback) {
+            return result
+        } else {
+            return false
+        }
     }
 
     fileprivate func handle(action: URLAction, in unauthenticatedSession: UnauthenticatedSession) {
-        delegate?.sessionManagerShouldExecuteURLAction(action) { shouldExecute in
+        let _ = delegate?.sessionManagerShouldExecuteURLAction(action) { shouldExecute in
```
typfel: `sessionManagerShouldExecuteURLAction ` returns a `Bool` now, but it's discarded here. Do you plan to do something with it or can it be removed?

billypchan: I will change `sessionManagerShouldExecuteURLAction` as `@discardable`. For `unauthenticatedSession` there is no change in this PR.

typfel: Ah I see 👍

PR #986, opened 2019-03-21
Don't allow notification fetching from APNS during syncing

typfel:
### Issues

It's been observed that the notification stream is fetched unnecessarily during the initial sync. Since this fetch doesn't include the `since` query parameter it will contain "old" events which are a wasted effort to consume.

### Causes

When logging in into an account the user creates a new client, which will cause the backend to send a push notice prompting us to update the `PushNotificationStatus`. This in turn causes the the `ZMMissingUpdateEventsTranscoder` to create a request to fetch notification stream, which will not include the `since` parameter because it is only be persisted after the sync is finished. Doing a `/notifications` request without the `since` parameter will return all events the backend still has stored.

### Solutions

Don't allow fetching the APNS notifications while we are syncing. 

NOTE: the fact that we were previously allowing this forced many strategies to include `.allowsRequestsDuringNotificationStreamFetch` in their configuration otherwise they would get stuck when the scenario described above happens.


typfel:
@zenkins release patch

PR #987, opened 2019-03-21
Notification fetching follow up

typfel:
## What's new in this PR?

Follow up on https://github.com/wireapp/wire-ios-sync-engine/pull/986

### Issues

We were still performing `/notifications` requests during sync

### Causes

`isFetchingStreamForAPNS` was not using the `notificationFetchStatus` logic.

### Solutions

Query the application status for the `notificationFetchStatus`


typfel:
@zenkins release patch

PR #988, opened 2019-03-22
Add show user profile method to ShowContentDelegate

billypchan:
## What's new in this PR?

Add a new `func showUserProfile(user: UserType)` to `ShowContentDelegate`.
~~For `ShowContentDelegate.showConversation` method, change the argument `session` to optional to allow the showConversation in current session.~~
Remove the return of `sessionManagerShouldExecuteURLAction`

### DIFF
--- a/Source/SessionManager/SessionManager+Push.swift
+++ b/Source/SessionManager/SessionManager+Push.swift
@@ -191,17 +191,22 @@ extension SessionManager: PKPushRegistryDelegate {
 
 @objc
 public protocol ShowContentDelegate: class {
-    
     func showConversation(_ conversation: ZMConversation, at message: ZMConversationMessage?)
     func showConversationList()
-    
+    func showUserProfile(user: UserType)
 }
 
 extension SessionManager {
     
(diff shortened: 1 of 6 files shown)


Source/SessionManager/SessionManager+Push.swift
```diff
@@ -191,17 +191,22 @@ extension SessionManager: PKPushRegistryDelegate {
...
-            self.showContentDelegate?.showConversation(conversation, at: message)
+    public func showConversation(_ conversation: ZMConversation,
+                                 at message: ZMConversationMessage? = nil,
+                                 in session: ZMUserSession? = nil) {
+        if let session = session {
+            activateAccount(for: session) {
+                self.showContentDelegate?.showConversation(conversation, at: message)
+            }
+        } else {
+            showContentDelegate?.showConversation(conversation, at: message)
```
typfel: Why is this change necessary?

typfel: `activateAccount()` won't do anything if it's already the current session.

PR #990, opened 2019-03-22
Fix: When the user/convsersation ID is not found in a user session, change URLAction to warning

billypchan:
## What's new in this PR?

Change the URLAction to .warnInvalidDeepLink when the user/convsersation ID is not found in a user session.


billypchan:
@zenkins release patch

PR #991, opened 2019-03-22
Bump components for persistent link attachments

alexaubry:
### Issues

Due to compatibility issues between clients, we need to treat YouTube and SoundCloud links as a separate type of links, since they cannot be sent as link previews.

Their parsing was previously done in the UI and not persisted; which means we needed to recompute them for every rendering of the text message cells.

In addition to this performance bottleneck, we used the YouTube and SoundCloud API to manually fetch the details of the media, which this could have been achieved using the OpenGraph data provided by these websites at faster speeds (both for displaying the preview and general scrolling) and lower costs.

### Solutions

In SE, we needed to fix an issue with a class whose initializer wasn't marked as Obj-C anymore.


alexaubry:
@zenkins release major

PR #992, opened 2019-03-25
Produce notLoggedIn error in SessionManagerURLHandler when no authenticated account exists

billypchan:
## What's new in this PR?

~~Create a new property `hasUnhandledDeepLink` for SessionManagerURLHandler. ~~

When opening an URL, if there is no authenticated account exists, change the `URLAction` to `.warnInvalidDeepLink(error: .notLoggedIn` and do not set `pendingAction`

Created a new property `isSelectedAccountAuthenticated` for `UserSessionSource` to figure out is the selected account authenticated.


typfel:
Couldn't we return `notLoggedIn` error directly if the account is not logged in?

billypchan:
@typfel I removed `.notLoggedIn` since the app state of logged in/out is handled by `AppStateController` in UI project.

Tests/Source/SessionManager/SessionManagerURLHandlerTests.swift
```diff
@@ -24,6 +24,9 @@ import XCTest
 final class UserSessionSourceDummy: UserSessionSource {
     var activeUserSession: ZMUserSession? = nil
     var activeUnauthenticatedSession: UnauthenticatedSession
+    var isSelectedAccountAuthenticated: Bool {
```
alexaubry: Would be nice if this could be a mutable variable, if we wanted to have tests with this set to false in the future.

billypchan: @alexaubry changed as suggestion, thanks!

billypchan:
@zenkins release patch

PR #994, opened 2019-03-26
Lookup user by UUID

typfel:
## What's new in this PR?

Add `lookup(userId: UUID) -> SearchTask` to the `SearchDirectory`. This method fetches a user by UUID and returns the results in the form of a `SearchUser`.


typfel:
@zenkins release major

PR #995, opened 2019-03-27
ZIOS-11310: No error message when opening a deep link with broken part

nicorsm:
### Issues

App wasn't displaying any error when opening a broken deep link, like `wire://userx/xxx` or `wire://connect/something`.

### Solutions

I've added support to a new deep link error case, `.malformedLink`.


alexaubry:
The tests seem to be failing still.

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -96,7 +96,8 @@ extension URLAction {
     init?(url: URL, validatingIn defaults: UserDefaults = .shared()) {
         guard let components = URLComponents(string: url.absoluteString),
             let host = components.host,
-            components.scheme == "wire" else {
+            let scheme = components.scheme,
+            scheme.starts(with: "wire") == true else {
```
alexaubry: Why do we need this? Our app will never be opened by a URL scheme that is not registered in the Info.plist.

nicorsm: There's a test that is trying to call this initializer using a link from google.com. With the last default case that returns an error, the test fails. But yes, both control and test are not used in real life.

nicorsm:
@zenkins release patch

PR #996, opened 2019-03-27
Fix/zios 11316 cannot open a non connected user profile with deep link

billypchan:
## What's new in this PR?

Create a new enum value `.connectToUser(id: UUID, searchTask: SearchTask)` when the user opens a deep link with a user id of an unconnected user.

### DIFF
--- a/Source/UserSession/Search/SearchTask.swift
+++ b/Source/UserSession/Search/SearchTask.swift
@@ -19,7 +19,7 @@
 import Foundation
 import WireUtilities
 
-public class SearchTask {
+final public class SearchTask {
     
     public enum Task {
         case search(searchRequest: SearchRequest)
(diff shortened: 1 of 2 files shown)


Source/UserSession/Search/SearchTask.swift
```diff
@@ -94,6 +94,17 @@ public class SearchTask {
     }
 }
 
+extension SearchTask: Equatable {
```
alexaubry: This seems to be unrelated to the change. Do we need to make it equatable?

billypchan: Since I added 'case connectToUser(id: UUID, searchTask: SearchTask)' to `URLAction` and `URLActio` is `Equatable`

alexaubry: Ah, makes sense now!

billypchan:
close this PR, this can be done with a simpler approach.

PR #998, opened 2019-03-28
ZIOS-7347: Do not play audio and video formats we don't understand

alexaubry:
### Issues

When rendering video or audio messages, we would offer the user to play video or audio files with the player, even if the format cannot be played by the system (such as MIDI or MPEG).

See details in https://github.com/wireapp/wire-ios-data-model/pull/704.
## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/129
- [x] https://github.com/wireapp/wire-ios-data-model/pull/704
- [x] https://github.com/wireapp/wire-ios-utilities/pull/75


alexaubry:
@zenkins release major

PR #1001, opened 2019-04-01
Disable object observation during slow sync

typfel:
### Issues

The app spent a lot of resources during the slow sync.

### Causes

We were calculating change notification for objects changed during the slow sync, which is unnecessary since we don't display any UI during the slow sync.

### Solutions

Distinguish between slow & quick during the sync phases and disable the notification dispatcher during the slow sync phases.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/707

## Notes

In order to keep the syncing the logic in one place, I've moved the notification dispatcher from the `ZMSyncStrategy` to `ZMUserSession`.


typfel:
@zenkins release major

PR #1002, opened 2019-04-02
Fix: Partner can see a team user profile which is not in conversation

billypchan:
### Issues

A partner user can see a team user profile which is not in the conversation.

### Causes

We should add the restriction for this case.

### Solutions

Do not allow open another user's profile when self-user is a partner and:
1. the user is not an inviter of self-user
2. the user does not exist in self-user's conversations.


typfel:
Maybe if we would always route the the profile opening through through the `lookup(userId:)` we wouldn't have to to deal with this special scenario here. The partner filtering would then we be contained to the search directory.

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -44,8 +44,35 @@ public enum URLAction: Equatable {
...
+                ///partner restriction - not allow to see other partner(w/o converation)'s profile, except the user invited(created) self user.
+
+                let showProfile: Bool
+                let selfUser = ZMUser.selfUser(in: moc)
+                if selfUser.teamRole == .partner {
+                    if selfUser.membership?.createdBy == user {
+                        showProfile = true
+                    } else {
+                        let activeConversations = selfUser.activeConversations
+                        let activeContacts = Set(activeConversations.flatMap({ $0.activeParticipants }))
```
alexaubry: We can probably flatten this by doing: 

```swift
showProfile = selfUser.activeConversations.contains {
    $0.activeParticipants.contains(user)
}
```

billypchan:
This story restarts in
https://github.com/wireapp/wire-ios-sync-engine/pull/1004

PR #1003, opened 2019-04-03
ZIOS-11332: Unsupported AVS version warning

nicorsm:
### Issues

Starting from AVS 4.9, `wcall_recv_msg` can return the error code 1000 `WCALL_ERROR_UNKNOWN_PROTOCOL`. That can happen if the receiver has an older version of AVS (and Wire) that doesn't support the protocol implemented in the version of the other side.

### Solutions

I've added a new `CallError` enum and a new `WireCallCenterCallErrorObserver` (+ notification), in a similar way as the call state observer. `CallError` can be nil when no errors are thrown.

## Notes

I'm adding some tests to this feature right now. It will require more testing when the new version of AVS will be out.


Source/Calling/CallReceivedResult.swift
```diff
@@ -0,0 +1,53 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
```
alexaubry: Small nitpick...

```suggestion
// Copyright (C) 2019 Wire Swiss GmbH
```

Source/Calling/WireCallCenterV3+Notifications.swift
```diff
@@ -115,6 +115,19 @@ public struct WireCallCenterMissedCallNotification : SelfPostingNotification {
     let video: Bool
 }
 
+// MARK:- Received call observer
+
+public protocol WireCallCenterCallErrorObserver : class {
+    func callCenterCallError(_ error: CallError)
```
alexaubry: I would suggest using the `<type>did<event>` syntax here, which is standard across Apple first-party API. We could use `callCenterDidReceiveCallError(_ error: CallError)`.

Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -150,9 +150,13 @@ extension CallingRequestStrategy : ZMEventConsumer {
                 
                 callEventStatus.scheduledCallEventForProcessing()
                 
-                callCenter?.processCallEvent(callEvent, completionHandler: { [weak self] in
+                callCenter?.processCallEvent(callEvent, completionHandler: { [weak self] (error) in
                     self?.zmLog.debug("processed calling message")
                     self?.callEventStatus.finishedProcessingCallEvent()
+                    
+                    if let context = self?.managedObjectContext, let error = error {
+                        WireCallCenterCallErrorNotification(context: context, error: error).post(in: context.notificationContext)
```
typfel: Can we post the notification inside the call center?

Source/Calling/AVSWrapper.swift
```diff
@@ -135,13 +135,16 @@ public class AVSWrapper: AVSWrapperType {
...
-    public func received(callEvent: CallEvent) {
+    public func received(callEvent: CallEvent) -> CallError? {
+        var result: CallError? = nil
         callEvent.data.withUnsafeBytes { (bytes: UnsafePointer<UInt8>) in
             let currentTime = UInt32(callEvent.currentTimestamp.timeIntervalSince1970)
             let serverTime = UInt32(callEvent.serverTimestamp.timeIntervalSince1970)
             zmLog.debug("wcall_recv_msg: currentTime = \(currentTime), serverTime = \(serverTime)")
-            wcall_recv_msg(handle, bytes, callEvent.data.count, currentTime, serverTime, callEvent.conversationId.transportString(), callEvent.userId.transportString(), callEvent.clientId)
+            result = CallError(wcall_error: wcall_recv_msg(handle, bytes, callEvent.data.count, currentTime, serverTime, callEvent.conversationId.transportString(), callEvent.userId.transportString(), callEvent.clientId))
+            result = CallError.unknownProtocol
```
alexaubry: Are you assigning result twice?

nicorsm: Oh yes, that was for testing purposes 😅

nicorsm:
@zenkins release minor

PR #1004, opened 2019-04-04
Searching user for opening a user profile deep link 

billypchan:
## What's new in this PR?

This PR removes `URLAction.openUserProfile`. All user profile deep links are handled by `URLAction.connectToUser` instead.
When searching the directory, team users are excluded. A new method `performLocalLookup` is created for searching local user by user id.


Source/UserSession/Search/SearchTask.swift
```diff
@@ -124,22 +157,28 @@ extension SearchTask {
...
+        let activeContacts = Set(activeConversations.flatMap({ $0.activeParticipants }))
+        let selfUser = ZMUser.selfUser(in: context)
+
+        let result = members.filter({
+            if let user = $0.user {
+                return selfUser.membership?.createdBy == user || activeContacts.contains(user)
+            } else {
+                return false
+            }
+        })
```
nicorsm: We could shrink the syntax a little bit here, something like:
```
return members.filter({
     guard let user = $0.user else { return false }
     return selfUser.membership?.createdBy == user || activeContacts.contains(user)
})
```

Source/SessionManager/SessionManagerURLHandler.swift
```diff
@@ -27,9 +27,8 @@ public enum URLAction: Equatable {
     case warnInvalidCompanyLogin(error: ConmpanyLoginRequestError)
 
     case openConversation(id: UUID, conversation: ZMConversation?)
-    case openUserProfile(id: UUID, user: ZMUser?)
 
-    // The user id from an unconnected user (The UI has to search for this user id)
+    // Search for the user ID and open the profile view for connection request if not connected
     case connectToUser(id: UUID)
```
typfel: Doesn't it make more sense keep the name `openUserProfile` and remove `connectToUser ` since we are not actually connecting to the user.

Source/UserSession/Search/SearchRequest.swift
```diff
@@ -45,6 +45,18 @@ public struct SearchOptions : OptionSet {
     
 }
 
+extension SearchOptions {
+    public mutating func updateForSelfUserTeamRole(selfUser: ZMUser) {
```
typfel: 👍

Source/UserSession/Search/SearchTask.swift
```diff
@@ -87,15 +87,48 @@ public class SearchTask {
...
+        context.performGroupedBlock {
+            let selfUser = ZMUser.selfUser(in: self.context)
+
+            var options = SearchOptions()
+
+            options.updateForSelfUserTeamRole(selfUser: selfUser)
+
+            ///search for the local user with matching user ID and active
+            let teamMembers: [Member]
+                let activeMembers = self.teamMembers(matchingQuery: "", team: selfUser.team, searchOptions: options)
```
typfel: Extra indentation here.

Source/UserSession/Search/SearchTask.swift
```diff
@@ -87,15 +87,48 @@ public class SearchTask {
...
+
+        context.performGroupedBlock {
+            let selfUser = ZMUser.selfUser(in: self.context)
+
+            var options = SearchOptions()
+
+            options.updateForSelfUserTeamRole(selfUser: selfUser)
+
+            ///search for the local user with matching user ID and active
+            let teamMembers: [Member]
```
typfel: No need to declare this ahead of time.

billypchan:
@zenkins release major

PR #1005, opened 2019-04-04
Fix initial sync-is-completed state handling

typfel:
### Issues

The certain elements of the UI (group settings) would be disabled when they shouldn't.

### Causes

The UI elements were waiting for the initial sync to complete which now never happens since we've split up the syncing into slow sync and quick sync.

### Solutions

Set `hasCompletedInitialSync` to `YES` when initialising the `ZMUserSession` if there's no need to perform a slow sync.


typfel:
@zenkins release patch

PR #1006, opened 2019-04-08
Remove usage of recentMessages property

typfel:
## What's new in this PR?

In https://github.com/wireapp/wire-ios-data-model/pull/708 the `recentMessage` property was removed, this PR removes all usage of that property.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/708


typfel:
@zenkins release major

PR #1010, opened 2019-04-17
Update copy for availability local notification

typfel:
## What's new in this PR?

Copy has been shortened to better fit on smaller displays.


typfel:
@zenkins release patch

PR #1011, opened 2019-04-17
Support switch to a custom backend with deeplink

vytis:
### Issues

We want to use a deeplink to switch to custom backend endpoints.

### Solutions

Switching backend involves several steps:
1. Parse the deeplink to extract the configuration JSON URL and signal UI to handle it
2. UI will call the method to switch to a custom backend using the URL from step 1.
3. The configuration is downloaded and all the relevant state is changed to sue new endpoints.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/135

## To Do

- [x] Add tests


Source/SessionManager/SessionManager+SwitchBackend.swift
```diff
@@ -0,0 +1,55 @@
...
+    public enum SwitchBackendError: Swift.Error {
+        case loggedInAccounts
+        case invalidBackend
+    }
+    
+    public typealias OnBackendSwitchError = (SwitchBackendError) -> ()
+    public typealias MakeSwitch = () -> ()
+    public typealias CompletedSwitch = (BackendEnvironment?) -> ()
+    
+    public func switchBackend(configuration url: URL, onError: @escaping OnBackendSwitchError, completed: @escaping CompletedSwitch) -> MakeSwitch? {
```
vytis: Signature of this method is a bit funky because it returns a closure that does the actual work of fetching the configuration. It was done this way to notify UI of any errors immediately, but also allow for showing a confirmation alert before making the network request to fetch the configuration.

typfel: Couldn't we have two methods to avoid this funky signature?

`func canSwitchBackend()`
`func switchBackend(configuration: URL,  result: (Result<BackendEnvironment>) -> Void )()`

`switchBackend` would internally also call `canSwitchBackend` so we wouldn't lose any safety.

vytis: Will try to come up with something simpler, initially there were more errors we had to handle and maybe it does not make much sense anymore to do it this way.

vytis: @typfel changed the API basically to what you suggested

vytis:
There are not many tests for the main functionality, because it uses Transport API that is not easily mockable.

Source/SessionManager/SessionManager+SwitchBackend.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    public enum SwitchBackendError: Swift.Error {
+        case loggedInAccounts
+        case invalidBackend
+    }
+    
+    public typealias CompletedSwitch = (Result<BackendEnvironment>) -> ()
+    
+    public func canSwitchBackend() -> SwitchBackendError? {
+        let authenticatedAccounts = accountManager.accounts.filter { environment.isAuthenticated($0) }
```
typfel: So what happens to a logged out account if you switch backend and then try to re-login to that account?

vytis: Since the account id will not exist on the other backend nothing will happen. If you logout from the "custom" backend account and then try to switch back then it will work I suppose.

vytis: @typfel updated, checking all accounts now

vytis:
@zenkins release major

PR #1012, opened 2019-04-18
Re-fetch team users after email field was introduced

typfel:
## What's new in this PR?

Since backend introduced the `email` for team users we need to re-fetch all users in the team.


typfel:
@zenkins release patch

PR #1014, opened 2019-04-24
Update tests after deleting method

typfel:
## What's new in this PR?

Update tests after method was deleted in https://github.com/wireapp/wire-ios-data-model/pull/715

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/715


typfel:
@zenkins release major

PR #1015, opened 2019-04-26
Fix: typing indicator still visible when user is removed from conversation

vytis:
### Issues

If a user that is currently typing gets removed from conversation the typing indicator stays visible.

### Causes

We would only expire the indicator after 60s of inactivity, but ignore the fact that user already left the conversation.

### Solutions

Listen for `member-leave` event and reset the typing status.


vytis:
@zenkins  release patch

vytis:
@zenkins release patch

PR #1016, opened 2019-04-26
move libPhoneNumber from DM

billypchan:
## What's new in this PR?

Move `libPhoneNumber` form DM since it is only used in class `AddressBook`

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/717


billypchan:
@zenkins release major

PR #1019, opened 2019-05-03
Include hot fix for re-fetching of users

typfel:
## What's new in this PR?

We need to re-fetch users after the e-mail field was introduced on the backend.


typfel:
@zenkins release patch

PR #1020, opened 2019-05-06
Add method enqueueDelayedChanges for performing changes with a delayed save

typfel:
## What's new in this PR?

Add `enqueueDelayedChanges` to `ZMUserSession`, which the UI can use when it needs to do many changes in a tight loop. The enqueued changes will then be coalesced into one save operation.


typfel:
@zenkins release minor

typfel:
@zenkins release minor

PR #1022, opened 2019-05-07
Forward call event to notification dispatcher if the conversation is muted

typfel:
### Issues

If CallKit is enabled you don't receive any notifications for muted conversations.

### Causes

We don't forward call events to the notification dispatcher if we have CallKit enabled.

### Solutions

We should make an exception and forward call events if the conversation is muted.


typfel:
@zenkins release patch

PR #1023, opened 2019-05-14
ZIOS-11422: Make the conversation not secure when ignoring the new client alert in a call

alexaubry:
### Issues

When the user tries to answer a call and we discover an new client in a verified conversation, we show an alert to ask them if they want to continue.

If they refuse, we drop the call and leave the conversation state unchanged in the security with ignored value, which means we will show the same alert the second time the call is picked up.

### Causes

We do not make the conversation not verified when leaving the call.

### Solutions

Make the conversation not verified when leaving the call.

### Dependencies

_Needs release with:_

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/722
- [ ] https://github.com/wireapp/wire-ios-request-strategy/pull/142


alexaubry:
@zenkins release major

PR #1028, opened 2019-05-24
Forwarding only successful responses to AVS

vytis:
### Issues

When requesting call config we can receive responses that are of no interest for AVS - e.g. clients missing (status code 412). There is no need to forward those, only successful ones are useful.

### Solutions

Filter all responses and forward only those with status code 200.

### DIFF
--- a/Source/Synchronization/Strategies/CallingRequestStrategy.swift
+++ b/Source/Synchronization/Strategies/CallingRequestStrategy.swift
@@ -184,9 +184,9 @@ extension CallingRequestStrategy : WireCallCenterTransport {
             let genericMessage = ZMGenericMessage.message(content: ZMCalling.calling(message: dataString))
             
             self.genericMessageStrategy.schedule(message: genericMessage, inConversation: conversation) { (response) in
-                
-                completionHandler(response.httpStatus)
-                
+                if response.httpStatus == 201 {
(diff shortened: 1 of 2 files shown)


Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -74,13 +74,15 @@ extension CallingRequestStrategy : ZMSingleRequestTranscoder {
     public func didReceive(_ response: ZMTransportResponse, forSingleRequest sync: ZMSingleRequestSync) {
         
         zmLog.debug("Received response for \(self): \(response)")
-        var payloadAsString : String? = nil
-        if let payload = response.payload, let data = try? JSONSerialization.data(withJSONObject: payload, options: []) {
-            payloadAsString = String(data: data, encoding: .utf8)
+        if response.httpStatus == 200 {
```
typfel: Isn't the response 201 when sending a message?

vytis: This is for fetching call config, so it should return 200. However I have added check for sending messages also now.

typfel: Ah ok

vytis:
@zenkins release patch

PR #1029, opened 2019-05-29
Fix: allow joining other call when you already have an active call

vytis:
### Issues

There are situations when you would want to join another call while there is an ongoing call already. It was not working and after ending the current call you would not be able to join another call.

### Causes

There is a race condition because ending call is asynchronous. We would immediately try to start a call  after finishing the previous one, but this needs to happen in a completion handler.

### Solutions

Add a completion handler to the end call method. Also relaxed a bit what is considered an active call - added all states where CallKit would be involved.

### DIFF
--- a/Source/Calling/CallKitDelegate.swift
+++ b/Source/Calling/CallKitDelegate.swift
@@ -249,7 +249,7 @@ extension CallKitDelegate {
         }
     }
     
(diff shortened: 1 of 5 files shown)


vytis:
@zenkins release major

PR #1030, opened 2019-06-05
ZIOS-11451: Set legal hold status when sending message

alexaubry:
## What's new in this PR?

Bump request strategy with updates from https://github.com/wireapp/wire-ios-request-strategy/pull/145.


alexaubry:
@zenkins release major

PR #1033, opened 2019-06-06
Fix logout not working on non active account

typfel:
### Issues

If you get logged out when switching accounts, pressing the log out button doesn't work.

### Causes

We attempt to log out the selected account, according to the account manager, when you press the log out button but that's actually the account you were switching from since the switch never completed.

### Solutions

Expose which account an unauthenticated session is associated with so that the UI can pick the correct account to log out from.


typfel:
@zenkins release minor

PR #1038, opened 2019-06-13
ZIOS-11455: Fix missing push notifications

nicorsm:
### Issues

When killing the app, the first push notifications were not delivered.

### Causes

After https://github.com/wireapp/wire-ios-sync-engine/pull/986, the app doesn't deliver notifications when killed because it appears that a sync is in progress (sync phase is `fetchingMissedEvents`).

### Solutions

We're avoiding that by returning `.done` when the system is slow syncing, `.inProgress` otherwise.

## Notes

Thanks @typfel for the support!


typfel:
Tests are failing but it's unrelated to your changes. Do a `carthage update` and it should work again.

typfel:
@zenkins release patch

typfel:
@zenkins release patch

PR #1039, opened 2019-06-17
Public logging: push notification events

vytis:
### Issues

We would like to have more actionable logging in public builds. A good place to start is push notifications processing pipeline. 

### Solutions

Added necessary conformances for objects to be logged and converted logs with "Push" tags to use the new logging method that will print data in public builds.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/741
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/150


vytis:
@zenkins release major

PR #1040, opened 2019-06-17
Don't attempt to download the blacklist in the tests

typfel:
### Issues

Tests are unstable because downloading the blacklist fails

### Causes

We are attempting download the production blacklist in the tests which is not guaranteed to always be stable.

### Solutions

Disable the the blacklist downloading if `blacklistDownloadInterval` is `< 0` and pass `-1` in the tests.

## Notes

One could imagine mocking the blacklist in the tests but that requires quite some re-factoring.


typfel:
@zenkins release patch

PR #1041, opened 2019-06-18
ZIOS-11483: Add logic to create legal hold client from request

alexaubry:
### Issues

We want to create the legal hold device client before sending the request, so that we can reflect the state of legal hold properly, before the server sends us an update event.

### Solutions

- In the accept legal hold method of the user session, add the logic to create the client before sending the request
- If the request succeeds, nil out the pending request from the user
- If the request fails, delete the client that we created previously
- Implement end to end integration test with mock transport (including processing the event)

### To-Do

- [ ] Implement negative integration test

### Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-mocktransport/pull/121
- [ ] https://github.com/wireapp/wire-ios-request-strategy/pull/151

### Notes

The UI will no longer be responsible for calling `userDidAcceptLegalHoldRequest `, and we will need to remove these calls from UI after merging this PR.


Tests/Source/UserSession/ZMUserSessionLegalHoldTests.swift
```diff
@@ -0,0 +1,118 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireSyncEngine
+
+class ZMUserSessionLegalHoldTests: IntegrationTest {
```
typfel: Let's move this test class into the integration test folder.

alexaubry: Done

alexaubry:
@zenkins release major

PR #1042, opened 2019-06-19
Fix application getting logged out until next re-launch

typfel:
### Issues

If the app was launched before the phone had been unlocked for the first time after a reboot, the app would get logged out until the next re-launch.

### Causes

We were creating a SessionManager and attempting to load the active account even though that would fail since we can't access the account data due to the file protection being to set to `completeUntilFirstUserAuthentication`.

This scenario was supposed to be covered by `FileManager.default.executeWhenFileSystemIsAccessible` which would delay the creation of the `SessionManager` until the file system is accessible. Unfortunately that solution stopped working at some iOS version.

### Solutions

Implement a new version of `executeWhenFileSystemIsAccessible` which lives on the `ZMApplication`.  It will use two checks determine if we can access the account data:

- `isProtectedDataAvailable` which is true when the phone is currently unlocked
- `ZMPersistentCookieStorage.hasAccessibleAuthenticationCookieData` which is true if there any accessible authentication data in keychain.

It's important to check both since we want to distinguish between:

- File system is available but there's no user accounts created yet
- File system is not available and we therefore can't access the account data

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/141


typfel:
@zenkins release minor

PR #1045, opened 2019-06-21
AVS 5.1: participants change

vytis:
## What's new in this PR?

A slightly improved API was added to handle participants in the call. It directly provides a list of participants in a JSON instead of relying on C structs.


Source/Calling/WireCallCenterV3+Events.swift
```diff
@@ -170,12 +170,31 @@ extension WireCallCenterV3 {
             self.isReady = true
         }
     }
-
-    /// Handles other users joining / leaving / connecting.
-    func handleGroupMemberChange(conversationId: UUID) {
-        handleEvent("group-member-change") {
-            let members = self.avsWrapper.members(in: conversationId)
```
typfel: Can the `func members(in conversationId: UUID)` and `AVSCallMember` be deleted now?

vytis: Good point, will check

vytis: `AVSCallMember` call member is used in quite a few places, but removed the function.

typfel: Ah right, so the new `handleParticipantChange` still creates AVSCallMembers.

PR #1046, opened 2019-06-21
Bump components

alexaubry:
## What's new in this PR?

Bump data model with the contents of https://github.com/wireapp/wire-ios-data-model/pull/743.


alexaubry:
@zenkins release major

PR #1048, opened 2019-06-24
ZIOS-11499: Fix redundant legal hold update messages

alexaubry:
### Issues

When receiving a message with a legal hold hint, we would insert too many system messages:

```
<legal hold activated>
<message content>
<legal hold deactivated>
<legal hold activated>
```

### Causes

The reason behind this is that after we finish the request to the BE to verify the legal hold hint, we'd call `updateSecurityLevelIfNeededAfterFetchingClients`, which performs an unconditional legal hold state update before we actually fetch the legal hold client. This caused a legal hold disabled system message to be inserted, and a hold-enabled message to be inserted later after we actually fetch the device. 

### Solutions

Fix the legal hold integration tests to take the new logic into account?

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/744
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/154


Tests/Source/Integration/ConversationTests+LegalHold.swift
```diff
@@ -203,18 +203,14 @@ class ConversationTests_LegalHold: ConversationTestsBase {
...
-        let legalHoldUser = self.user(for: user1)!
         let selfUserClient = self.selfUser.clients.anyObject() as! MockUserClient
         let otherUserClient = user1.clients.anyObject() as! MockUserClient
         let conversation = self.conversation(for: selfToUser1Conversation)
         
         mockTransportSession.performRemoteChanges { (session) in
             session.registerClient(for: self.user1, label: "Legal Hold", type: "legalhold", deviceClass: "legalhold")
         }
-        
-        legalHoldUser.fetchUserClients()
```
typfel: Why are deleting this part of the test?

alexaubry: I thought we were testing that the client is fetched when processing the hint in the message? Besides, this line was making the test fail because it seems like it blocked the request from the hint processing to go through, because it hadn't completed at the time of sending the message with the hint.

typfel: No, this test is testing the case when someone sends an incorrect legal hold flag when there's not client changes. That's why it's important to fetch the clients before we receive the metadata flag.

alexaubry: After our discussion yesterday I updated the test to send a message, and fixed the issue in request strategy.

alexaubry:
@zenkins release major

quenna999:
hi there,we need apple enterprise account...
would like to pay for it.
contact me freely....
fanboer@gmail.com
skype:live:fanboer

PR #1049, opened 2019-06-26
Process the legal hold disabled event

typfel:
## What's new in this PR?

Process the legal hold disabled event in order detect when a pending legal hold request is canceled.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/142


typfel:
@zenkins release minor

PR #1050, opened 2019-06-26
Update error labels for when the password is wrong

typfel:
### Issues

An incorrect password for a legal hold request was reported as an unknown error.

### Causes

The backend sends `access-denied` or `invalid-payload` when password is wrong and not `invalid-credentials`.

### Solutions

Update error label matching.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/122


typfel:
@zenkins release patch

PR #1052, opened 2019-07-08
Fix: When removing a device, no alert dialog is shown after entered a short invalid password

billypchan:
### Issues

When removing a device, no alert dialog is shown after entered a short invalid password. The spinner never dismisses. 

### Causes

The back-end gives a different error response other than invalid password error.

### Solutions

Treat `"bad-request"` response as invalid password response


billypchan:
@zenkins release patch

PR #1057, opened 2019-07-10
Add option for deleting account data when authentication is invalidated

typfel:
## What's new in this PR?

Add an option for deleting account data when users's authentication is invalidated

### Solutions

I'm introducing `SessionManagerConfiguration` which holds configurable options of the `SessionManager`. This will allow us to add new options in the future without having to change the signature of the `SessionManager` init method.

In this PR I'm adding `deleteAccountOnAuthenticationFailure` which controls if the account data is deleted in addition to the cookie when a user session authentication gets invalidated.

## Notes

While adding tests for the new authentication behaviour I noticed that the integration tests where using two separate cookie storages: `ztest.example.com` and `example.com`.  I fixed this by always creating the cookie storage from the `MockEnvironment`.

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -559,6 +559,18 @@ public protocol ForegroundNotificationResponder: class {
         }
     }
     
+    fileprivate func logout(account: Account, error: Error? = nil) {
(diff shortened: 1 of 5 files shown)


nicorsm:
@zenkins release major

PR #1058, opened 2019-07-15
Jailbreak detection

nicorsm:
## What's new in this PR?

Draft PR for jailbreak detection with blacklisting.

- [x] Jailbreak detection: using `JailbreakDetector.swift` by calling the public method `isJailbroken()` (true/false)
- [x] Add support for jailbreak detection inside `SessionManager` and add a delegate method to the `SessionManagerDelegate` in order to interact with UI
- [x] Add to the `SessionManagerConfiguration` object and to the configuration file (https://github.com/wireapp/wire-ios-build-configuration/pull/11/files) the two flags `blacklistAccountOnJailbreakDetection` and 
`deleteAccountOnJailbreakDetection`
- [x] Support on UI side: similar to what we're doing right now with blacklisting and `BlacklistViewController` (reuse with different copy?)
- [x] Avoid iOS Simulator to be recognized as jailbroken device

### DIFF
--- a/Source/SessionManager/JailbreakDetector.swift
+++ b/Source/SessionManager/JailbreakDetector.swift
@@ -1,0 +1,126 @@
+//
+// Wire
+// Copyright (C) 2019 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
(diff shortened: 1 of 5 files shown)


typfel:
Do you have an idea how can test this?

nicorsm:
@typfel not right now... I could try to write those files from the simulator and mock a jailbroken device, but I don't know how it will behave...

typfel:
@nicorsm I don't think we can test the jailbreak detection in meaningful way but we wan't to test the the logic we add to the `SessionManager`. We could for example inject a mock JailbreakDetector in the tests.

Source/SessionManager/JailbreakDetector.swift
```diff
@@ -19,10 +19,18 @@
...
+    func isJailbroken() -> Bool
+}
+
+public final class JailbreakDetector: JailbreakDetectorProtocol {
     
     private let fm = FileManager.default
     
+    public init() {
+        
+    }
```
billypchan: Is it still needed?

PR #1061, opened 2019-07-23
ZIOS-11523: App crashes after "log out" is tapped on "Too Many Devices" screen (Sync Engine)

nicorsm:
### Issues

App was crashing when tapping logout button in the "Too Many Devices" screen during login flow because of a fatal error.

### Solutions

Fatal error is thrown because the `unauthenticatedSession` object doesn't have an `accountId`. This happens because `accountId` doesn't get updated if the object is already instantiated. I'm now generating the user session if the `accountId` changes.

Previous PR: https://github.com/wireapp/wire-ios/pull/3511


nicorsm:
@zenkins release minor

PR #1062, opened 2019-07-24
Fix: iOS 13 -app crashes when calling appendNewConversationSystemMessage

billypchan:
## What's new in this PR?

This PR tries to fix an crash occurs on iOS13 only which occurs at the line calling `appendNewConversationSystemMessageAtTimestamp`. The reason of the crash is unclear and I translate it to swift to monitor `serverTimeStamp` and `conversation.activeParticipants`.

Bump frameworks for iOS 13 crash fix.
Rename method name for the framework changes.


billypchan:
@zenkins release major

PR #1065, opened 2019-07-31
Fix: renamed session manager flags

nicorsm:
## What's new in this PR?

Renamed session manager flags, following https://github.com/wireapp/wire-ios-build-configuration/pull/13


billypchan:
@zenkins release major

PR #1067, opened 2019-08-09
Fix: app crash when calling EventProcessingTracker.debugDescription

billypchan:
### Issues

App crashes when calling `EventProcessingTracker.debugDescription`

### Causes

Getting `eventAttributes` from the queue other than `isolationQueue` may cause an access violation. 

### Solutions

Get `persistedAttributes` from `isolationQueue`.

### DIFF
--- a/Source/Synchronization/EventProcessingTracker.swift
+++ b/Source/Synchronization/EventProcessingTracker.swift
@@ -114,8 +114,11 @@ import WireDataModel
     }
     
     override public var debugDescription: String {
-        let description = "\(persistedAttributes(for: eventName))"
+        let description = isolationQueue.sync {
+            "\(persistedAttributes(for: eventName))"
+        }
         dispatchEvent()
+        
         return description
     }
 }
--- a/Source/Synchronization/EventProcessingTracker.swift
+++ b/Source/Synchronization/EventProcessingTracker.swift
@@ -117,7 +117,6 @@ import WireDataModel
         let description = isolationQueue.sync {
             "\(persistedAttributes(for: eventName))"
         }
-        dispatchEvent()
         
         return description
     }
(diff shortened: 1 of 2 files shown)


Source/Synchronization/EventProcessingTracker.swift
```diff
@@ -114,8 +114,11 @@ import WireDataModel
     }
     
     override public var debugDescription: String {
-        let description = "\(persistedAttributes(for: eventName))"
+        let description = isolationQueue.sync {
+            "\(persistedAttributes(for: eventName))"
+        }
         dispatchEvent()
```
typfel: I know this was already here but it's weird because it seems to mutate the persisted attributes. I wouldn't expect calling debugDescription to mute anything. @nicorsm maybe you could explain this code?

nicorsm: Mmh it seems to be a simple getter of the content inside `eventAttributes`, don't know why it's mutating. Maybe it mutates just in the case of `eventAttributes[event]` is nil, then it returns an empty dictionary (`[:]`)? (see the lines above)

typfel: Maybe, but the mutation happens after we calculate description. From reading the code, it seems if you would call `debugDescription` it would print the collected attributes and then clear them.

I suggest looking at where we use `debugDescription` to see if it depends on `dispatchEvent()` being called, otherwise we should remove it from here.

ping @billypchan

billypchan: I remove the `dispatchEvent` call `debugDescription`, it should be called outside it.

billypchan:
@zenkins release patch

PR #1064, opened 2019-07-30
Feature: wipe database on jailbroken device

nicorsm:
## What's new in this PR?

- Added checks to `configuration.deleteAccountOnJailbreakDetection` flag
- Added support for deleting the accounts (still WIP, it crashes with multiple accounts)
- Added a test (needs to be improved)

PR #1068, opened 2019-08-13
Feature: wipe database on jailbroken device - PR reopen

nicorsm:
## What's new in this PR?

PR reopened from https://github.com/wireapp/wire-ios-sync-engine/pull/1064


Source/SessionManager/SessionManager.swift
```diff
@@ -575,19 +574,22 @@ public protocol ForegroundNotificationResponder: class {
...
-            tearDownBackgroundSession(for: account.userIdentifier)
-            deleteAccountData(for: account)
+            self.tearDownSessionAndDelete(account: account)
         } else {
             // Deleted the last account so we need to return to the logged out area
             logoutCurrentSession(deleteCookie: true, deleteAccount:true, error: NSError(code: .addAccountRequested, userInfo: nil))
         }
     }
     
+    public func tearDownSessionAndDelete(account: Account) {
```
typfel: This can be `fileprivate`?

Source/SessionManager/SessionManager.swift
```diff
@@ -827,6 +829,24 @@ public protocol ForegroundNotificationResponder: class {
...
     }
+
+    @objc public func checkJailbreakIfNeeded() -> Bool {
+        if jailbreakDetector?.isJailbroken() == true {
+            if configuration.blockOnJailbreakOrRoot {
+                self.delegate?.sessionManagerDidBlacklistJailbrokenDevice()
+                return true
+            } else if configuration.wipeOnJailbreakOrRoot {
+                logoutCurrentSession()
+                accountManager.accounts.forEach {
```
typfel: This will potentially tear down the user session while the UI is still using them since `logoutCurrentSession()` is not blocking until the UI is torn down.

Source/SessionManager/SessionManager.swift
```diff
@@ -827,6 +829,24 @@ public protocol ForegroundNotificationResponder: class {
             activeUserSession?.useConstantBitRateAudio = useConstantBitRateAudio
         }
     }
+
+    @objc public func checkJailbreakIfNeeded() -> Bool {
```
typfel: This only used front the tests? Then we can let it default to `internal` and remove the `@objc`?

Source/SessionManager/SessionManager.swift
typfel: A safer approach would be:

1. delete non active accounts using `delete(account:)`
2. delete active account `delete(account:)`

typfel: Ah but I see now that this is only ever run before any session have started.

Source/SessionManager/SessionManager.swift
```diff
@@ -482,13 +478,16 @@ public protocol ForegroundNotificationResponder: class {
...
+        
         // register for voIP push notifications
         self.pushRegistry.delegate = self
         self.pushRegistry.desiredPushTypes = Set(arrayLiteral: PKPushType.voIP)
         self.urlHandler = SessionManagerURLHandler(userSessionSource: self)
 
         postLoginAuthenticationToken = PostLoginAuthenticationNotification.addObserver(self, queue: self.groupQueue)
         callCenterObserverToken = WireCallCenterV3.addGlobalCallStateObserver(observer: self)
+        
+        guard !checkJailbreakIfNeeded() else { return }
```
typfel: From the specs it seems as if we should also perform this check when the app comes into foreground.

Source/SessionManager/SessionManager.swift
nicorsm: Yes, probably a leftover.

Source/SessionManager/SessionManager.swift
nicorsm: `@objc` can be removed, but not `public` if we want to use it inside of  `applicationWillEnterForeground` method.

typfel: I moved the call to `SessionManager` which also observes application state this also has the side effect that we will also run the detection also when not logged in.

Source/SessionManager/SessionManager.swift
typfel: I moved this into separate method `deleteAllAccounts(reason:)`

typfel:
@zenkins release minor

PR #1069, opened 2019-08-15
Support account deletion on launch

typfel:
### Issues

Deleting the account when cookie is expired didn't always work.

### Causes

We take a different code path when the app launches and we try to load the account. 

### Solutions

Also check if we should delete the account in this code path instead of just logging the user out.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/146

## Notes

`BackendEnvironmentProvider.isAuthenticated`is now verifying the expiration date in addition to verifying that cookie is present. This will allow us to know if we are not authenticated when offline.


typfel:
@zenkins release patch

PR #1071, opened 2019-08-22
Fix call participants not appearing in group calls

typfel:
### Issues

Call participants don't appear in group calls

### Causes

Since AVS 5.1 the call participants are received as JSON which fails to decode.

### Solutions

Fix call participant JSON decoding by correcting typo: `vrrecv` -> `vrecv`


typfel:
@zenkins release patch

PR #1072, opened 2019-08-22
Fix: support for new validators in Swift [SE]

nicorsm:
## What's new in this PR?

Added support for new validators in swift - type changes, functions rename, etc.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/772
- [ ] https://github.com/wireapp/wire-ios-utilities/pull/84


Source/Registration/AddressBook.swift
```diff
@@ -232,9 +232,9 @@ extension String {
                 .joined(separator: "")) // remove all non-digit
         }
         
-        var number : AnyObject? = self as AnyObject?
+        var number : Any? = self as Any?
         do {
-            try ZMPhoneNumberValidator.validateValue(&number)
+            _ = try ZMPhoneNumberValidator.validateValue(&number)
```
billypchan: We can add `@discardable` signature to prevent this.

Source/UserSession/ZMCredentials.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+public extension ZMPhoneCredentials {
+  
+    @objc static func validatePhone(_ phoneNumber: String) -> String? {
```
typfel: How about adding a `func validatePhoneNumber(_ phoneNumber: String)` on`ZMPhoneNumberValidator`? Then you would not need to duplicate this method on a `ZMPhoneCredentials` and `ZMAuthenticationStatus`

typfel: Or can can we use the already existing `ZMUser.validate(phoneNumber:)` method?

nicorsm: Mmh `ZMUser.validate(phoneNumber:)` uses an `inout` parameter, which can't be seen from objc. Better move it into `ZMPhoneNumberValidator`.

nicorsm:
@zenkins release major

PR #1073, opened 2019-08-22
Introduce MediaManagerType for easier mocking of the AVSMediaManager

typfel:
## What's new in this PR?

Introduce MediaManagerType for easier mocking of the AVSMediaManager

This is done in a effort to remove all usage of OCMock in the integration tests so that we can move them completely to Swift.


typfel:
@zenkins release patch

PR #1075, opened 2019-08-27
Add support for configuring message retention

typfel:
## What's new in this PR?

Add support for configuring message retention. If the message retention is non nil the session manager will delete expired messages when loading the user session.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/774


typfel:
@zenkins release minor

PR #1081, opened 2019-09-09
Expose mute state observer method

typfel:
## What's new in this PR?

Expose mute state observer method on the `VoiceChannel`


typfel:
@zenkins release minor

PR #1082, opened 2019-09-11
[Fix] Stop processing the team.conversation-delete event

typfel:
## What's new in this PR?

Stop processing the `team.conversation-delete event` since it has been replaced by the `conversation.delete` event in all cases.


typfel:
@zenkins release patch

PR #1083, opened 2019-09-11
Updated localization

typfel:
## What's new in this PR?

Pulled latest localization from crowdin.


typfel:
@zenkins release patch

PR #1085, opened 2019-09-13
[Fix] Don't report incoming calls from yourself as "missed"

typfel:
### Issues

If you have two devices A & B and you make an outgoing call from device A, it would get reported as a missed call on device B. In addition to the missed call system message this would also trigger a push notification.

### Solutions

It doesn't make sense to miss a call from yourself so we should instead report this as a "performed call".


typfel:
@zenkins release patch

PR #1086, opened 2019-09-17
Add method for logging out from a user session

typfel:
## What's new in this PR?

Add `logout(credentials:, completion:)` to the `ZMUserSession` which logs a user out  by deleting the self client. Deleting a client invalidates any cookie associated with the client.


typfel:
@zenkins release minor

PR #1088, opened 2019-09-23
Feature: authentication after reboot

nicorsm:
## What's new in this PR?

Added logic to logout a user after rebooting the device.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-build-configuration/pull/16


Source/SessionManager/SessionManagerConfiguration.swift
```diff
@@ -0,0 +1,116 @@
...
+    public static func load(from URL: URL) -> SessionManagerConfiguration? {
+        guard let data = try? Data(contentsOf: URL) else { return nil }
+        
+        let decoder = JSONDecoder()
+        
+        return  try? decoder.decode(SessionManagerConfiguration.self, from: data)
+    }
+}
+
+public extension SessionManagerConfiguration {
```
typfel: Let's move this extension to the `SessionManager`

Source/SessionManager/SessionManagerConfiguration.swift
```diff
@@ -0,0 +1,116 @@
...
+    
+    private static let previousSystemBootTimeContainer = "PreviousSystemBootTime"
+    
+    static var previousSystemBootTime: Double? {
```
typfel: Let's store an absolute time here (`Date`)

Source/SessionManager/SessionManager.swift
```diff
@@ -886,6 +820,16 @@ public protocol ForegroundNotificationResponder: class {
             self.delegate?.sessionManagerDidBlacklistJailbrokenDevice()
         }
     }
+    
+    internal func checkDeviceUptimeIfNeeded() {
+        let currentUptime = ProcessInfo.processInfo.systemUptime
```
typfel: In the specification it mentions that we should calculate the bootTime as `currentTime - currentUptime`. We could add an extension on processInfo which exposes this as `systemBootTime`.

typfel: Or if you think it's better to store the uptime directly were should rename the properties to reflect this.

typfel: The reason to store it as an absolute time is because otherwise  we would miss the check if you restart the phone and wait long enough.

Source/SessionManager/SessionManager.swift
```diff
@@ -886,6 +820,16 @@ public protocol ForegroundNotificationResponder: class {
             self.delegate?.sessionManagerDidBlacklistJailbrokenDevice()
         }
     }
+    
+    internal func checkDeviceUptimeIfNeeded() {
+        let currentUptime = ProcessInfo.processInfo.systemBootTime
```
typfel: Maybe rename to `let systemBootTime`

typfel:
There's no UI changes necessary for this feature?

nicorsm:
@typfel no, since it reuses the logic already implemented in the `logoutCurrentSession` function.

nicorsm:
@zenkins release minor

PR #1090, opened 2019-09-27
bump request strategy

billypchan:
## What's new in this PR?

bump request strategy


billypchan:
@zenkins release patch

PR #1091, opened 2019-09-27
Fix: re-authenticate on reboot triggering too often

typfel:
### Issues

When re-authenticate on reboot is enabled:

- User is logged out every time the app goes into the foreground
- User is logged out if no previous boot time was stored

#### User is logged out every time the app goes into the foreground:
This was due the clock can't be guaranteed to be the same on the milliseconds between runs

#### User is logged out if no previous boot time was stored
Value wasn't unwrapped so default comparison behaviour was applied

#### User is logged out every time the app goes into the foreground

Tolerate clock shifts up to 1 seconds (maybe needs to be higher?)

#### User is logged out if no previous boot time was stored

Unwrap the value before performing the comparison.

### DIFF
--- a/Source/SessionManager/SessionManager+Authentication.swift
+++ b/Source/SessionManager/SessionManager+Authentication.swift
@@ -18,7 +18,7 @@
 
 public extension SessionManager {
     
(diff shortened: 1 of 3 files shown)


typfel:
@zenkins release patch

PR #1092, opened 2019-09-27
Fix: logout not working for SSO users

typfel:
### Issues

SSO users would get an error when trying to logout

### Causes

SSO user's don't have a password but still included the `password` field in the DELETE request which the backend didn't accept.

### Solutions

Don't include an empty password in the DELETE request.


typfel:
@zenkins release patch

PR #1093, opened 2019-09-30
Fix: crash on canManageAccess - unified permissions control

nicorsm:
## What's new in this PR?

Unified permission control on group management after https://github.com/wireapp/wire-ios-data-model/pull/788 - now it's calling `.canModifyAccessControlSettings ` to check permissions.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/788


nicorsm:
@zenkins release minor

PR #1095, opened 2019-10-04
Chore: bump components

typfel:
## What's new in this PR?

Bump components due to major release


typfel:
@zenkins release major

PR #1097, opened 2019-10-10
Feature: download any existing labels when updating the app

typfel:
## What's new in this PR?

Fetch the labels when updating to a version which supports labels.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/798


typfel:
@zenkins release minor

PR #1098, opened 2019-10-11
Fix: incorrectly deleting folders when processing the event stream

typfel:
### Issues

Folders you just created would go missing if you put the app into background / foreground

### Causes

When we create folders we don't this will end up as events in the notification stream. If the put the app into background / foreground this will trigger us to process notification steam which will includes these events if nothing else happened in between.

create folder A
create folder B
create folder C

current state: [A, B, C]

process folder events:
event1 [A]
event2 [A, B]
event3 [A, B, C]

When processing event 1 we will delete folder B and C since they are not included in the update. This should be a problem since they will get added back in event 2 & 3. But this didn't happen because when we process event 2 & 3 we will fetch the existing folder B & C and update them instead because they haven't actually been deleted yet.

### Solutions

Perform a save on the context after deleting folders to make the change immediately


typfel:
@zenkins release patch

PR #1100, opened 2019-10-15
Fix: don't crash when modifying guest link without sufficient permissions

typfel:
### Issues

We would crash the app if you try to create guest link without sufficient permissions. This can happen if you are removed from the conversation while being at the guest settings screen.

### Causes

We intentionally crash the app if the user doesn't have sufficient permissions

### Solutions

Return an `.invalidOperation` error instead of crashing.


typfel:
@zenkins release patch

PR #1101, opened 2019-10-15
Feature: support for multi device calling

typfel:
## What's new in this PR?

Remove the assumption in the VoiceChannel API that a single user will not join a call from multiple devices.

- `VoiceChannel.participants` is now represented by a [CallParticipant] instead of a `[ZMuser]`
- Expose the `clientId` of a connected user so that we can distinguish the video streams from each other.


Source/Calling/VoiceChannelV3.swift
```diff
@@ -256,19 +236,3 @@ extension VoiceChannelV3 : CallObservers {
         return WireCallCenterV3.addMuteStateObserver(observer: observer, context: conversation!.managedObjectContext!)
     }
 }
-
-public extension CallState {
```
typfel: Deleted this extension since it was not used.

PR #1104, opened 2019-10-29
Fix: app crash when calling completedDownsampleOperation

billypchan:
## What's new in this PR?

Unwrap `operation.downsampleImageData` which the type is `Data!` before access.


billypchan:
@zenkins release patch

PR #1107, opened 2019-10-29
Clean up Cartfile

billypchan:
clean up version format in Cartfile


billypchan:
@zenkins release patch

PR #1108, opened 2019-10-30
[Fix] Remote video render

johnxnguyen:
### Issues

The video stream of another participant is black.

### Causes

In the video state handler, the client id is not passed down the chain to the necessary methods. Furthermore, two call members are still considered equal if they have the same user id. The client id was not being taken into account.

### Solutions

- Pass the client id down to the necessary methods.
- Define equality between two `AVSCallMembers` by their hashValue, which is a combination of their `userId` and `clientId`

## Notes

Some logic could be simplified. Although we should still be careful not to include duplicates in the call participant snapshot, the logic was originally in place to handle the case when a user joins the call from two different devices. But now we support multi device participation, so we only need to ensure there a no duplicate members.

### DIFF
--- a/Source/Calling/AVSCallMember.swift
+++ b/Source/Calling/AVSCallMember.swift
@@ -100,7 +100,7 @@ public struct AVSCallMember: Hashable {
     }
     
     public static func == (lhs: AVSCallMember, rhs: AVSCallMember) -> Bool {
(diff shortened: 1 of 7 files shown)


johnxnguyen:
@zenkins release patch

PR #1109, opened 2019-11-05
[Fix] Call participant notification

johnxnguyen:
## What's new in this PR?

This PR contains changes to address an AVS related crash.

### Issues

Since AVS 5.3, we now allow a user to join a group call with more than one device. This means that a call participant is no longer unique by their user id, we must also take their client id into account. Although we already store the client id, there are several places where we still only refer to the user id. This is due to the fact that when we post a `WireCallCenterCallParticipantNotification`, we were sending on the `UUID` referring to the user's remote identifier.

### Solutions

Pass the entire `CallParticipant` through the notification and make `CallParticipant` hashable over the user id and client id.

### DIFF
--- a/Source/Calling/AVSCallMember.swift
+++ b/Source/Calling/AVSCallMember.swift
@@ -99,8 +99,4 @@ public struct AVSCallMember: Hashable {
         hasher.combine(clientId)
     }
     
-    public static func == (lhs: AVSCallMember, rhs: AVSCallMember) -> Bool {
-        return lhs.hashValue == rhs.hashValue
-    }
-
 }
--- a/Source/Calling/AVSCallMember.swift
+++ b/Source/Calling/AVSCallMember.swift
@@ -99,4 +99,7 @@ public struct AVSCallMember: Hashable {
         hasher.combine(clientId)
     }
     
+    public static func == (lhs: AVSCallMember, rhs: AVSCallMember) -> Bool {
+        return lhs.hashValue == rhs.hashValue
+    }
 }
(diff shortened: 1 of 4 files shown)


billypchan:
@johnxnguyen test failed.

Source/Calling/AVSCallMember.swift
```diff
@@ -99,8 +99,4 @@ public struct AVSCallMember: Hashable {
         hasher.combine(clientId)
     }
     
-    public static func == (lhs: AVSCallMember, rhs: AVSCallMember) -> Bool {
-        return lhs.hashValue == rhs.hashValue
-    }
```
David-Henner: in `CallParticipantsSnapshot:77` 
```
    func update(updatedMember: AVSCallMember) {
        members = OrderedSetState(array: members.array.map({ member in
            member == updatedMember ? updatedMember : member
        }))
        notifyChange()
    }
```
This method relies on the `==` overload to determine if it should update the members list. 
Since the overload was removed, the list isn't updated when the participants network quality changes.
Hence the failing tests.

@johnxnguyen Was there a specific reason for removing this code?

johnxnguyen: Ahh, I assumed because we implemented hashable that it equates two instances by the hash, so I thought the overload was redundant.

David-Henner: I think it compares the properties of each instance passed as parameter

johnxnguyen:
@zenkins release major

PR #1110, opened 2019-11-08
[Fix] No video received in one to one call

johnxnguyen:
### Issues

When the other participant in a one to one call has video enabled, their video is not visible in the call.

### Causes

In group calls, whenever a participant joins the call we get their client id via the participants changed callback. However this is not the case in one to one calls. As a result, when the other participant enables video, the video state changed callback is invoked (with the client id), and we try to update our internal list of participants (in this case just one). There was a logic error when trying to find the participant instance to update (with the client id and the video state), where we were already expecting the client id to be set. As a result, we didn't update the participant, because we were trying to match the user id and client id.

### Solutions

Refactor the logic used in the video state changed handler that searches for the call member with the client id  (then falling back to only the user id), and use it when looking for the member to update in the `update(updatedMember:)` method.

## Notes

This is a tentative workaround until AVS 5.4 is ready, where we will get the client id in the call established callback.

### DIFF
--- a/Source/Calling/CallParticipantSnapshot.swift
+++ b/Source/Calling/CallParticipantSnapshot.swift
@@ -54,10 +54,7 @@ class CallParticipantsSnapshot {
     }
 
     func callParticpantVideoStateChanged(userId: UUID, clientId: String, videoState: VideoState) {
-        let participantsByUser = members.array.filter { $0.remoteId == userId }
-        let participant = participantsByUser.first { $0.clientId == clientId } ?? participantsByUser.first
-
-        guard let callMember = participant else { return }
+        guard let callMember = findMember(userId: userId, clientId: clientId) else { return }
 
         update(updatedMember: AVSCallMember(userId: userId, clientId: clientId, audioEstablished: callMember.audioEstablished, videoState: videoState))
     }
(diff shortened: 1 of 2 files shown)


Source/Calling/CallParticipantSnapshot.swift
```diff
@@ -94,4 +91,12 @@ class CallParticipantsSnapshot {
         
         return callMember.callParticipantState
     }
+
+    /// Tries to find the call member with the matching userId and clientId, otherwise the first member
+    /// with the matching userId.
+    ///
+    private func findMember(userId: UUID, clientId: String?) -> AVSCallMember? {
+        let participantsByUser = members.array.filter { $0.remoteId == userId }
+        return participantsByUser.first { $0.clientId == clientId } ?? participantsByUser.first
```
billypchan: We can merge filter and first into signal first closure.

johnxnguyen: Can you show an example what you mean?

billypchan:
```
 return members.array.first {
            $0.remoteId == userId && $0.clientId == clientId
        } ?? members.array.first
```

johnxnguyen: This isn't the same thing though. What we want to do is first find if there is a user that matches user id and client I'd, or if there is no matches just take the first member that matches user id.

johnxnguyen:
@zenkins release patch

Source/Calling/CallParticipantSnapshot.swift
billypchan: Then it should be:

```
return members.array.first {
            $0.remoteId == userId && $0.clientId == clientId
        } ?? members.array.first { $0.remoteId == userId }
```

but it is not simplified that much and I prefer not to change it.

PR #1113, opened 2019-11-20
bump request strategy

billypchan:
## What's new in this PR?

bump for the status limit fix.
move `ZMManagedObjectContextProvider` properties to public header since they are used in UI project.


billypchan:
@zenkins release major

PR #1114, opened 2019-11-21
downgrade AVS

billypchan:
## What's new in this PR?

downgrade AVS for release


johnxnguyen:
That is the public AVS version. If you want to downgrade the private version, do it in the  UI project in a file called "avs-versions"

PR #1123, opened 2019-12-17
Feature: Add a test to check the conversation.member-join event

KaterinaWire:
## What's new in this PR?

Add a test to check the conversation.member-join event which contains the new defined roles in the field `conversation_role`.


zenkins:

PR #1124, opened 2019-12-17
[Feature] Sync conversation role updates

johnxnguyen:
## What's new in this PR?

This PR provides a method to update the role for a given user in a given conversation. Invoking this method will perform a request to update the backend and then upon successful completion update the local database.


marcoconti83:
@zenkins release minor

PR #1125, opened 2019-12-17
Parse participant update events with roles

marcoconti83:
## What's new in this PR?

Add parsing roles from `conversation.member-update` events.
 
Before this PR, the `conversation.member-update` event was assuming that this event was only received for the self user. 

This PR adds parsing of roles for other members too, when the `conversation.member-update` refers to another user.


marcoconti83:
@zenkins release minor

PR #1127, opened 2019-12-18
Fix parsing member update event

marcoconti83:
the key is 'target', not 'id'


marcoconti83:
@zenkins release patch

PR #1129, opened 2019-12-19
[Fix] Save context after updating roles

johnxnguyen:
### Issues

Conversation change notifications were not being fired when we update the participant roles. This led to UI states being out of sync with the local database.

### Causes

The CoreData context was not being saved after modifying the participant roles.

### Solutions

Invoke a save.


johnxnguyen:
@zenkins release patch

PR #1130, opened 2019-12-19
Fix: Personal account admin has no "+ Participants" button

billypchan:
### Issues

Personal account admin can not add or remove users

### Causes

Actions is not loaded for personal account user

### Solutions

When a conversation is inserted, set `needsToDownloadRoles` to true


billypchan:
@zenkins release patch

PR #1131, opened 2019-12-19
Fix: processReceiptModeUpdate has nil lastServerTimestamp

billypchan:
## What's new in this PR?

Fix the crash when receipt event has a nil time stamp.


billypchan:
@zenkins release patch

PR #1132, opened 2019-12-19
Fetch conversation roles for conversations in other teams

marcoconti83:
## What's new in this PR?

Roles for a conversation should be fetched from a team or the conversation. When we have access to the team metadata (i.e. we are part of the team), we should fetch roles from the team. If not, then from the conversation.

The code was checking if the conversation is in a team, but not if it's our team. If the conversation is in another team, we should fetch the roles from the conversation, as we won't have access to that team.


marcoconti83:
@zenkins release minor

PR #1134, opened 2019-12-23
Remove roles refetch logic

marcoconti83:
The logic to fetch roles when needed has been moved one level down, to the data model layer. See https://github.com/wireapp/wire-ios-data-model/pull/860


billypchan:
@zenkins release patch

PR #1142, opened 2020-01-15
Log boot time updates

David-Henner:
## What's new in this PR?

Added logging for boot time updates


David-Henner:
@zenkins release patch

PR #1143, opened 2020-01-15
[Fix] wrong system boot time

David-Henner:
### Issues

`systemUptime` isn't updated when iOS system is asleep, which leads to calculating incorrect boot time.

### Solutions

Calculate boot time more accurately using the kernel boot time


David-Henner:
@zenkins release patch

PR #1145, opened 2020-01-16
Fix: bump data model version

KaterinaWire:
## What's new in this PR?

Bump the data model version.


KaterinaWire:
@zenkins release major

PR #1147, opened 2020-01-20
[Refactor] Don't publicly expose `ZMUser` in Typing Classes

johnxnguyen:
## What's new in this PR?

This PR is part of an ongoing effort to all public references that leak concrete user objects. See the architecture issues for more info: https://github.com/wireapp/ios-architecture/issues/89

Public exposure of `ZMUser` has been removed from the following cluster of files:

- `ZMTyping.m`
- `ZMTypingUsers.m`
- `ZMTypingUsersTimeout.m`

`ZMUser` was publicly exposed due to Objc/Swift interoperability. By converting these classes to Swift, I was able to make them all internal.

### Note
I also converted the tests to Swift too.


Source/Data Model/Typing.swift
```diff
@@ -0,0 +1,127 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class Typing {
+
+    #if DEBUG
+    public static var defaultTimeout: TimeInterval = 60
```
typfel: Debug `defaultTimeout` is the same as non-debug.

johnxnguyen: The difference is that in DEBUG it is mutable (for testing purposes).

typfel: Ah 🧐

Source/Data Model/TypingUsers.swift
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+}
+
+@objc public extension ZMConversation {
+
+    @objc (setIsTyping:)
+    func setIsTyping(_ isTyping: Bool) {
+        TypingStrategy.notifyTranscoderThatUser(isTyping: isTyping, in: self)
+    }
```
billypchan: Can we change this to a getter `var isTyping: Bool`?

Source/Data Model/Typing.swift
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+}
+
+// MARK: - TimerClient
+
+extension Typing: ZMTimerClient {
+
+    func timerDidFire(_ timer: ZMTimer!) {
+        guard timer == expirationTimer else { return }
```
typfel: Maybe do a `===` for pointer comparison which is what `==` means in Objective-C

Source/Data Model/TypingUsers.swift
johnxnguyen: Oh wait, this is not a stored property though.

Source/Data Model/Typing.swift
johnxnguyen: Done

Source/Data Model/TypingUsersTimeout.swift
```diff
@@ -0,0 +1,60 @@
...
+        let userIds = timeouts.keys
+            .filter { $0.conversationObjectId == conversation.objectID }
+            .map(\.userObjectId)
+
+        return Set(userIds)
+    }
+
+    func pruneConversationsThatHaveTimoutAfter(date pruneDate: Date) -> Set<NSManagedObjectID> {
+        let keysToRemove = timeouts
+            .filter { $0.value < pruneDate }
```
johnxnguyen: @typfel I was wondering about this. The method says "prune conversations that have timeout **after** this date", but we remove entries whose date is actually **before** the prune date. This was a direct translation from ObjC, and the tests are fine. So I'm guess it's just the wrong name for the method?

typfel: Cleared in person.

Source/Data Model/Conversation+TypingUsers.swift
```diff
@@ -0,0 +1,40 @@
...
+        TypingStrategy.notifyTranscoderThatUser(isTyping: isTyping, in: self)
+    }
+
+    var typingUsers: [UserType] {
+        guard let users = managedObjectContext?.typingUsers?.typingUsers(in: self) else { return [] }
+        return Array(users)
+    }
+
+    /// Strictly for UI tests. Remove once mockable conversation abstraction exists.
+    func setTypingUsers(_ users: [UserType]) {
```
johnxnguyen: @typfel , regarding the UI tests that need to set the typing users for the conversation: I couldn't find a way to mock the typing users without a large scale refactoring. So I've decided to make this public method for testing purposes.

Ideally we shouldn't expose this, but also ideally we should have a conversation abstraction like `UserType` with the aim of removing all CoreData from our UI tests. I'll create a ticket to track this issue.

typfel: 👍

PR #1149, opened 2020-01-21
Log error instead of fatal

billypchan:
### Issues

App crash after adding a catch in  https://github.com/wireapp/wire-ios-sync-engine/pull/1103

### Causes

Calling `fatal` always crashes

### Solutions

In this case, we should not crash the app when `enumerateContacts` throws. Logging the error instead.

### DIFF
--- a/Source/Registration/ContactAddressBook.swift
+++ b/Source/Registration/ContactAddressBook.swift
@@ -19,8 +19,10 @@
 import Foundation
 import Contacts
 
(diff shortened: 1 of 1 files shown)


johnxnguyen:
Do we know what the caught error is?

billypchan:
@johnxnguyen the thrown error  is not include in the crash log and let see if we can log it.

billypchan:
@zenkins release patch

PR #1150, opened 2020-01-22
[Refactor] UserType in Conversation Classes

KaterinaWire:
## What's new in this PR?

This PR is part of an ongoing effort to all public references that leak concrete user objects. See the architecture issues for more info: wireapp/ios-architecture#89

Public exposure of ZMUser has been removed from the following cluster of files:

- Conversation+Participants.swift
- Conversation+Role.swift

### DIFF
--- a/Source/Data Model/Conversation+Participants.swift
+++ b/Source/Data Model/Conversation+Participants.swift
@@ -51,14 +51,15 @@ import Foundation
 
 extension ZMConversation {
     
-    public func addParticipants(_ participants: Set<ZMUser>, userSession: ZMUserSession, completion: @escaping (VoidResult) -> Void) {
+    public func addParticipants(_ participants: [UserType], userSession: ZMUserSession, completion: @escaping (VoidResult) -> Void) {
         
+        let users = participants.compactMap { $0 as? ZMUser }
         guard conversationType == .group,
-              !participants.isEmpty,
-              !participants.contains(ZMUser.selfUser(inUserSession: userSession))
+              !users.isEmpty,
+              !users.contains(ZMUser.selfUser(inUserSession: userSession))
         else { return completion(.failure(ConversationAddParticipantsError.invalidOperation)) }
         
-        let request = ConversationParticipantRequestFactory.requestForAddingParticipants(participants, conversation: self)
+        let request = ConversationParticipantRequestFactory.requestForAddingParticipants(Set(users), conversation: self)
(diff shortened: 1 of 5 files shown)


Source/Data Model/Conversation+Role.swift
```diff
@@ -22,13 +22,16 @@ extension ZMConversation {
...
     private typealias Factory = ConversationRoleRequestFactory
 
-    public func updateRole(of participant: ZMUser,
+    public func updateRole(of participant: UserType,
                            to newRole: Role,
                            session: ZMUserSession,
                            completion: @escaping (VoidResult) -> Void) {
 
+        guard let user = participant as? ZMUser else {
+            return
```
johnxnguyen: Should we also call the completion handler here with a failure?

KaterinaWire: Yes, it makes sense.

KaterinaWire: Done

Source/Data Model/Conversation+Participants.swift
```diff
@@ -51,14 +51,15 @@ import Foundation
 
 extension ZMConversation {
     
-    public func addParticipants(_ participants: Set<ZMUser>, userSession: ZMUserSession, completion: @escaping (VoidResult) -> Void) {
+    public func addParticipants(_ participants: [UserType], userSession: ZMUserSession, completion: @escaping (VoidResult) -> Void) {
         
+        let users = participants.compactMap { $0 as? ZMUser }
```
typfel: Why not simply cast the whole array like:

let user = participants as? [ZMUser]

Could be a part of the guard.

typfel:
So the Set of users was changed to an array because `UserType` doesn't confirm to `Hashable` and `Equatable`?

johnxnguyen:
> So the Set of users was changed to an array because `UserType` doesn't confirm to `Hashable` and `Equatable`?

Yeah correct. I'm facing this too in my refactoring, at first it seemed as though using an `Array` was fine here and there, but I think now it might be better to get `UserType` to implement `Hashable` and `Equatable`.

Source/Data Model/Conversation+Participants.swift
KaterinaWire: Done

KaterinaWire:
@zenkins release major

PR #1152, opened 2020-01-23
[Feature] Add function for looking up a custom domain

marcoconti83:
## What's new in this PR?

This PR adds a request strategy to look up custom domain configurations.

The class is `CustomDomainLookupRequestStrategy` and can be invoked by the UI layer using `ZMUserSession.lookupCustomDomain`.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-transport/pull/156


typfel:
Thanks for the PR 👍  AFAIK the domain look up will happen when the user is in the logged out state so the user session will not be available.

The existing business logic arounds SSO are living in `CompanyLoginRequester`

marcoconti83:
> Thanks for the PR 👍 AFAIK the domain look up will happen when the user is in the logged out state so the user session will not be available.
> 
> The existing business logic arounds SSO are living in `CompanyLoginRequester`. The `UnauthenticatedSession` might also be a fitting place.

Yeah, good point! I overlooked the authentication part.

typfel:
@zenkins release major

PR #1153, opened 2020-01-23
[Refactor] UserType in Legal Hold Classes

KaterinaWire:
## What's new in this PR?

This PR is part of an ongoing effort to all public references that leak concrete user objects. See the architecture issues for more info: wireapp/ios-architecture#89

Public exposure of ZMUser has been removed from the following cluster of files:

LegalHold.swift

### DIFF
--- a/Source/UserSession/ZMUserSession+LegalHold.swift
+++ b/Source/UserSession/ZMUserSession+LegalHold.swift
@@ -20,9 +20,9 @@ import Foundation
 import WireDataModel
 import WireTransport
 
-public enum LegalHoldActivationError: Error {
-    case userNotInTeam(UserType)
-    case invalidUser(UserType)
+public enum LegalHoldActivationError: Error, Equatable {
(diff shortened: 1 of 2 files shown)


Source/UserSession/ZMUserSession+LegalHold.swift
```diff
@@ -20,9 +20,9 @@ import Foundation
 import WireDataModel
 import WireTransport
 
-public enum LegalHoldActivationError: Error, Equatable {
-    case userNotInTeam(ZMUser)
-    case invalidUser(ZMUser)
+public enum LegalHoldActivationError: Error {
+    case userNotInTeam(UserType)
```
typfel: Maybe we just drop the `UserType` parameter in the error? It's not really useful since it will always be the self user. We could rename the errors to `selfUserNotInTeam` and `invalidSelfUser`

KaterinaWire:
@zenkins release major

PR #1155, opened 2020-01-27
[Fix] Add hotfix patch for migration the backend environment to the shared user defaults

typfel:
### Issues

If a user has his custom backend environment stored in the `.standard` user defaults, it's not accessible from the share extension.

### Solutions

Migrate any custom backend environment from the `.standard` to the `.shared` user defaults.

### Dependencies

- [x] https://github.com/wireapp/wire-ios-transport/pull/157


typfel:
@zenkins release minor

PR #1156, opened 2020-01-28
[Feature] Add method on the unauthenticated session to fetch SSO settings

typfel:
## What's new in this PR?

Add method on the unauthenticated session to fetch SSO settings


typfel:
@zenkins release minor

PR #1158, opened 2020-01-28
[Chore] Bump Data model

KaterinaWire:
## What's new in this PR?

A major DataModel bump to include changes from:

wireapp/wire-ios-data-model/pull/884
wireapp/wire-ios-data-model/pull/883


KaterinaWire:
@zenkins release major

PR #1159, opened 2020-01-29
[Refactor] Convert ZMUserSession to Swift

typfel:
## What's new in this PR?

We want to convert `ZMUserSession` to `Swift` since it will unlock several other classes to also be converted to Swift.

## TODO

- [x] Convert ZMUserSesson+Authentication.m to Swift
- [x] Convert ZMUserSesson+Background.m to Swift
- [x] Convert ZMUserSesson+Proxy.m to Swift
- [x] Convert ZMUserSesson+OTR.m to Swift
- [x] Convert ZMUserSesson+Registration.m to Swift
- [ ] ~~Convert ZMUserSession.m to swift~~
- [ ] ~~Rename ZMUserSesson to UserSession~~

## Notes

I'll do last two points in a separate PR since this PR is already big enough.


Source/UserSession/ZMUserSession+Authentication.swift
```diff
@@ -20,6 +20,66 @@ import Foundation
...
+        let uiMOC = managedObjectContext
+        let syncMOC = syncManagedObjectContext
+        
+        uiMOC?.performGroupedBlockAndWait {}
+        syncMOC?.performGroupedBlockAndWait {}
+        
+        tearDown()
+        
+        uiMOC?.performGroupedBlockAndWait {}
+        syncMOC?.performGroupedBlockAndWait {}
```
johnxnguyen: Just curious what this does.

Source/UserSession/ZMUserSession+Clients.swift
```diff
@@ -0,0 +1,97 @@
...
+    @objc(failedToDeleteClientsWithError:)
+    func failedToDeleteClients(_ error: Error)
+}
+
+extension ZMUserSession {
+    
+    /// Fetch all selfUser clients to manage them from the settings screen
+    /// The current client must be already registered
+    ///
+    /// Calling this method without a registered client will throw an error
```
johnxnguyen: Don't you mean it was cause a crash?

Source/UserSession/ZMUserSession+Authentication.swift
```diff
@@ -20,6 +20,66 @@ import Foundation
...
     
+    /// Whether the user completed the registration on this device
+    
+    @objc
+    public var registeredOnThisDevice: Bool {
+        return managedObjectContext.registeredOnThisDevice
+    }
+    
+    @objc(setEmailCredentials:)
+    func setEmailCredentials(_ emailCredentials: ZMEmailCredentials?) {
```
billypchan: it is possible to use a setter?

Source/UserSession/ZMUserSession+Authentication.swift
typfel: We wait until all enqueued tasks have finished running on the both contexts.

Source/UserSession/ZMUserSession+Authentication.swift
typfel: There's no corresponding getter method so a property doesn't fit, besides that I want to keep the API the same.

Source/UserSession/ZMUserSession+Clients.swift
typfel: Don't what behaviour it has actually but it obviously won't throw an error. 

I just copied the comment from the previous implementation 😬 

I'll update the comment.

typfel:
@zenkins release major

PR #1160, opened 2020-01-31
[Feature] parse company login input

David-Henner:
## What's new in this PR?

Added method to support parsing of the SSO / email given by the user for enterprise login


David-Henner:
@zenkins release minor

PR #1161, opened 2020-02-03
[Fix]  Link Preview Tests

KaterinaWire:
## What's new in this PR?

Fixed Link preview tests: we don't need to check all properties of LinkPreview object since Article property is deprecated.


KaterinaWire:
@zenkins release patch

PR #1162, opened 2020-02-05
[Chore] Bump components

KaterinaWire:
## What's new in this PR?

A major DataModel bump to include changes from:

wireapp/wire-ios-data-model/pull/890


KaterinaWire:
@zenkins release patch

PR #1163, opened 2020-02-05
Remove the need for mocking the user session with OCMock (Part 1)

typfel:
## What's new in this PR?

In order to convert the `ZMUserSession` to Swift we need to remove any usage of OCMock where it mocks the user session.

---

 Refactor search classes and tests to inject the `ZMManagedObjectContextProvider` and `TransportSessionType` instead of the `ZMUserSession`.


WireSyncEngine.xcodeproj/project.pbxproj
```diff
@@ -2551,6 +2593,30 @@
 		};
 /* End PBXProject section */
 
+/* Begin PBXReferenceProxy section */
```
typfel: I have data model pulled in. Will revert it

typfel:
@zenkins release major

PR #1164, opened 2020-02-07
[Refactor] Remove the need for mocking the user session with OCMock (Part 2)

typfel:
## What's new in this PR?

---

I've introduced two new internal protocols: `UserSessionInternal` and `SessionManagerInternal`. The idea is that this the protocols that should be used when internal classes needs to be coupled with the user session or the session manager. This allows us to replace them with mock objects in the tests and thus remove the need for `OCMock`.

I've applied this on two tests:

`Conversation_ParticipantsTests`
`CallKitDelegateTests`

I'm open comments & suggestions on how  to improve this.


typfel:
> An example might be redefining `SessionManagerInternal: SessionManagerType`, which has only one single method `performWithSession(for:block:)` as a `SessionProvider`, and using it like so:

After looking at bit more on this, I think it's not a good idea to expose a method like `performWithSession()` since it forces us to work with a large type like `UserSession`. An alternative approach would be to create a protocol/delegate, which provides the functionally that `CallKitDelegate` needs from the `SessionManager`, namely:

`lookupConversation(by callIdentifier: CallIdentifier, completionHandler: (ZMConversation) -> Void)`
`endAllCallsInAllSession()`

johnxnguyen:
> An alternative approach would be to create a protocol/delegate, which provides the functionally that CallKitDelegate needs from the SessionManager

This is essentially narrowing down the scope of the abstraction and turning it into something that the `CallKitManager` uses in a specific way, rather than the `SessionManager` providing some object that can be used an any way. I like it 👍 

It also reminds me of the discussion we had about how to define use cases.

typfel:
@zenkins release patch

PR #1165, opened 2020-02-11
[Refactor]  Remove the need for mocking the user session with OCMock (Part 3)

typfel:
## What's new in this PR?

---

Following the strategy we've arrived at in https://github.com/wireapp/wire-ios-sync-engine/pull/1164 I'm removing more usage of the `mockTransportSession`.

### DIFF
--- a/Tests/Source/DatabaseTest.swift
+++ b/Tests/Source/DatabaseTest.swift
@@ -88,23 +88,6 @@ class DatabaseTest: ZMTBaseTest {
             self.syncMOC.zm_fileAssetCache = fileAssetCache
             self.uiMOC.zm_userImageCache = userImageCache
         }
-        
-//        FileAssetCache *fileAssetCache = [[FileAssetCache alloc] initWithLocation:nil];
-//        UserImageLocalCache *userImageCache = [[UserImageLocalCache alloc] initWithLocation:nil];
-//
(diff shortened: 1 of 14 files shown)


Tests/Source/DatabaseTest.swift
```diff
@@ -77,19 +77,54 @@ class DatabaseTest: ZMTBaseTest {
...
+//        [self.syncMOC performGroupedBlockAndWait:^{
+//            self.syncMOC.userInfo[@"TestName"] = self.name;
+//            [self.syncMOC addGroup:self.dispatchGroup];
+//            [self.syncMOC saveOrRollback];
+//
+//            [self.syncMOC setZm_userInterfaceContext:self.uiMOC];
+//            [self.syncMOC setPersistentStoreMetadata:@(notificationContentVisible) forKey:@"ZMShouldNotificationContentKey"];
+//            self.syncMOC.zm_fileAssetCache = fileAssetCache;
+//            self.syncMOC.zm_userImageCache = userImageCache;
+//            }];
```
johnxnguyen: What's this?

billypchan: old objc code?

typfel: I ported some code from `MessagingTests.m` which should be deleted.

typfel:
@zenkins release patch

PR #1166, opened 2020-02-14
[Refactor] Self User Provider

johnxnguyen:
## What's new in this PR?

This PR introduces two new concepts, the `SelfUserProvider` and the `SelfUser` helper class. They are intended to solve issues in the UI project related to the tight coupling of the `MockUser` and `ZMUser` classes. 

This is achieved by providing a mechanism whereby we could dynamically inject a single `SelfUserProvider` instance (some object that can give us a reference to the self user, whether it is concrete or mocked). The provider is injected into a the `class var provider` of the `SelfUser` helper class.

An example of its usage.

```swift
// before
ZMUser.selfUser(in: someManagedObjectContext)

// after
SelfUser.provider = someProvider // set this once (like when the user session changes)
SelfUser.current // use this any time you need the self user
```

## Needs releases with

- [x] https://github.com/wireapp/wire-ios-data-model/pull/893
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/196


johnxnguyen:
@zenkins release major

PR #1167, opened 2020-02-14
Fix: move AVSLogObserver from UI project

billypchan:
## What's new in this PR?

Move `AVSLogObserver` form UI project and convert it to Swift.


billypchan:
@zenkins release minor

PR #1168, opened 2020-02-17
[Refactor] Remove the need for mocking the user session with OCMock (Part 4)

typfel:
## What's new in this PR?

---

Split up the `SessionManagerURLHandler` into smaller components, each dedicated to one task. The business logic for handling a url action is now done inside a `URLActionProcessor`, which is owned by either the `UserSession` or the `UnauthenticatedSession`. The session manager's role is now to coordinate which session gets to process the url actions. 

* CompanyLoginURLActionProcessor (UnauthenticatedSession)
* ConnectToBotURLActionProcessor (UserSession)
* DeepLinkURLActionProcessor (UserSession)

The `URLAction` enum no longer contains any error cases since `URLAction` will now throw an error if it's initialised with a malformed URL.

Some of the business logic regarding was previously done in the UI project, I've tried to move as much as possible of it into the processors. The UI still needs asks the users if he/she wants to perform an action or display error messages, this is now exposed via the `URLActionDelegate` protocol.

### TODOs

- [x] Write tests for `ConnectToBotURLActionProcessor`


Source/SessionManager/SessionManager+URLActions.swift
```diff
@@ -0,0 +1,92 @@
...
+    
+    /// React to the application being opened via a URL
+    ///
+    /// - parameter url: URL the application was launched with
+    /// - parameter options: options associated with the URL
+    @discardableResult
+    public func openURL(_ url: URL, options: [UIApplication.OpenURLOptionsKey: Any]) throws -> Bool {
+        guard let action = try URLAction(url: url) else { return false }
+        
+        if action.requiresAuthentication, let userSession = activeUserSession {
```
johnxnguyen: Is it correct that if the action requires authentication, we process it on the active user session?

johnxnguyen: Hmm.. maybe I misunderstand what `requiresAuthentication` means.

typfel: Yes, I think the meaning is = the action requires that you are logged in in order to be performed.

typfel:
@zenkins release major

PR #1169, opened 2020-02-17
[Refactor] Add property to safely get the current self user

johnxnguyen:
## What's new in this PR?

This PR adds a class property to `SelfUser` to optionally get the current self user. I realize that while in most cases can safely assume that one exists, there are also situations where we might not be so sure that one exists and need to safely access it. Such a case may be when accessing the self user in closures that for some reason exist beyond the life of a user session.

### DIFF
--- a/Source/Data Model/SelfUser.swift
+++ b/Source/Data Model/SelfUser.swift
@@ -36,4 +36,10 @@ public class SelfUser {
         guard let provider = provider else { fatalError("Self user provider not set") }
         return provider.selfUser
     }
+
+    /// The current self user, if it exists.
+
+    public class var currentSafe: (UserType & ZMEditableUser)? {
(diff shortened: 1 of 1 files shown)


typfel:
If you want to get it safely could also go through the provider like: `SelfUser.provider?.selfUser`

johnxnguyen:
Yeah it's true, but perhaps it's more visible through a named property?

typfel:
Also I was though the  static property `provider` would live in the UI. Do we need it in the sync engine?

johnxnguyen:
> Also I was though the static property `provider` would live in the UI. Do we need it in the sync engine?

Discussed offline: `SelfUser` should be moved to UI to avoid its usage in SyncEngine as it is not thread safe.

PR #1170, opened 2020-02-17
[Refactor] Remove SelfUser

johnxnguyen:
### Issues

The `SelfUser` helper is not thread safe and therefore shouldn't be used in SyncEngine due to thread requirements of the different managed object contexts.

### Solutions

Remove the class. (It will be added to the UI project instead).


johnxnguyen:
@zenkins release major

PR #1171, opened 2020-02-17
[Fix] Observer not being notified when clients list is empty

typfel:
### Issues

The device screen shows a spinner which doesn't go away.

### Causes

Spinner is shown when fetching self clients but it never goes away since the UI never gets notified that clients have been fetched. It doesn't get notified because there's a condition not call the observer if the list of clients is empty.

### Solutions

Always notify the observer, the existing logic got broken when it was translated to Swift https://github.com/wireapp/wire-ios-sync-engine/pull/1159.


typfel:
@zenkins release patch

PR #1173, opened 2020-02-19
[Refactor] Set message flags using Swift Protobuf API

KaterinaWire:
## What's new in this PR?

Tests fix in accordance with data model refactoring wireapp/wire-ios-data-model/pull/896


KaterinaWire:
@zenkins release major

PR #1174, opened 2020-02-20
[Fix] Deep link error alert while logged out

typfel:
### Issues

If the app is launched with a deep link action we should show an error dialogue if the user is not logged in.

### Causes

These actions were delayed to in order to be performed when a user session becomes available.

### Solutions

Throw an error immediately if the user is not authenticated.


typfel:
@zenkins release patch

PR #1175, opened 2020-02-20
[Fix] Cancel notifications on logout instead of on tear down

typfel:
### Issues

The `LocalNotificationDispatcher` was canceling notifications and performing a delayed save from its `tearDown` method. This is problematic since: 
- The user session is not in a known state during tear down and causes flakyness in tests.
- This could in theory get triggered when the system is under memory pressure, and we release our background user sessions. That would have the side effect of clearing all notifications from the Notification Center, which doesn't make any sense.

### Solutions

Remove the `tearDown` method and cancel notifications on logout instead.


typfel:
@zenkins release patch

PR #1176, opened 2020-02-20
[Refactor] Convert ZMUserSession to Swift

typfel:
## What's new in this PR?

Convert `ZMUserSession` to Swift. This will enable us clean up the public interface of many classes/protocols, which previously had to be public in order to be exposed to Objective-C.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-transport/pull/161


Source/UserSession/ZMUserSession.m
```diff
@@ -1,707 +0,0 @@
...
-}
-
-@end
-
-@implementation ZMUserSession(PushChannel)
-
-- (void)pushChannelDidChange:(NotificationInContext *)note
-{
-    BOOL newValue = [note.userInfo[ZMPushChannelIsOpenKey] boolValue];
-    self.pushChannelIsOpen = newValue;
```
typfel: This was never read anywhere so removed it and the observer.

Source/UserSession/ZMUserSession.m
```diff
@@ -1,707 +0,0 @@
...
-                                         context: self.managedObjectContext.notificationContext
-                                         object:nil
-                                         queue:nil
-                                         using:^(NotificationInContext *note) {
-                                             ZM_STRONG(self);
-                                             [self pushChannelDidChange:note];
-                                         }
-        ]];
-        self.networkIsOnline = YES;
-        self.managedObjectContext.isOffline = NO;
```
typfel: This property was never read so I removed it.

Source/UserSession/ZMUserSession.m
```diff
@@ -1,707 +0,0 @@
...
-                                                                                            syncMOC:self.syncManagedObjectContext];
-            
-            __weak id weakSelf = self;
-            [transportSession setAccessTokenRenewalFailureHandler:^(ZMTransportResponse * _Nonnull response) {
-                ZMUserSession *strongSelf = weakSelf;
-                [strongSelf transportSessionAccessTokenDidFail:response];
-            }];
-        }];
-        
-        self.commonContactsCache = [[NSCache alloc] init];
```
typfel: Not read anywhere so I deleted it.

Source/UserSession/ZMUserSession.m
```diff
@@ -1,707 +0,0 @@
...
-        [self registerForBackgroundNotifications];
-        
-        [self enableBackgroundFetch];
-
-        self.storedDidSaveNotifications = [[ContextDidSaveNotificationPersistence alloc] initWithAccountContainer:self.storeProvider.accountContainer];
-        [self observeChangesOnShareExtension];
-        
-        
-        [self.syncManagedObjectContext performBlockAndWait:^{
-            if (self.clientRegistrationStatus.currentPhase != ZMClientRegistrationPhaseRegistered) {
```
typfel: I didn't keep this `if statement` since this condition is checked again inside `prepareForClientRegistration`

Source/UserSession/ZMUserSession.m
```diff
@@ -1,707 +0,0 @@
...
-static NSString * const AppstoreURL = @"https://itunes.apple.com/us/app/zeta-client/id930944768?ls=1&mt=8";
-
-
-@interface ZMUserSession ()
-@property (nonatomic) ZMSyncStrategy *syncStrategy;
-@property (nonatomic) ZMOperationLoop *operationLoop;
-@property (nonatomic) ZMTransportRequest *runningLoginRequest;
-@property (nonatomic) id<TransportSessionType> transportSession;
-@property (atomic) ZMNetworkState networkState;
-@property (nonatomic) ZMBlacklistVerificator *blackList;
```
typfel: The blacklist was moved to the `SessionManager` when it was introduced. This property was always `nil` and can thus be removed.

Source/UserSession/ZMUserSession.m
```diff
@@ -1,707 +0,0 @@
...
-    self.syncStrategy = nil;
-    [self.operationLoop tearDown];
-    self.operationLoop = nil;
-    [self.transportSession tearDown];
-    self.transportSession = nil;
-    self.applicationStatusDirectory = nil;
-    self.localNotificationDispatcher = nil;
-    [self.blackList tearDown];
-    
-    __block NSMutableArray *keysToRemove = [NSMutableArray array];
```
typfel: The user info in the `NSManagedObjectsContext` gets torn down when the `ManagedObjectContextDirectory` is torn down: https://github.com/wireapp/wire-ios-data-model/blob/3223154b2d1ab591a835af499ad0a0389fa294d9/Source/ManagedObjectContext/ManagedObjectContextDirectory.swift#L58

So I didn't keep this code.

Source/UserSession/ZMUserSession.m
```diff
@@ -1,707 +0,0 @@
...
-
-@end
-
-
-@implementation ZMUserSession (LaunchOptions)
-
-- (void)didLaunchWithURL:(NSURL *)URL;
-{
-    if ([URL isURLForPhoneVerification]) {
-        [[NSNotificationCenter defaultCenter] postNotificationName:ZMLaunchedWithPhoneVerificationCodeNotificationName
```
typfel: No one was listening to this notification so I removed it.

Source/Synchronization/ZMHotFixDirectory.m
```diff
@@ -203,7 +202,7 @@ - (NSArray *)patches
 
 + (void)resetPushTokens
 {
-    [[NSNotificationCenter defaultCenter] postNotificationName:ZMUserSessionResetPushTokensNotificationName object:nil];
+    [[NSNotificationCenter defaultCenter] postNotificationName:ZMUserSession.registerCurrentPushTokenNotificationName object:nil];
```
johnxnguyen: Is this a different notification or just a name change?

Source/Data Model/ServiceUser.swift
```diff
@@ -181,7 +181,7 @@ public extension ServiceUser {
     
     func createConversation(in userSession: ZMUserSession, completionHandler: @escaping (Result<ZMConversation>) -> Void) {
         createConversation(transportSession: userSession.transportSession,
-                           eventProcessor: userSession.operationLoop.syncStrategy,
+                           eventProcessor: userSession.operationLoop!.syncStrategy,
```
billypchan: Is this force unwrapping safe?

Source/UserSession/ZMUserSession.swift
```diff
@@ -0,0 +1,485 @@
...
+    var notificationDispatcher: NotificationDispatcher
+    var localNotificationDispatcher: LocalNotificationDispatcher?
+    var applicationStatusDirectory: ApplicationStatusDirectory?
+    var topConversationsDirectory: TopConversationsDirectory
+    var callStateObserver: CallStateObserver?
+    var messageReplyObserver: ManagedObjectContextChangeObserver?
+    var likeMesssageObserver: ManagedObjectContextChangeObserver?
+    var urlActionProcessors: [URLActionProcessor]?
+    
+    public var managedObjectContext: NSManagedObjectContext { // TODO jacob we don't want this to be public
```
johnxnguyen: Do these TODOs depend on changes in the UI project?

Source/UserSession/ZMUserSession.swift
```diff
@@ -0,0 +1,485 @@
...
+            guard let value = managedObjectContext.persistentStoreMetadata(forKey: LocalNotificationDispatcher.ZMShouldHideNotificationContentKey) as? NSNumber else {
+                return false
+            }
+            
+            return value.boolValue
+        }
+        set {
+            managedObjectContext.setPersistentStoreMetadata(NSNumber(booleanLiteral: newValue), key: LocalNotificationDispatcher.ZMShouldHideNotificationContentKey)
+        }
+    }
```
johnxnguyen: As a side note, this would be an excellent use case for a property wrapper.

Source/Synchronization/ZMHotFixDirectory.m
typfel: Same notification was just exposed as different name in Swift.

Source/UserSession/ZMUserSession.swift
typfel: I saw that something like that recently. A proxy property delegate that delegating using key paths :)

Source/Data Model/ServiceUser.swift
typfel: No, doesn't look safe. I'll have another look.

Source/UserSession/ZMUserSession.swift
johnxnguyen: Something to do on a rainy day :)

typfel: When we have Swift 5 ...

johnxnguyen: Ah yeah... that's right!

Source/UserSession/ZMUserSession.swift
typfel: No, It's mostly tests written in Objective-C, which forces us to expose them.

typfel:
@zenkins release major

PR #1177, opened 2020-02-21
[Feature] Remove "displayName"

KaterinaWire:
## What's new in this PR?

Remove the "displayName" logic from the user.

## Needs releases with

- [ ] wireapp/wire-ios-data-model#897


typfel:
@KaterinaWire in the future it would be nice it you link to PRs which a PR depend on, in this case it would be the https://github.com/wireapp/wire-ios-data-model/pull/897

KaterinaWire:
> @KaterinaWire in the future it would be nice it you link to PRs which a PR depend on, in this case it would be the [wireapp/wire-ios-data-model#897](https://github.com/wireapp/wire-ios-data-model/pull/897)

Ok, I will

KaterinaWire:
@zenkins release major

PR #1178, opened 2020-02-25
[Refactor] Prepare for moving strategies to request-strategy framework

typfel:
## What's new in this PR?

We want to move the event processing strategies in sync-engine to the request-strategy framework. This PR is preparing for that move by removing any coupling in those strategies to other sync engine classes.

Strategies to move:

```
ZMConnectionTranscoder
ZMConversationTranscoder
LegalHoldRequestStrategy
PushTokenStrategy
UserClientRequestStrategy
CallingRequestStrategy
LabelDownstreamRequestStrategy
ZMMissingUpdateEventsTranscoder
```

Coupled classes which also need to be moved:
```
OperationStatus
SyncStatus
PushNotificationStatus
CallEventStatus
```

### TODOs

- [x] Extract the event processing part of UserClientRequestStrategy into a separate class.
- [ ] ~~Introduce a delegate/notification for when the self clients gets deleted~~

### Notes
I limited this PR to decouple the transcoder/strategies from the `ZMSyncStrategy` and to extract the event processing part of `UserClientRequestStrategy` into `UserClientEventConsumer`. In a follow up PR I will move the status classes out of the sync engine.


Source/Synchronization/Strategies/UserClientEventConsumer.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+
+import Foundation
+
+/// Consumes self user client update events
+///
+/// Self user clients are clients belonging to the self user.
+
+@objcMembers
+public class UserClientEventConsumer: NSObject, ZMEventConsumer {
```
billypchan, line 26: not related to PR but we can remove all `@objc` signature for public classes since UI project no longer needs it.

typfel: This class is public in order to expose it to Objective-C (ZMSyncStrategy.m), the UI doesn't care about this class.

typfel:
@zenkins release patch

PR #1181, opened 2020-03-04
[Fix] User session still being retained after being torn down

typfel:
### Issues

The UserSession would be retained after being torn down after we receive a memory pressure warning.

### Causes

We where iterating over the `backgroundUserSessions` dictionary and modifying it at the same time. This was causing the user session to be retained even though it's been removed from the dictionary.

### Solutions

Extract the list of sessions to tear down into a temporary variable.


typfel:
@zenkins release patch

PR #1183, opened 2020-03-10
[Refactor] Integrate AVS 5.6

johnxnguyen:
## What's new in this PR?

This PR integrates [API changes](https://github.com/wearezeta/avs/blob/master/docs/api/api_changes_5.6.txt) for AVS 5.6

### Refactoring

We now get the clientId in more callbacks, which allows us to refactor how we maintain the list of call participants. Currently, we only get the clientId when the video state changes or the list of participants changes. A consequence of this is that any changes to the participants where we only have the userId might update the wrong participant (since a single userId can be shared with different participants). This issue is now resolved with these changes.

### DIFF
--- a/Source/Calling/AVSWrapper.swift
+++ b/Source/Calling/AVSWrapper.swift
@@ -209,13 +209,12 @@ public class AVSWrapper: AVSWrapperType {
         }
     }
 
-    // TODO: Pass clientId down or discard it.
     private let closedCallHandler: CloseCallHandler = { reason, conversationId, messageTime, userId, clientId, contextRef in
         zmLog.debug("closedCallHandler: messageTime = \(messageTime)")
         let nonZeroMessageTime: UInt32 = messageTime != 0 ? messageTime : UInt32(Date().timeIntervalSince1970)
 
-        AVSWrapper.withCallCenter(contextRef, reason, conversationId, nonZeroMessageTime) {
-            $0.handleCallEnd(reason: $1, conversationId: $2, messageTime: $3, userId: UUID(rawValue: userId))
+        AVSWrapper.withCallCenter(contextRef, reason, conversationId, nonZeroMessageTime, clientId) {
+            $0.handleCallEnd(reason: $1, conversationId: $2, messageTime: $3, userId: UUID(rawValue: userId), clientId: $4)
         }
     }
 
--- a/Source/Calling/AVSWrapper.swift
+++ b/Source/Calling/AVSWrapper.swift
@@ -197,17 +197,15 @@ public class AVSWrapper: AVSWrapperType {
         }
     }
 
-
     private let dataChannelEstablishedHandler: DataChannelEstablishedHandler = { conversationId, userId, clientId, contextRef in
         AVSWrapper.withCallCenter(contextRef, conversationId, userId, clientId) {
             $0.handleDataChannelEstablishement(conversationId: $1, userId: $2, clientId: $3)
         }
     }
 
-    // TODO: Pass clientId down or discard it.
     private let establishedCallHandler: CallEstablishedHandler = { conversationId, userId, clientId, contextRef in
-        AVSWrapper.withCallCenter(contextRef, conversationId, userId) {
-            $0.handleEstablishedCall(conversationId: $1, userId: $2)
+        AVSWrapper.withCallCenter(contextRef, conversationId, userId, clientId) {
+            $0.handleEstablishedCall(conversationId: $1, userId: $2, clientId: $3)
         }
     }
 
--- a/Source/Calling/AVSWrapper.swift
+++ b/Source/Calling/AVSWrapper.swift
@@ -163,8 +163,7 @@ public class AVSWrapper: AVSWrapperType {
 
     // MARK: - C Callback Handlers
 
-    // TODO: Pass clientId down or discard it.
-    private let constantBitRateChangeHandler: ConstantBitRateChangeHandler = { _, clientId, enabledFlag, contextRef in
+    private let constantBitRateChangeHandler: ConstantBitRateChangeHandler = { userId, clientId, enabledFlag, contextRef in
         AVSWrapper.withCallCenter(contextRef, enabledFlag) {
             $0.handleConstantBitRateChange(enabled: $1)
         }
--- a/Source/Calling/AVSWrapper.swift
+++ b/Source/Calling/AVSWrapper.swift
@@ -175,13 +175,13 @@ public class AVSWrapper: AVSWrapperType {
         }
     }
 
-    private let incomingCallHandler: IncomingCallHandler = { conversationId, messageTime, userId, isVideoCall, shouldRing, contextRef in
-        AVSWrapper.withCallCenter(contextRef, conversationId, messageTime, userId, isVideoCall, shouldRing) {
-            $0.handleIncomingCall(conversationId: $1, messageTime: $2, userId: $3, isVideoCall: $4, shouldRing: $5)
+    private let incomingCallHandler: IncomingCallHandler = { conversationId, messageTime, userId, clientId, isVideoCall, shouldRing, contextRef in
+        AVSWrapper.withCallCenter(contextRef, conversationId, messageTime, userId, clientId, isVideoCall, shouldRing) {
+            $0.handleIncomingCall(conversationId: $1, messageTime: $2, userId: $3, clientId: $4, isVideoCall: $5, shouldRing: $6)
         }
     }
 
-    private let missedCallHandler: MissedCallHandler = { conversationId, messageTime, userId, isVideoCall, contextRef in
+    private let missedCallHandler: MissedCallHandler = { conversationId, messageTime, userId, clientId, isVideoCall, contextRef in
         zmLog.debug("missedCallHandler: messageTime = \(messageTime)")
         let nonZeroMessageTime: UInt32 = messageTime != 0 ? messageTime : UInt32(Date().timeIntervalSince1970)
 
--- a/Source/Calling/AVSWrapper.swift
+++ b/Source/Calling/AVSWrapper.swift
@@ -163,6 +163,7 @@ public class AVSWrapper: AVSWrapperType {
 
     // MARK: - C Callback Handlers
 
+    // TODO: Pass clientId down or discard it.
     private let constantBitRateChangeHandler: ConstantBitRateChangeHandler = { _, clientId, enabledFlag, contextRef in
         AVSWrapper.withCallCenter(contextRef, enabledFlag) {
             $0.handleConstantBitRateChange(enabled: $1)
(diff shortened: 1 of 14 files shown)


Source/Calling/AVSWrapper.swift
```diff
@@ -163,7 +163,8 @@ public class AVSWrapper: AVSWrapperType {
 
     // MARK: - C Callback Handlers
 
-    private let constantBitRateChangeHandler: ConstantBitRateChangeHandler = { _, enabledFlag, contextRef in
+    // TODO: Pass clientId down or discard it.
```
johnxnguyen: Oops, forgot about this. I think we can just discard the clientId.

Source/Calling/NetworkQuality.swift
```diff
@@ -22,4 +22,5 @@ public enum NetworkQuality: Int32 {
     case normal = 1
     case medium = 2
     case poor = 3
+    case problem = 4
```
typfel: What type of problem do we have here? Is it that there's no network route available? Maybe Julian has a proper term for this scenario?

Source/Calling/CallParticipantSnapshot.swift
```diff
@@ -20,83 +20,110 @@ import Foundation
...
     func callParticpantVideoStateChanged(userId: UUID, clientId: String, videoState: VideoState) {
-        guard let callMember = findMember(userId: userId, clientId: clientId) else { return }
-
-        update(updatedMember: AVSCallMember(userId: userId, clientId: clientId, audioEstablished: callMember.audioEstablished, videoState: videoState))
+        updateMember(userId: userId, clientId: clientId, videoState: videoState)
     }
 
-    func callParticpantAudioEstablished(userId: UUID) {
-        guard let callMember = members.array.first(where: { $0.remoteId == userId }) else { return }
+    func callParticpantAudioEstablished(userId: UUID, clientId: String) {
```
typfel: These methods all have the same typo: `Particpant` > `Participant`

Source/Calling/CallParticipantSnapshot.swift
```diff
@@ -20,83 +20,110 @@ import Foundation
...
+        updateMember(userId: userId, clientId: clientId, audioState: .established)
+    }
 
-        update(updatedMember: AVSCallMember(userId: userId, clientId: callMember.clientId, audioEstablished: true, videoState: callMember.videoState))
+    // FIXME: This never get's called. We would likely want to call this for the new network state.
```
typfel: Is this a lot of work to fix?

Source/Calling/CallParticipantSnapshot.swift
```diff
@@ -20,83 +20,110 @@ import Foundation
...
+    // MARK: - Helpers
 
-        update(updatedMember: AVSCallMember(userId: userId, clientId: callMember.clientId, audioEstablished: callMember.audioEstablished, videoState: callMember.videoState, networkQuality: networkQuality))
-    }
-    
-    func update(updatedMember: AVSCallMember) {
-        guard let targetMember = findMember(userId: updatedMember.remoteId, clientId: updatedMember.clientId) else { return }
+    /// Updates the locally stored member for the given userId and clientId with the given non nil properties.
+    ///
+    /// The locally stored member may not yet have a clientId; we know the userId from the incoming call event but
```
typfel: Have you asked if AVS could also provide the clientID in the incoming call handler?

Source/Calling/NetworkQuality.swift
johnxnguyen: According to the [docs](https://github.com/wearezeta/avs/blob/eaf7f75fa3d576477daa696d6131a053af8de352/docs/api/api_changes_5.6.txt#L13) there's a missing relay candidate.

Source/Calling/CallParticipantSnapshot.swift
johnxnguyen: I did fix this, but forgot to remove this comment. It's called in `WireCallCenterV3+Events`

Source/Calling/CallParticipantSnapshot.swift
johnxnguyen: No I haven't, but I will now :)

Source/Calling/NetworkQuality.swift
typfel: Ah I see, and AVS themselves use the case `.problem`.

Source/Calling/CallState.swift
```diff
@@ -66,17 +66,34 @@ public struct CallParticipant: Hashable {
 public enum CallParticipantState: Equatable {
     /// Participant is not in the call
     case unconnected
+    /// A network problem occured but calls may still connect
```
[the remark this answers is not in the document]
johnxnguyen: Fixed

Source/Calling/CallParticipantSnapshot.swift
johnxnguyen: Fixed

Source/Calling/CallParticipantSnapshot.swift
johnxnguyen: The AVS team added the clientId for the incoming and ended call handlers. This allowed me to make the `clientId` on `AVSCallMember` non nil, and so the snapshot code could be simplified.

Source/Calling/WireCallCenterV3.swift
```diff
@@ -400,14 +414,8 @@ extension WireCallCenterV3 {
         let started = avsWrapper.startCall(conversationId: conversationId, callType: callType, conversationType: conversationType, useCBR: useConstantBitRateAudio)
         if started {
             let callState: CallState = .outgoing(degraded: isDegraded(conversationId: conversationId))
-            
-            let members: [AVSCallMember] = {
-                guard let user = conversation.connectedUser, conversation.conversationType == .oneOnOne else { return [] }
-                return [AVSCallMember(userId: user.remoteIdentifier)]
-            }()
-
```
johnxnguyen: @typfel one consequence of making the `clientId` of `AVSCallMember` non optional means that we can't insert the connected user as the participant in the snapshot since although we know their user id, we don't know from which client they will answer from. The solution to this is to add the participant through the participant change handler. For me this makes sense, because starting a call should not have any participants until someone connects.

Source/Calling/WireCallCenterV3.swift
```diff
@@ -549,29 +557,36 @@ extension WireCallCenterV3 {
...
+                         clientId: String? = nil,
+                         messageTime: Date? = nil)
+    {
         callState.logState()
         var callState = callState
 
         switch callState {
         case .incoming(video: let video, shouldRing: _, degraded: _):
-            createSnapshot(callState: callState, members: [AVSCallMember(userId: userId!)], callStarter: userId, video: video, for: conversationId)
+            createSnapshot(callState: callState, members: [AVSCallMember(userId: userId!, clientId: clientId!)], callStarter: userId, video: video, for: conversationId)
```
johnxnguyen: I will clean this up to avoid force unwrapping.

Tests/Source/Calling/WireCallCenterV3Tests.swift
```diff
@@ -419,27 +521,7 @@ class WireCallCenterV3Tests: MessagingTest {
...
-        checkThatItPostsNotification(expectedCallState: .outgoing(degraded: false), expectedCallerId: selfUserID, expectedConversationId: oneOnOneConversationID) {
-            // when
-            _ = sut.startCall(conversation: oneOnOneConversation, video: false)
-            
-            // then
-            XCTAssertEqual(mockAVSWrapper.startCallArguments?.conversationType, AVSConversationType.oneToOne)
-            XCTAssertEqual(mockAVSWrapper.startCallArguments?.callType, AVSCallType.normal)
-            XCTAssertEqual(sut.callParticipants(conversationId: oneOnOneConversationID), [CallParticipant(user: oneOnOneConversation.connectedUser!, state: .connecting)])
-        }
-    }
```
johnxnguyen: I removed this test since we would no longer preemptively add the other user as a participant when the call starts.

Tests/Source/Calling/WireCallCenterV3Tests.swift
```diff
@@ -867,20 +1035,20 @@ extension WireCallCenterV3Tests {
...
-    func testThatItIgnoresIt_WhenGroupHandlerIsCalledForOneToOne() {
+    func testThatItDoesNotIgnore_WhenGroupHandlerIsCalledForOneToOne() {
         // when
         _ = sut.startCall(conversation: oneOnOneConversation, video: false)
-        callBackMemberHandler(conversationId: oneOnOneConversationID, userId: otherUserID, audioEstablished: false)
+        callBackMemberHandler(conversationId: oneOnOneConversationID, userId: otherUserID, clientId: otherUserClientID, audioEstablished: false)
         XCTAssert(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
         
         // then
-        XCTAssertTrue(sut.callParticipants(conversationId: oneOnOneConversationID).isEmpty)
```
johnxnguyen: And here we test to make sure that the participant in a one to one is actually added when they answer.

Source/Calling/WireCallCenterV3.swift
```diff
@@ -314,20 +314,34 @@ extension WireCallCenterV3 {
         return callSnapshots[conversationId]?.callStarter
     }
 
-    /// Call this method when the callParticipants changed and avs calls the handler `wcall_group_changed_h`
+    /// Call this method when the callParticipants changed and avs calls the handler `wcall_participant_changed_h`
     func callParticipantsChanged(conversationId: UUID, participants: [AVSCallMember]) {
-        guard callSnapshots[conversationId]?.isGroup == true else { return }
```
johnxnguyen: Here we no longer constrain this path for group conversations only, since now we don't preemptively add the other participant in a one to one call when the call is started.

Source/Calling/WireCallCenterV3.swift
typfel: I agree 👍

johnxnguyen:
@typfel could you please check my comments regarding the change of behavior and tell me if I'm missing something that could cause a regression?

typfel:
> @typfel could you please check my comments regarding the change of behavior and tell me if I'm missing something that could cause a regression?

So behaviour change is that `VoiceChannel.participants` will be empty for outgoing calls until someone has answered, if I understood correctly. 

Looking at where we query the call participants I didn't see any issue with this, except maybe `firstCallingParticipantOtherThanSelf` in the UI project. That property doesn't seem to be used though.

johnxnguyen:
> So behaviour change is that VoiceChannel.participants will be empty for outgoing calls until someone has answered, if I understood correctly.

Yes this is correct. I'll do a more thorough look in the UI.

> That property doesn't seem to be used though.

On this node, I've found quite a lot of calling code that doesn't seem to be used, I guess due to the evolution of AVS.

Source/Calling/CallState.swift
```diff
@@ -114,33 +131,6 @@ public enum CallState: Equatable {
...
-            self = .established
-        case WCALL_STATE_TERM_LOCAL: fallthrough
-        case WCALL_STATE_TERM_REMOTE:
-            self = .terminating(reason: .unknown)
-        default:
-            // WCALL_STATE_UNKNOWN can happen when we check the call state of a
-            // conversation id that isn't in the list of calls
-            self = .none
-        }
-    }
```
johnxnguyen: This wasn't being called anywhere.

johnxnguyen:
@zenkins release major

PR #1184, opened 2020-03-11
[Fix] Add more detailed logging around sending calling messages

typfel:
## What's new in this PR?

Add more detailed logging around sending calling messages


typfel:
@zenkins release patch

PR #1186, opened 2020-03-16
[FIX] Missing "You called" message after hanging up

marcomaddy1982:
### Issues

 No system message saying "You called" shows up after calling and hanging up

### Causes

client id was returning nil and it was not necessary

### Solutions

Removed not necessary handling clientId parameter


marcomaddy1982:
@zenkins release major

typfel:
Doesn't matter much for SE but this was not major release since no **public** interface changed.

PR #1187, opened 2020-03-16
[Fix]: tests were failing

marcomaddy1982:
### Issues

Some test after the last merge are failing

### Causes

Not updating tests before last merge

### Solutions

Remove not necessary clientId parameter from certain tests


marcomaddy1982:
@zenkins release minor

PR #1193, opened 2020-03-27
[Feature] Ignore team.member-join events

typfel:
## What's new in this PR?

We no longer want to maintain a local copy of all team members in the clients and can therefore ignore team.member-join events

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-12836


typfel:
> Will we also remove the enum case `.teamMemberJoin`?

Good question, if we remove unused events there's quite a few other events to clean up as well I think.

But I think it makes sense to remove them.

PR #1194, opened 2020-03-30
[Chore] Revert to AVS 5.5.6

johnxnguyen:
### Issues

There were some regressions introduced after integrating AVS 5.6. Some of these regressions can only be fixed with new changes on the AVS side. These changes are included as part of AVS 6, however AVS 6 integration is not complete yet.

So that these regressions don't block the next release, it has been decided to rollback to the previous AVS version.


johnxnguyen:
@zenkins release major

PR #1195, opened 2020-03-30
[Feature]: Digital signature request

marcomaddy1982:
## What's new in this PR?

- Added digital signature request in the strategy to get the Swisscom URL.

## Dependencies

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-12764
data-model: https://github.com/wireapp/wire-ios-data-model/pull/930


Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -25,30 +25,41 @@ public var signatureStatusPublic: SignatureStatus?
...
+        
         switch status.state {
         case .initial:
             break
          case .waitingForURL:
-            // TODO: post request (to get URL)
+            guard let requestSync = requestSync else {
+                return nil
+            }
+            signatureStatusPublic?.state = .pendingURL
```
typfel: Why do we need to switch to a different state here? Isn't it enough to switch the state when handling the response?

If the signature request would fail due to network issues, would expect it to be tried again. This won't happen if we switch to different state here.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -62,19 +73,57 @@ public final class SignatureRequestStrategy: AbstractRequestStrategy {
...
+        case requestSync:
+            switch (response.result) {
+                case .success:
+                    guard let responseDictionary = response.payload?.asDictionary() else {
+                        return
+                    }
+                    signatureStatusPublic?.state = .waitingForSignature
+                    consentURL = responseDictionary["consentURL"] as? String
+                    responseId = responseDictionary["responseId"] as? String
+                case .permanentError,
```
typfel: I would separate the .permanentError from the temporary errors. For temporary errors I would reset requestSync so that we can try the request again.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -62,19 +73,57 @@ public final class SignatureRequestStrategy: AbstractRequestStrategy {
...
-    public func didReceive(_ response: ZMTransportResponse, forSingleRequest sync: ZMSingleRequestSync) {
+        guard
+            let encodedHash = signatureStatus?.encodedHash,
+            let documentID = signatureStatus?.documentID,
+            let fileName = signatureStatus?.fileName
+        else {
+            return nil
+        }
         
+        let payload: [String: String] = [
```
typfel: How about defining a private struct, which implements the `Codable` protocol. See PushTokenStrategy for an example.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -62,19 +73,57 @@ public final class SignatureRequestStrategy: AbstractRequestStrategy {
...
+                           forSingleRequest sync: ZMSingleRequestSync) {
+        switch sync {
+        case requestSync:
+            switch (response.result) {
+                case .success:
+                    guard let responseDictionary = response.payload?.asDictionary() else {
+                        return
+                    }
+                    signatureStatusPublic?.state = .waitingForSignature
+                    consentURL = responseDictionary["consentURL"] as? String
```
typfel: We could also use a private `Codable` struct for the response parsing.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
marcomaddy1982: I added a new state cause the nextRequestIfAllowed() was called also during the time that the request was performing and an new one was created when not necessary. Basically the app was going in an infinite loop.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
marcomaddy1982: Do you mean that I should do:

`requestSync?.resetCompletionState()`

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
typfel: That should not be necessary since you are using `readyForNextRequestIfNotBusy`:

https://github.com/wireapp/wire-ios-request-strategy/blob/86d225ac3731a6eaa8cdda95ef4585728b1a4572/Sources/Request%20Syncs/ZMSingleRequestSync.m#L68

It shouldn't allow another request until the first one has finished. I would suggest to debug more in details why the request loop is happening.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
typfel: It's already done in https://github.com/wireapp/wire-ios-request-strategy/blob/develop/Sources/Request%20Syncs/ZMSingleRequestSync.m#L68

Take look at the response handling, you'll see that it already handles the retry logic.

PR #1196, opened 2020-03-30
[Feature] Add protocol type to help UI trigger remote data fetching

johnxnguyen:
### Issues

The UI may need to trigger remote data fetches for certain objects. This is typically done by setting flags such as `needsToBeUpdatedFromBackend`, which will automatically schedule downstream requests to the backend.

These properties however are defined on concrete managed objects and are therefore not exposed through the various abstraction protocols, such as `UserType` and `TeamType`.

### Solutions

To avoid casting to concrete types in the UI, a helper class is needed. It simply accepts an abstract type, casts to the concrete type and sets the flags. An additional benefit of this helper is that it will be easier to test that the UI code calls these trigger methods.

### DIFF
--- a/Source/Data Model/TeamMetadataFetchTrigger.swift
+++ b/Source/Data Model/TeamMetadataFetchTrigger.swift
@@ -1,0 +1,70 @@
+//
+// Wire
+// Copyright (C) 2020 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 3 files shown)


typfel:
What's the benefit of this approach over adding methods to the protocols?

```
protocol UserType {
   func fetchProfile()
   func fetchRichprofile
   func fetchMembership()
}

protocol TeamType {
   func fetchMetadata()
}
```

johnxnguyen:
> What's the benefit of this approach over adding methods to the protocols?

I was thinking that wasn't necessarily something we wanted to expose in those protocols, but thinking about it now does look like a better place to put this logic.

PR #1199, opened 2020-03-31
[Feature] Digital Signature Strategy Style Improvement

marcomaddy1982:
## Dependencies

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-12764
data-model: https://github.com/wireapp/wire-ios-data-model/pull/935
ui: wireapp/wire-ios#4236


Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -152,14 +160,22 @@ private struct SignaturePayload: Codable, Equatable {
     }
 }
 
+// MARK: - SignatureResponse
 private struct SignatureResponse: Codable, Equatable {
-    let consentURL: String?
+    let consentURLString: String?
     let responseId: String?
+    var consentURL: URL? {
```
typfel: `URL` is implements `Codable` so can declare it directly on type and remove this computed property.

marcomaddy1982: Done

PR #1200, opened 2020-03-31
[Fix] Support composite message in notification types

David-Henner:
## What's new in this PR?

- Add support for composite message in notifications

## Dependencies

https://github.com/wireapp/wire-ios-data-model/pull/937


Source/Notifications/Push notifications/Notification Types/LocalNotificationContentType.swift
```diff
@@ -54,6 +54,12 @@ public enum LocalNotificationContentType : Equatable {
             return .text(text, isMention: messageData.isMentioningSelf, isReply: messageData.isQuotingSelf)
         }
         
+        if let compositeData = (message as? ConversationCompositeMessage)?.compositeMessageData,
+            let textData = compositeData.items.compactMap({ $0.textData }).first,
+            let text = textData.messageText {
```
billypchan: I wrote similar logic in UI, we can group them up and put them back to Data Model later.

David-Henner:
@zenkins release minor

PR #1201, opened 2020-04-01
[Feature] Remove member download strategy

typfel:
## What's new in this PR?

Even though we removed member downloading from the slow sync we were still downloading them after the slow sync.

`MemberDownloadRequestStrategy` is strategy which downloads all members of team when the `needsToRedownloadMembers` flag set on the `Team`. This flag was set to true after the Team was downloaded.

Changes in this PR:
- Delete the MemberDownloadRequestStrategy
- Re-write a hot fix for re-downloading team members to use the `needsToBeUpdatedFromBackend` on the `Member`.
- Disable availability integration tests since they will fail until the new behaviour is implemented.

## Notes

I will delete `Team.needsToRedownloadMembers` property from data model separately, since it requires a migration.


Tests/Source/Synchronization/ZMHotFixTests.swift
```diff
@@ -147,7 +147,7 @@ class ZMHotFixTests_Integration: MessagingTest {
         
         syncMOC.performGroupedBlock {
             let selfUser = ZMUser.selfUser(in: self.syncMOC)
-            XCTAssertTrue(selfUser.team!.needsToRedownloadMembers)
+            XCTAssertTrue(selfUser.membership!.needsToBeUpdatedFromBackend)
```
johnxnguyen, line 150: Did you mean to check if all the members have `needsToBeUpdatedFromBackend` set to true?

typfel: Yes, but in this case the self user is the only member.

typfel: But yeah, maybe makes sense to add another user to the test, to verify that all members gets the flag set.

typfel: @johnxnguyen made the test a bit more robust.

PR #1206, opened 2020-04-06
[Feature]: Post Digital Signature

marcomaddy1982:
## Dependencies

JIRA:  https://wearezeta.atlassian.net/browse/ZIOS-12767
data-model: https://github.com/wireapp/wire-ios-data-model/pull/947


Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -233,5 +258,30 @@ private struct SignatureResponse: Codable, Equatable {
 // MARK: - SignatureRetrieveResponse
 private struct SignatureRetrieveResponse: Codable, Equatable {
     let documentId: String?
-    let cms: String?
+    let cms: Data?
+    
+    public init(from decoder: Decoder) throws {
```
johnxnguyen, line 263: Is this supposed to be public?

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -233,5 +258,30 @@ private struct SignatureResponse: Codable, Equatable {
...
+        cms = cmsEncodedData
+    }
 }
+
+// MARK: - CMSFileMetaDataInfo
+private struct CMSFileMetaDataInfo {
+    let url: URL?
+    let fileName: String?
+    
+    public init(url: URL? = nil, fileName: String? = nil) {
```
johnxnguyen, line 282: Is this supposed to be public?

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
marcomaddy1982: you are right, no it is not necessary.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -169,11 +169,36 @@ extension SignatureRequestStrategy: ZMSingleRequestTranscoder {
...
         } catch {
             Logging.network.debug("Failed to decode SignatureRetrieveResponse with \(error)")
         }
     }
+    
+    private func writeCMSSignatureFile(for data: Data?) -> CMSFileMetaDataInfo {
+        guard
+            let cmsData = data,
+            let docDirectory = FileManager.default.urls(for: .documentDirectory,
+                                                        in: .userDomainMask).first,
```
johnxnguyen, line 184: So we write the file outside of the the app container?

marcomaddy1982: Right question, this was not clear for me how usually you do, let's see what @typfel says about.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -169,11 +169,36 @@ extension SignatureRequestStrategy: ZMSingleRequestTranscoder {
...
+            return CMSFileMetaDataInfo()
+        }
+    
+        let cmsFileName = "\(fileName)(\(assetID))"
+        let cmsFileNameAndExtention = "\(cmsFileName).cms"
+        let cmsURL = docDirectory.appendingPathComponent(cmsFileNameAndExtention)
+        do {
+            try cmsData.write(to: cmsURL)
+        } catch {
+            // TO DO: Add logging
```
marcomaddy1982, line 197: @johnxnguyen and here which log do you usually use in this cases?

marcomaddy1982: ZMSLog(tag: "Conversations")?

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
typfel: The document directory is still inside the app container since every iOS app has its own sandbox: https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html

I would suggest to store this file in the tmp directory though since we will never reference it again after the signature has been sent.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -233,5 +259,30 @@ private struct SignatureResponse: Codable, Equatable {
...
+        else {
+            cms = nil
+            return
+        }
+        cms = cmsEncodedData
+    }
 }
+
+// MARK: - CMSFileMetaDataInfo
+private struct CMSFileMetaDataInfo {
```
typfel, line 279:
```suggestion
private struct CMSFileMetadataInfo {
```

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -19,6 +19,7 @@
 
 import Foundation
 
+private let log = ZMSLog(tag: "Conversations")
```
typfel, line 22: Let's use `Logging.network` also for the new logs.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -169,11 +170,36 @@ extension SignatureRequestStrategy: ZMSingleRequestTranscoder {
...
-            signatureStatus?.didReceiveSignature(data: decodedResponse.cms?.data(using: .utf8))
+            let cmsURLAndFileName = writeCMSSignatureFile(for: decodedResponse.cms)
+            signatureStatus?.didReceiveSignature(at: cmsURLAndFileName.url,
+                                                 fileName: cmsURLAndFileName.fileName)
         } catch {
             Logging.network.debug("Failed to decode SignatureRetrieveResponse with \(error)")
         }
     }
+    
+    private func writeCMSSignatureFile(for data: Data?) -> CMSFileMetaDataInfo {
```
typfel, line 181: I think we consider moving this method to the `SignatureStatus` then we wouldn't need to implicitly report errors by passing a `nil` filename.

marcomaddy1982: Done

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
marcomaddy1982: Done

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
marcomaddy1982: Done

PR #1209, opened 2020-04-08
[Feature] AVS 6: introduce AVSClient

johnxnguyen:
### Issues

A call participant is identified by both a user id and a client id. Currently the client id of a participant is shared by AVS through certain callbacks. As a result, we often have an incomplete / insufficient data to uniquely identify participants and it leads to the extra complexity of having an optional client id when it should be non optional.

### Solutions

With AVS 6, we are given the client id in more callbacks, which allows us to make the call participants client id non optional. This leads to further refactoring, such as modeling an `AVSClient` as a pair of user id and client id, to represent some device in a call.

## Notes

The AVS binary is not yet published, therefore the current version of AVS is running in this PR. As a result this may not build or tests may fail.

### DIFF
--- a/Source/Calling/AVSCallMember.swift
+++ b/Source/Calling/AVSCallMember.swift
@@ -91,7 +91,7 @@ public struct AVSCallMember: Hashable {
         case .connecting:
             return .connecting
         case .established:
-            return .connected(videoState: videoState, clientId: client.clientId)
+            return .connected(videoState: videoState)
         case .networkProblem:
             return .unconnectedButMayConnect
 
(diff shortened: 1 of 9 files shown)


typfel:
Is the `removeDuplicateMembers()` still necessary now that we have clientID & userID everywhere?

johnxnguyen:
Ah yeah that's a good point, I'll remove that.

johnxnguyen:
@typfel actually, the method is still useful because we need to convert the array to an ordered set state. However, we could pass in a set of members to ensure uniqueness. I would like to look into this in a later PR because I already have quite a few changes stack on top of this PR.

PR #1210, opened 2020-04-08
[Feature] AVS 6: provide a list of clients

johnxnguyen:
## What's new in this PR?

A new requirement of AVS is that a list of user clients for a particular conversation must be provided on demand. AVS makes this request through the new `wcall_req_clients_h` handler (aka `Handler.RequestClients`).

When the handler is called, we must determine all clients (except the self client) in the given conversation and return it to AVS in json format. A client (aka `AVSClient`) is simply a user id and client id pair which was introduce in https://github.com/wireapp/wire-ios-sync-engine/pull/1209.

This PR enables json encoding of `AVSClient`, defines the handler and sets up the callbacks.

## Notes

The AVS binary is not yet published, therefore the current version of AVS is running in this PR. As a result this may not build or tests may fail.

### DIFF
--- a/Source/Calling/AVSCallMember.swift
+++ b/Source/Calling/AVSCallMember.swift
@@ -139,21 +139,9 @@ extension AVSClient: Codable {
 
     enum CodingKeys: String, CodingKey {
         
-        case userid
-        case clientid
+        case userId = "userid"
+        case clientId = "clientid"
 
-    }
-
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        userId = try container.decode(UUID.self, forKey: .userid)
-        clientId = try container.decode(String.self, forKey: .clientid)
-    }
-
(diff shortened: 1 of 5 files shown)


Source/Calling/AVSCallMember.swift
```diff
@@ -117,23 +117,55 @@ public struct AVSClient: Hashable {
...
 }
 
-extension AVSClient {
+extension AVSClient: Codable {
 
-    struct CodingData: Codable {
+    enum CodingKeys: String, CodingKey {
+        
+        case userid
+        case clientid
```
typfel, line 143:
```suggestion
        case userId = "userid"
        case clientId= "clientid"
```

This way you can define the name-mapping without implementing `init(from decoder: Decoder)` and `encode(to encoder: Encoder)`

johnxnguyen:
I'm going to merge this despite the incorrect AVS binary so that I can produce UI builds. Once the right binary is released we can verify the feature branch.

PR #1211, opened 2020-04-08
[Fix] SSO detection on iOS 13

typfel:
### Issues

SSO code would get detected over and over again on iOS 13.

### Causes

I suspect that `UIPasteboard.changeCount` behaves differently on iOS 13 

### Solutions

Calculate `isNew` by comparing the SSO string instead.


typfel:
@zenkins release major

PR #1215, opened 2020-04-09
[Feature] AVS 6: send SFT call message

johnxnguyen:
## What's new in this PR?

This PR implements the logic to send an SFT calling message. 

The SFT call begins by setting the conversation type to `conference`. AVS will then invoke the `wcall_sft_req_h` handler (aka `SFTCallMessageSend`). In this handler we need to trigger a PUT request to the given url and passing the given data. After the request is made and the response received, the result is then passed back to AVS through the `wcall_sft_resp` function.

## Notes

This PR relies on an unpublished AVS binary and therefore will not compile in this PR. Nonetheless, if approved I would merge it to the feature branch and validate it later when possible.

### DIFF
--- a/Source/Calling/AVSWrapper.swift
+++ b/Source/Calling/AVSWrapper.swift
@@ -78,6 +78,7 @@ public class AVSWrapper: AVSWrapperType {
                               clientId,
                               readyHandler,
                               sendCallMessageHandler,
+                              sendSFTCallMessageHandler,
                               incomingCallHandler,
                               missedCallHandler,
                               answeredCallHandler,
--- a/Source/Calling/AVSWrapper.swift
+++ b/Source/Calling/AVSWrapper.swift
@@ -220,14 +220,14 @@ public class AVSWrapper: AVSWrapperType {
     }
 
     private let dataChannelEstablishedHandler: Handler.DataChannelEstablished = { conversationId, userId, clientId, contextRef in
-        AVSWrapper.withCallCenter(contextRef, conversationId, userId, clientId) {
-            $0.handleDataChannelEstablishement(conversationId: $1, client: AVSClient(userId: $2, clientId: $3))
+        AVSWrapper.withCallCenter(contextRef, conversationId) {
+            $0.handleDataChannelEstablishement(conversationId: $1)
         }
     }
 
     private let establishedCallHandler: Handler.CallEstablished = { conversationId, userId, clientId, contextRef in
-        AVSWrapper.withCallCenter(contextRef, conversationId, userId, clientId) {
-            $0.handleEstablishedCall(conversationId: $1, client: AVSClient(userId: $2, clientId: $3))
+        AVSWrapper.withCallCenter(contextRef, conversationId) {
+            $0.handleEstablishedCall(conversationId: $1)
         }
     }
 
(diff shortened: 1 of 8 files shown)


Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -190,7 +190,37 @@ extension CallingRequestStrategy : WireCallCenterTransport {
...
-    
+
+    public func sendSFT(data: Data, url: URL, completionHandler: @escaping ((Result<Data>) -> Void)) {
+        var request = URLRequest(url: url)
+        request.httpMethod = "PUT"
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        request.setValue("application/json", forHTTPHeaderField: "Accept")
+        request.httpBody = data
+
+        URLSession.shared.task(with: request) { data, response, error in
```
typfel, line 201: I'm thinking if we should not use the default shared `URLSession` since it will potentially cache things on disk. A better options would probably be to use an ephemeral session like we do when [downloading link previews](https://github.com/wireapp/wire-ios-link-preview/blob/1d53c1ce0038604a3ca7cfe1a2a1081af030b040/WireLinkPreview/PreviewDownloader.swift#L50)

Maybe should expose such session/method on the transport session.

johnxnguyen: Ah yes, I wasn't sure what was the right way to do this. I'll look into your suggestions.

Source/Calling/AVSWrapper.swift
```diff
@@ -145,6 +147,24 @@ public class AVSWrapper: AVSWrapperType {
...
+    /// Passes the response of the SFT calling config request to AVS.
+    public func handleSFTResponse(data: Data?, context: WireCallMessageToken) {
+        let error: Int32
+        let buffer: Data
+
+        if let data = data {
+            error = 0
+            buffer = data
+        } else {
+            error = EPROTO
```
typfel, line 159: Will add more error codes later? Now AVS won't now if the request failed due to network error or because the URL is malformed.

johnxnguyen: Yes, apparently they only care for a 0 or a non zero. But if they expect certain error codes then I'd be happy to supply them. @c-g-owen

Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -190,7 +190,37 @@ extension CallingRequestStrategy : WireCallCenterTransport {
...
-    
+
+    public func sendSFT(data: Data, url: URL, completionHandler: @escaping ((Result<Data>) -> Void)) {
+        var request = URLRequest(url: url)
+        request.httpMethod = "POST"
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        request.setValue("application/json", forHTTPHeaderField: "Accept")
+        request.httpBody = data
+
+        URLSession(configuration: .ephemeral).task(with: request) { data, response, error in
```
johnxnguyen, line 201: @typfel in your opinion is it sufficient to have `.ephemeral`?

typfel: Yes, I think that sufficient, that's what we do in the link preview framework. But are we creating a new `URLSession` for every request now? Shouldn't we create it once when we initialise the strategy?

johnxnguyen: Yup, I'll fix that.

PR #1220, opened 2020-04-20
[Feature] Add back syncing of team members

typfel:
## What's new in this PR?

It was decided that we still want to keep a local copy of all the team member for smaller teams

- Re-introduce the `. fetchingTeamMembers ` slow sync step, which downloads all the team members. If the team is too big keeping a local copy of all team members the `hasMore` flag will be `true`, which cause us to skip the syncing step.
- Re-introduce the processing of the `.teamMemberJoin` event. This is necessary in order to keep the list of local user up to date.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/149

## Notes

I've also migrated `PermissionsDownloadRequestStrategy` to use the `MembershipPayload` struct.

**JIRA:** https://wearezeta.atlassian.net/browse/ZIOS-13036


billypchan:
2 tests still fails
`testThatEstimatedUnreadCountIsIncreasedAfterRecevingATextMessage`
`testThatLatestConversationIsAlwaysOnTop`

billypchan:
failed tests:
`testThatLatestConversationIsAlwaysOnTop`
`testThatEstimatedUnreadCountIsIncreasedAfterRecevingATextMessage`

PR #1223, opened 2020-04-24
[Fix] Request loop when fetching members

typfel:
### Issues

We sometimes get into a request loop when fetching members.

### Causes

The permissions field on the member response was incorrectly marked as non-optional, causing the payload decoding to fail if that field was not present.

### Solutions

Mark permissions field as optional.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/970


typfel:
@zenkins release major

PR #1224, opened 2020-04-27
[Feature] AVS 5.6

johnxnguyen:
## What's new in this PR?

This PR integrates the following AVS 5.6 changes:

- new network state `problem`
- new audio state `networkProblem`
- changed callback signatures to include the client id
## Notes

The previous integration of AVS 5.6 (which has been reverted) contained several refactorings which are not included here, because it required some changes only introduced in AVS 6. This implementation solely addresses the API changes from the AVS library.


johnxnguyen:
@zenkins release major

PR #1226, opened 2020-04-27
[Feature] Refresh user data when adding a participant to a group fails

typfel:
## What's new in this PR?

In large teams you'll lazily discover that team users have been deleted by sending them a message, if you try to add them to a group conversation before that it will fail and we should recover by re-fetching the user profiles of the users you tried to add.


typfel:
@zenkins release patch

PR #1228, opened 2020-04-29
[Fix] Missing system messages when member deleted

johnxnguyen:
### Issues

When typing a message into a conversation that contains a deleted member, we discover they are deleted and remove them from the conversation, but we don't display a system message.

### Causes

Sending a message to a deleted member will cause the  member to be synced with the backend. This is how we discover if they have been deleted or not. However, existing logic will also cause the conversation to be synced to the backend. The conversation is synced first and as at result it already removes the deleted member, but it does not add the system message. When syncing the deleted member, we append a system message and remove them from all group conversations they belong to, except for the initial conversation because they have already been removed.

### Solutions

We need to sync the potentially deleted member before syncing the conversation. This can be achieved by modifying the predicate used to fetch conversations that need to be sync. We modify it so that it only fetches conversations that need to be synced and don't have any participants that are waiting to be synced.

Once the all the participants are synced, the conversation would be picked next time and synced as usual.


johnxnguyen:
@zenkins release minor

PR #1230, opened 2020-05-07
Bump Proto and data model

billypchan:
bump for fixing proto issue


billypchan:
@zenkins release patch

PR #1234, opened 2020-05-12
[Feature] Inform AVS of client list updates when conversation changes

johnxnguyen:
### Issues

When a client starts an SFT call, AVS will ask for the list of all other clients in the conversation. However, it's possible that the list is out of sync with the backend. 

### Causes

This can happen if new participants are added to the conversation after the call connects. It can also happen if the self client is unaware of other clients in the conversation. In this case, the backend will report the missing clients when the OTR call message is sent.

### Solutions

Observe changes to the participants for the conversation and recalculate the client list and send it to AVS when needed. Observation only begins once calculate the client list for the first time, and ends when the SFT call is ended. If missing clients were detected from the backend response, then existing logic will trigger a sync of the conversation, and any changes will be propagated to the conversation observer.

## Notes

This PR relies on an unpublished AVS binary and therefore will not compile in this PR. Nonetheless, if approved I would merge it to the feature branch and validate it later when possible.

### DIFF
--- a/Source/Calling/WireCallCenterV3+Events.swift
+++ b/Source/Calling/WireCallCenterV3+Events.swift
@@ -26,6 +26,11 @@ private let zmLog = ZMSLog(tag: "calling")
 extension WireCallCenterV3 : ZMConversationObserver {
 
     public func conversationDidChange(_ changeInfo: ConversationChangeInfo) {
+        handleSecurityLevelChange(changeInfo)
+        handleActiveParticipantsChange(changeInfo)
+    }
+
+    private func handleSecurityLevelChange(_ changeInfo: ConversationChangeInfo) {
         guard
             changeInfo.securityLevelChanged,
             let conversationId = changeInfo.conversation.remoteIdentifier,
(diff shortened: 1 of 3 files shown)


Source/Calling/WireCallCenterV3+Events.swift
```diff
@@ -26,6 +26,11 @@ private let zmLog = ZMSLog(tag: "calling")
 extension WireCallCenterV3 : ZMConversationObserver {
 
     public func conversationDidChange(_ changeInfo: ConversationChangeInfo) {
```
typfel, line 28: Be aware that observers are not called when application is running the background.

Source/Calling/WireCallCenterV3+Events.swift
```diff
@@ -56,6 +61,18 @@ extension WireCallCenterV3 : ZMConversationObserver {
...
+    private func handleActiveParticipantsChange(_ changeInfo: ConversationChangeInfo) {
+        guard
+            changeInfo.activeParticipantsChanged,
+            let conversationId = changeInfo.conversation.remoteIdentifier,
+            let completion = clientsRequestCompletionsByConversationId[conversationId]
+        else {
+            return
+        }
+
+        handleClientsRequest(conversationId: conversationId, completion: completion)
```
typfel, line 73: Note that we probably don't have clients for a newly added participant in a group. We only discover the clients when send/receive messages.

PR #1235, opened 2020-05-18
[Refactor] AVS 6 client discovery

johnxnguyen:
## What's new in this PR?

This PR extends the capability of `CallingRequestStrategy` to provide a way to fetch a list of all clients in a given conversation. This method is intended to be used to provide AVS with a client list for use with SFT calls and replaces the previous approach of generating a (potentially out of date) list via the conversation object.

It works by sending an OTR message with an empty payload and zero recipients in a specific conversation. The backend will return a 412 response indicating that the message is missing recipients (as expected). In the payload of the response we will find a list of all missing clients, which will be the list of all clients in the conversation. This list is then passed into a completion handler.

### Note

This is a follow up of https://github.com/wireapp/wire-ios-request-strategy/pull/224, where it was decided it was best to implement this functionality inside the `CallingRequestStrategy`.

### DIFF
--- a/Source/Synchronization/Strategies/CallingRequestStrategy.swift
+++ b/Source/Synchronization/Strategies/CallingRequestStrategy.swift
@@ -29,10 +29,16 @@ public final class CallingRequestStrategy : NSObject, RequestStrategy {
     private let managedObjectContext: NSManagedObjectContext
     private let genericMessageStrategy: GenericMessageRequestStrategy
     private let flowManager: FlowManagerType
+
+    private let callEventStatus: CallEventStatus
+
     private var callConfigRequestSync: ZMSingleRequestSync! = nil
     private var callConfigCompletion: CallConfigRequestCompletion? = nil
-    private let callEventStatus: CallEventStatus
-    private let clientDiscoveryStrategy: ClientDiscoveryRequestStrategy
+
+    private var clientDiscoverySync: ZMSingleRequestSync! = nil
+    private var clientDiscoveryRequest: ClientDiscoveryRequest?
+
+
     private let ephemeralURLSession = URLSession(configuration: .ephemeral)
     
     public init(managedObjectContext: NSManagedObjectContext, clientRegistrationDelegate: ClientRegistrationDelegate, flowManager: FlowManagerType, callEventStatus: CallEventStatus) {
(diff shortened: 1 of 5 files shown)


Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -254,3 +343,73 @@ extension CallingRequestStrategy : WireCallCenterTransport {
...
+            let container = try decoder.container(keyedBy: CodingKeys.self)
+            let nestedContainer = try container.nestedContainer(keyedBy: Clients.CodingKeys.self, forKey: .missing)
+
+            let userIds = nestedContainer.allKeys.compactMap { UUID(uuidString: $0.stringValue) }
+
+            clients = try userIds.map { userId in
+                let userIdKey = Clients.CodingKeys.userId(userId)
+                let clientIds = try nestedContainer.decode([String].self, forKey: userIdKey)
+                return Clients(userId: userId, clientIds: clientIds)
+            }
```
johnxnguyen, line 372: There's a few annoying things about the response payload, first it stores the userIds as keys, secondly the entry for the missing clients is an object, not an array.

So the decoding is a bit more explicit, where we have to get a nested container and create the coding keys manually. If there is a better way I'm all for it!

Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -234,6 +314,15 @@ extension CallingRequestStrategy : WireCallCenterTransport {
         }
     }
 
+    public func requestClientsList(conversationId: UUID, completionHandler: @escaping ([AVSClient]) -> Void) {
+        self.zmLog.debug("requestClientList() called, moc = \(managedObjectContext)")
+        managedObjectContext.performGroupedBlock { [unowned self] in
+            self.clientDiscoveryRequest = ClientDiscoveryRequest(conversationId: conversationId, completion: completionHandler)
```
typfel, line 320: AVS has promised not request two clients lists in parallel?

johnxnguyen: There's no explicit promise, why?

typfel: Doesn't the second `requestClientsList()` overwrite `self.clientDiscoveryRequest`. Imagine this would case problems if there are two calls active simultaneously.

johnxnguyen: I didn't think there could be more than one active call, but yeah better not to assume it.

johnxnguyen: I would like to address that in a separate PR though so that I can get a new AVS build ready for tomorrow's call.

typfel: Sure, don't how well it fits here but in other places we use `IdentifierObjectSync` when we are syncing multiple things.

typfel: > I didn't think there could be more than one active call, but yeah better not to assume it.

Depends on when this is requested. You can only join one call but there can of course be multiple ongoing calls which you have the option to join.

PR #1249, opened 2020-06-15
[Fix] Is typing updates sent too often

typfel:
### Issues

Is typing update requests are sent so often while typing that they trigger request loop warnings on internal builds.

### Causes

For every keystroke we call `ZMConversation.setIsTyping(true)` which posts a notification to the `TypingStrategy`. Upon receiving this notification the self user is marked as "typing" in that conversation, this leads to the creation of a `TypingEvent` which will create a request if the typing event is not equal to the typing event stored in `lastSentTypingEvent`. This equality is among other things based on the date when the event was created, if the event are closer in time than `Typing.defaultTimeout / Typing.relativeSendTimeout` they are considered equal. Those constants are currently defined as:

`Typing.defaultTimeout = 60`
`Typing.relativeSendTimeout = 5`

That means we'll send a request every 12 seconds while typing, which is enough to generate a request loop warning, which has it's limit set to 60 seconds.

### Solutions

Consider typing events equal if they are closer in time than 60 seconds.

### DIFF
--- a/Source/Data Model/Typing.swift
+++ b/Source/Data Model/Typing.swift
@@ -26,9 +26,6 @@ class Typing {
     public static let defaultTimeout: TimeInterval = 60
     #endif
 
-    /// We only send typing events to the backend every ZMTypingDefaultTimeout / ZMTypingRelativeSendTimeout seconds.
-    public static let relativeSendTimeout: TimeInterval = 5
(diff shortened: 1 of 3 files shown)


typfel:
@zenkins release patch

PR #1250, opened 2020-06-22
Clean up @objc public func

billypchan:
## What's new in this PR?

As the UI project is pure Swift, `@objc` signature is no longer needed. (Some of them are kept since they are called by objc callers in this project.)


billypchan:
@zenkins release major

PR #1251, opened 2020-06-22
Clean up @objc public enum

billypchan:
## What's new in this PR?

Clean up unnecessary `@objc public enum` and class


billypchan:
@zenkins release patch

PR #1252, opened 2020-06-22
App crash at lazy.filter

billypchan:
### Issues

App crash at `requestedPreviewAssets.lazy.filter` where `requestedPreviewAssets` is `[UUID : SearchUserAssetKeys?]`

### Causes

Unkown, but the `lazy.filter` can be refactored. 

### Solutions

User `first(where:)` instead and refactor with De Morgan's laws.


billypchan:
@zenkins release patch

PR #1255, opened 2020-06-25
[Feature]: digital signature feature flag

marcomaddy1982:
## What's new in this PR?

feature flag to enable the digitally signature.

## Dependencies

data-model: https://github.com/wireapp/wire-ios-data-model/pull/999
JIRA: https://wearezeta.atlassian.net/browse/ZIOS-13522


Source/Synchronization/Strategies/FeatureFlagRequestStrategy.swift
```diff
@@ -0,0 +1,141 @@
...
+    
+    private func update(with response: SignatureFeatureFlagResponse) {
+        guard let team = ZMUser.selfUser(in: syncContext).team else {
+            return
+        }
+        
+        FeatureFlag.insert(with: .digitalSignature,
+                           value: response.status,
+                           team: team,
+                           context: syncContext)
```
johnxnguyen, line 122: If I understand correctly, this will always create a new flag and insert it in Core Data. Don't we want to update an existing object if it exists?

Source/Synchronization/Strategies/FeatureFlagRequestStrategy.swift
```diff
@@ -0,0 +1,141 @@
...
+    }
+    
+    @objc
+    public override func nextRequestIfAllowed() -> ZMTransportRequest? {
+        guard
+            syncStatus.currentSyncPhase == .fetchingFeatureFlag,
+            let singleRequestSync = singleRequestSync
+        else {
+            return nil
+        }
```
johnxnguyen, line 55: Currently we only generate a request if we're doing a slow sync but will need to also make one every 24 hours. I've been thinking of a way to do this, I have this simple idea:

- Statically store a `Date` indicating the last time the request was made.
- In this guard statement, check that we're in a slow sync **OR** more than 24h have passed since the last request.

This is a very simple approach and has some downsides. For one, we wait for the strategy to be asked for a request, but I think it's safe to assume that it would happen at least once a day. Secondly, the date would be stored statically (with a default value in the distant past), so each time the app is launched we would probably fetch the feature flags. But I also don't think this is a major issue.

Source/Synchronization/Strategies/FeatureFlagRequestStrategy.swift
```diff
@@ -0,0 +1,162 @@
...
+        
+        FeatureFlag.fetchOrCreate(with: .digitalSignature,
+                                  value: response.status,
+                                  team: team,
+                                  context: syncContext)
+        syncContext.saveOrRollback()
+    }
+    
+    private func getSignatureFlagThresholdUpdateDate() -> Date? {
```
johnxnguyen, line 134: Name suggestion: `calculateDigitalSignatureFlagRefreshDate()`

Source/Synchronization/Strategies/FeatureFlagRequestStrategy.swift
```diff
@@ -0,0 +1,162 @@
...
+        guard
+            let digitalSignatureFlag = ZMUser.selfUser(in: syncContext).team?
+            .fetchFeatureFlag(with: .digitalSignature) else {
```
johnxnguyen, line 137: You could split this into two statements to make it a bit more readable:

```swift
let team = ...
let flag = ...
```

Source/Synchronization/Strategies/FeatureFlagRequestStrategy.swift
```diff
@@ -0,0 +1,163 @@
...
+    
+    private func update(with response: SignatureFeatureFlagResponse) {
+        guard let team = ZMUser.selfUser(in: syncContext).team else {
+            return
+        }
+        
+        FeatureFlag.fetchOrCreate(with: .digitalSignature,
+                                  value: response.status,
+                                  team: team,
+                                  context: syncContext)
```
johnxnguyen, line 130: Looking at this method in Data Model, it appears that we are still not updating the existing flag. I would expect to see some logic like:

```swift
if let existing = ... {
 // update existing
} else {
  // create new
}

// save
```

marcomaddy1982: I guess here the name of the method is not descriptive, if you look inside it if there is the existing is updated there. So here we have two solution remade the method in something like `fetchOrCreate` or we remove the update part in the method and we apply your suggestion. Feel free to pick the one you prefer.

johnxnguyen: I see [here](https://github.com/wireapp/wire-ios-data-model/pull/999/files#diff-91e751262e217fd851b1051141016ac5R54) that we return the existing flag without updating it.

marcomaddy1982: sorry I just forgot to push the changes!!!

marcomaddy1982: Done now!!!

Tests/Source/Synchronization/Strategies/FeatureFlagRequestStrategyTests.swift
```diff
@@ -0,0 +1,154 @@
+//
```
typfel, line 1: Wrong licence header

Source/Synchronization/Strategies/FeatureFlagRequestStrategy.swift
```diff
@@ -0,0 +1,171 @@
...
+    public init(withManagedObjectContext managedObjectContext: NSManagedObjectContext,
+                applicationStatus: ApplicationStatus,
+                syncStatus: SyncStatus) {
+        syncContext = managedObjectContext
+        self.syncStatus = syncStatus
+        
+        super.init(withManagedObjectContext: managedObjectContext,
+                   applicationStatus: applicationStatus)
+        
+        self.configuration = [.allowsRequestsDuringSync]
```
typfel, line 51: We should probably add `.duringEventProcessing` since the 24 hours re-fetch will not happen during a slow sync. 
Would also like to have a test for that we generate a request when a feature flag was updated more than 24 hours ago.

marcomaddy1982: Done!!

PR #1259, opened 2020-07-06
[Feature]: Avoid Digital Signature Multiple Requests

marcomaddy1982:
The PR try to solve the following issue: 
If the user tap multiple time in a super short interval on the "Complete Signing" button before we close the OTP screen the request for the digital signature is made multiple times, this cause we don't have control to enable end disable the button (Swisscom web page). 

## Dependencies:

data-model: https://github.com/wireapp/wire-ios-data-model/pull/1001/


typfel:
So if understand correctly the root cause if this issue comes from the fact that we end up calling `retrieveSignature()` multiple times. The reason why we call `retrieveSignature()` is because we started to treat `sas-error-invalid-transaction-token ` as a success scenario.

Some thoughts:

- Can't we make `retrieveSignature()` more resilient by only making switch to the `.waitingForSignature` if it was previously in the `.waitingForCodeVerification` state. That would make the second call to `retrieveSignature()` do nothing.
- To me it seems we have not correctly understood the Swisscom responses. 
- Maybe we should only allow the `DigitalSignatureVerificationViewController` completion handler to fire once

marcomaddy1982:
> So if I understood correctly the root cause if this issue comes from the fact that we end up calling `retrieveSignature()` multiple times. The reason why we call `retrieveSignature()` is because we started to treat `sas-error-invalid-transaction-token ` as a success scenario.

Actually a sort of: I mean when we got the 'sas-error-invalid-transaction-token' the cms is generated by Swisscom and the request from our BE to Swisscom succeed -> we should post in the chat the digital signature cms file

> Can't we make retrieveSignature() more resilient by only making switch to the .waitingForSignature if it was previously in the .waitingForCodeVerification state. That would make the second call to retrieveSignature() do nothing.

I guess it is a good idea. I can try.

> Maybe we should only allow the DigitalSignatureVerificationViewController completion handler to fire once
I don't know if it makes sense cause if the request is not performed cause the solution you proposed works fine it would not be necessary this check.

> To me it seems we have not correctly understood the Swisscom responses.
Actually we didn't know we could have the  'sas-error-invalid-transaction-token' value, in the documentation it was not specified.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -85,6 +87,12 @@ extension SignatureRequestStrategy: ZMSingleRequestTranscoder {
         case requestSync:
             return makeSignatureRequest()
         case retrieveSync:
+            guard
+                let signatureStatus = syncContext.signatureStatus,
+                case .waitingForSignature = signatureStatus.state
```
typfel, line 92: I think this condition is already checked in `nextRequestIfAllowed()`

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -152,11 +160,17 @@ extension SignatureRequestStrategy: ZMSingleRequestTranscoder {
...
-        guard let responseId = signatureResponse?.responseId else {
+        
+        guard
+            let responseID = signatureResponse?.responseID,
+            let signatureStatus = syncContext.signatureStatus
+        else {
             return nil
         }
+    
+        signatureStatus.updateLoadingSignatureState()
```
typfel, line 171: Switching to different state here means that we wouldn't retry the request if it fails. I think it would better to solve this problem in data model by modifying `retrieveSignature ` like this:
```
public func retrieveSignature() {
     guard case .waitingForCodeVerification = state else { return }
     state = .waitingForSignature
     RequestAvailableNotification.notifyNewRequestsAvailable(nil)
}
```

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -152,11 +154,17 @@ extension SignatureRequestStrategy: ZMSingleRequestTranscoder {
...
```
typfel, line 165: I would remove `signatureStatus.updateLoadingSignatureState()` since because it prevents us from retrying as I mentioned above. Given the change to `retrieveSignature` I don't think the `.loadingSignature` state is necessary.

Source/Synchronization/Strategies/SignatureRequestStrategy.swift
```diff
@@ -152,11 +152,15 @@ extension SignatureRequestStrategy: ZMSingleRequestTranscoder {
     }
     
     private func makeRetrieveSignatureRequest() -> ZMTransportRequest? {
-        guard let responseId = signatureResponse?.responseId else {
+        
+        guard
+            let responseID = signatureResponse?.responseID,
+            let signatureStatus = syncContext.signatureStatus
```
typfel, line 158: Not used anymore.

PR #1263, opened 2020-07-13
[Refactor] Make the ZMLocalNotification public

KaterinaWire:
## What's new in this PR?

To be able to use the `ZMLocalNotification` in the UI project, this class must be public.


KaterinaWire:
> Why do we need to use this class in the UI project?

It's a temporary decision. I need to convert `ZMUpdateEvent` to the `ZMLocalNotification` object. The `ZMLocalNotification` class belongs to SE and I can't convert events to the  [ZMLocalNotification] in the Notification Engine framework. Also, I can't move `ZMLocalNotification` to the Request strategy because it has some dependencies with other classes.

PR #1264, opened 2020-07-14
Fix: app crash during registration, UserChangeInfo.add(observer:)

billypchan:
### Issues

The app sometimes crashes during the registration process.

### Causes

`userSession.managedObjectContext` would be nil when login is not yet done.

### Solutions

Mark`userSession.managedObjectContext` as optional. Add a guard for this case in `UserChangeInfo.add(observer:)`

### DIFF
--- a/Source/Data Model/UserChangeInfo+UserSession.swift
+++ b/Source/Data Model/UserChangeInfo+UserSession.swift
@@ -29,7 +29,9 @@ extension UserChangeInfo {
     public static func add(observer: ZMUserObserver,
                            for user: UserType,
                            in userSession: ZMUserSession) -> NSObjectProtocol? {
-        guard let managedObjectContext = userSession.managedObjectContext else { return nil }
+        guard let managedObjectContext = userSession.managedObjectContext else {
+            return nil            
+        }
         
         return add(observer: observer, for: user, in: managedObjectContext)
     }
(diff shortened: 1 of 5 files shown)


typfel:
I'm trying to understand what's happening here. The for producing this crash was:

1. create a new acct with UK location 
2. advance to fill in phone number step

This would cause crash because we would fire a UserChangeInfo before the UserSession has initialised its managedObjectContext?

billypchan:
@typfel I guess the location for the new user does not matter for the crash and I can not reproduce it. Review the code I found that it is caused by accessing a nilable `userSession.managedObjectContext` during registration.

typfel:
> @typfel I guess the location for the new user does not matter for the crash and I can not reproduce it. Review the code I found that it is caused by accessing a nilable `userSession.managedObjectContext` during registration.

Too bad that we can't reproduce it because now we are trading a crash for  for some undefined behaviour.  

I agree that we shouldn't hide the fact that `managedObjectContext` can be nil but then we should do same change for `syncMangedObjectContext`.

typfel:
Looking at the stack trace of the crash, I can see that it crashes inside a `UserCell` inside `TopPeopleLineCollectionViewController`. Since this comes from the start UI maybe a reasonable scenario is this:

1. User is logged into account A
2. Opens the startUI
3. Gets logged out / tries to add a new account from settings
4. During login the app crashes

I would suggest to verify that we are not leaking the `TopPeopleLineCollectionViewController` when tearing down the UI on logout.

billypchan:
@typfel would we merge this PR first, then I debug `TopPeopleLineCollectionViewController` for another possible reason?

billypchan:
close this PR until we record this crash again

PR #1266, opened 2020-07-17
[Feature] Encryption keys management

typfel:
# What's new in this PR?

- Create/Delete encryption keys when modifying `encryptMessagesAtRest`.
- Support fetching the encryption keys by providing an authenticated `LAContext`.
- Store fetched encryption keys in memory until the app goes into background.
- Allow the UI to observe when the database gets locked/unlocked.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1004

## Notes

I still don't know where's the best place is to store the encryption keys, but I need to place them somewhere so for now they live in the `SyncStatus`.


Source/UserSession/SyncStatus.swift
```diff
@@ -114,6 +114,16 @@ extension Notification.Name {
         return currentSyncPhase.isSyncing
     }
     
+    var encryptionKeys: EncryptionKeys? { // TODO jacob move somewhere more appropriate
+        didSet {
+            if encryptionKeys == nil {
+                DatabaseEncryptionLockNotification(databaseIsEncrypted: true).post(in: managedObjectContext.notificationContext)
+            } else {
+                DatabaseEncryptionLockNotification(databaseIsEncrypted: false).post(in: managedObjectContext.notificationContext)
+            }
```
johnxnguyen, line 123: You could remove this duplication by inlining the condition:

```swift
DatabaseEncryptionLockNotification(databaseIsEncrypted: encryptionKeys == nil).post(...)
```

or storing it a value

```swift
let databaseIsEncrypted = encryptionKeys == nil
DatabaseEncryptionLockNotification(databaseIsEncrypted: databaseIsEncrypted).post(...)
```

Source/UserSession/ZMUserSession+EncryptionAtRest.swift
```diff
@@ -17,19 +17,60 @@
...
         get {
             return managedObjectContext.encryptMessagesAtRest
         }
     }
     
+    public var isDatabaseLocked: Bool {
+        managedObjectContext.encryptMessagesAtRest && applicationStatusDirectory?.syncStatus.encryptionKeys == nil
+    }
+        
+    public func onDatabaseLockedChange(_ change: @escaping (_ isDatabaseLocked: Bool) -> Void) -> Any {
```
johnxnguyen, line 53: I suggest renaming this method to something like `registerDatabaseLockedChangeHandler`, emphasizing that the method registers a handler and is not the handler itself.

PR #1267, opened 2020-07-20
[Fix] make conference calling setting property mutable

David-Henner:
## What's new in this PR?

- Make the `useConferenceCalling` property mutable to allow UI to toggle conference calling


David-Henner:
@zenkins release minor

PR #1269, opened 2020-07-20
[Feature] Targeted calling messages

johnxnguyen:
## What's new in this PR?

This PR includes changes to send targeted calling messages when requested by AVS. Only some messages should be targeted, in which case AVS will invoke the `wcall_send_h` handler with a list of target clients passed in through the `targets` argument. If `targets` is `null`, then the message is not targeted and all clients participating in the conversation will receive the message.

**JIRA:** https://wearezeta.atlassian.net/browse/ZIOS-13424

### DIFF
--- a/Source/Calling/AVSWrapper.swift
+++ b/Source/Calling/AVSWrapper.swift
@@ -262,7 +262,7 @@ public class AVSWrapper: AVSWrapperType {
         }
     }
 
-    private let sendCallMessageHandler: Handler.CallMessageSend = { token, conversationId, senderUserId, senderClientId, targetsRef, _, data, dataLength, _, contextRef in
+    private let sendCallMessageHandler: Handler.CallMessageSend = { token, conversationId, senderUserId, senderClientId, targetsCString, _, data, dataLength, _, contextRef in
         guard let token = token else {
             return EINVAL
         }
@@ -270,7 +270,7 @@ public class AVSWrapper: AVSWrapperType {
         let bytes = UnsafeBufferPointer<UInt8>(start: data, count: dataLength)
         let transformedData = Data(buffer: bytes)
 
-        let targets = targetsRef
+        let targets = targetsCString
             .flatMap { String(cString: $0)?.data(using: .utf8) }
             .flatMap { AVSClientList($0) }
 
--- a/Source/Calling/AVSWrapper.swift
+++ b/Source/Calling/AVSWrapper.swift
@@ -262,7 +262,7 @@ public class AVSWrapper: AVSWrapperType {
         }
     }
 
-    private let sendCallMessageHandler: Handler.CallMessageSend = { token, conversationId, senderUserId, senderClientId, _, _, data, dataLength, _, contextRef in
+    private let sendCallMessageHandler: Handler.CallMessageSend = { token, conversationId, senderUserId, senderClientId, targetsRef, _, data, dataLength, _, contextRef in
         guard let token = token else {
             return EINVAL
         }
@@ -270,8 +270,13 @@ public class AVSWrapper: AVSWrapperType {
         let bytes = UnsafeBufferPointer<UInt8>(start: data, count: dataLength)
         let transformedData = Data(buffer: bytes)
 
+        let targets = targetsRef
+            .flatMap { String(cString: $0)?.data(using: .utf8) }
+            .flatMap { AVSClientList($0) }
+
+
         return AVSWrapper.withCallCenter(contextRef, conversationId, senderUserId, senderClientId) {
-            $0.handleCallMessageRequest(token: token, conversationId: $1, senderUserId: $2, senderClientId: $3, data: transformedData)
+            $0.handleCallMessageRequest(token: token, conversationId: $1, senderUserId: $2, senderClientId: $3, targets: targets, data: transformedData)
         }
     }
     
--- a/Source/Calling/CallCenterSupport.swift
+++ b/Source/Calling/CallCenterSupport.swift
@@ -66,7 +66,7 @@ public typealias CallConfigRequestCompletion = (String?, Int) -> Void
 
 public protocol WireCallCenterTransport: class {
 
-    /// Sends a calling OTR message.
+    /// Sends a calling message.
     ///
     /// - Parameters:
     ///     - data: The message payload.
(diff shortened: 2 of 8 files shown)


johnxnguyen:
@franziskuskiefer this is the Sync Engine side to targeted calling messages. 

On [this line](https://github.com/wireapp/wire-ios-sync-engine/pull/1269/files#diff-f7da177ad1b932ea4071d2ced8b83735R289) we cross over to the [request strategy framework](https://github.com/wireapp/wire-ios-request-strategy/blob/8ae49418c56967132783c3651a3718b3ea7f05da/Sources/Request%20Strategies/Generic%20Message/GenericMessageRequestStrategy.swift#L121). Then we [encrypt for transport](https://github.com/wireapp/wire-ios-request-strategy/blob/8ae49418c56967132783c3651a3718b3ea7f05da/Sources/Request%20Strategies/Generic%20Message/GenericMessageRequestStrategy.swift#L75). This will take us to the [data model framework](https://github.com/wireapp/wire-ios-data-model/blob/d3bfff512783b690a02fdf3a5cfb1c2c05fbc800/Source/Model/Message/ZMClientMessage%2BEncryption.swift#L136), where the encrypted payloads for each recipient are generated.

Let me know if you have any questions.

Source/Calling/AVSWrapper.swift
```diff
@@ -262,16 +262,21 @@ public class AVSWrapper: AVSWrapperType {
         }
     }
 
-    private let sendCallMessageHandler: Handler.CallMessageSend = { token, conversationId, senderUserId, senderClientId, _, _, data, dataLength, _, contextRef in
+    private let sendCallMessageHandler: Handler.CallMessageSend = { token, conversationId, senderUserId, senderClientId, targetsRef, _, data, dataLength, _, contextRef in
```
typfel, line 265: Is `targetsRef` really a reference? Maybe it's more suitable to call it `targetsCString`/`targetsRaw` or similar.

Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -360,6 +362,15 @@ extension CallingRequestStrategy: WireCallCenterTransport {
         }
 
     }
+
+    private func recipients(for targets: [AVSClient], in managedObjectContext: NSManagedObjectContext) -> GenericMessageEntity.Recipients {
+        let clientsByUser = targets
+            .compactMap { UserClient.fetchExistingUserClient(with: $0.clientId, in: managedObjectContext) }
+            .partition(by: \.user)
```
typfel, line 369: Nice

Source/Calling/AVSWrapper.swift
johnxnguyen: By reference I was thinking a pointer (which it is), but indeed `targetsCString` would be clearer. How did you understand reference to be?

typfel: Yes, I was also understanding reference as a C-style pointer.

johnxnguyen: I renamed it to `targetsCString` anyway.

Source/Calling/CallCenterSupport.swift
```diff
@@ -19,53 +19,85 @@
...
 
-/**
- * An object that can perform requests on behalf of the call center.
- */
+/// An object that can perform requests on behalf of the call center.
 
 public protocol WireCallCenterTransport: class {
-    func send(data: Data, conversationId: UUID, userId: UUID, completionHandler: @escaping ((_ status: Int) -> Void))
+
+    /// Sends a calling OTR message.
```
franziskuskiefer, line 69: We should really get rid of these OTR references. There's no OTR anywhere near this 😬 (Nothing to fix here though.)

Source/Calling/CallCenterSupport.swift
```diff
@@ -19,53 +19,85 @@
...
+    ///
+    /// - Parameters:
+    ///     - data: The message payload.
+    ///     - conversationId: The conversation in which the message is sent.
+    ///     - targets: Exact recipients of the message. If `nil`, all conversation participants are recipients.
+    ///     - completionHandler: A handler when the network request completes with http status code.
+
+    func send(data: Data, conversationId: UUID, targets: [AVSClient]?, completionHandler: @escaping ((_ status: Int) -> Void))
+
+    /// Send an calling message to the SFT server (for conference calling).
```
franziskuskiefer, line 79: "Send a"

Tests/Source/Synchronization/Strategies/CallingRequestStrategyTests.swift
```diff
@@ -193,4 +199,160 @@ class CallingRequestStrategyTests : MessagingTest {
...
+
+        syncMOC.performGroupedBlock {
+            nextRequest = self.sut.nextRequest()
+        }
+
+        XCTAssertTrue(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
+
+        guard let request = nextRequest else { return XCTFail("Expected next request") }
+
+        // Then we told backend to ignore missing clients (the non targeted conversation participants)
```
franziskuskiefer, line 252: "Then we tell"?

Tests/Source/Synchronization/Strategies/CallingRequestStrategyTests.swift
```diff
@@ -193,4 +199,160 @@ class CallingRequestStrategyTests : MessagingTest {
...
+        XCTAssertEqual(request.path, "/conversations/\(conversation.remoteIdentifier!.transportString())/otr/messages?ignore_missing")
+
+        guard
+            let data = request.binaryData,
+            let otrMessage = try? NewOtrMessage(serializedData: data)
+        else {
+            return XCTFail("Expected OTR message")
+        }
+
+        // Then we only sent the otr message to the targeted clients
```
franziskuskiefer, line 262: "send"?

johnxnguyen:
> I didn't check the underlying functions (we really have to unify these repositories). But what's happening in here looks good to me.
> (Inline comments are only nits.)

@franziskuskiefer I fixed the comments. Was thinking since we were making assertions about things that happened previously that the past tense was appropriate, but reading it out aloud it sounded weird aha. 

Also we are trying to move to a mono repository, but that project got put on hold.

johnxnguyen:
@zenkins release major

PR #1270, opened 2020-07-22
[Feature] Generate push notifications from update events

KaterinaWire:
## What's new in this PR?

Add a new helpers to the` ZMLocalNotification+Events` that create message local notifications from the update event.


Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification+Events.swift
```diff
@@ -269,3 +279,98 @@ private class NewUserEventNotificationBuilder: EventNotificationBuilder {
...
+        // we don't want to create notifications when other people join or leave conversation
+        guard let dataPayload = (event.payload as NSDictionary).dictionary(forKey: "data"),
+            let userIds = dataPayload["user_ids"] as? [String] else {
+                return false
+        }
+        let users = userIds.compactMap({ UUID.init(uuidString: $0)}).compactMap({ZMUser(remoteID: $0, createIfNeeded: true, in: moc)})
+        let forSelf = users.count == 1 && users.first!.isSelfUser
+        if !forSelf && event.type != .conversationMessageTimerUpdate {
+            return false
+        }
```
typfel, line 372: Since this logic only applies to `.conversationMemberJoin` and `.conversationMemberLeave` I would restrict it to that otherwise we might have a bug the next time we add new system message case and forget to update this logic.

like this:

```
switch  contentType 
case .participantsAdded, .participantsRemoved:
   // check that if self was added or removed
default:
  super.shouldCreateNotification()
}
```

Source/Notifications/Push notifications/Notification Types/LocalNotificationContentType.swift
```diff
@@ -103,13 +104,96 @@ public enum LocalNotificationContentType : Equatable {
...
+            let timeoutIntegerValue = (payload["message_timer"] as? Int64) ?? 0
+            
+            let value = MessageDestructionTimeoutValue(rawValue: TimeInterval(timeoutIntegerValue / 1000))
+            if value == .none {
+                return .messageTimerUpdate(nil)
+            } else {
+                return .messageTimerUpdate(value.displayString)
+            }
+        default:
+            guard let message = GenericMessage(from: event) else {
```
typfel, line 127: This is only valid when event.type is `.conversationOtrMessageAdd `. I would add this a separate `case` and default can then return nil.

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests.swift
```diff
@@ -160,4 +160,66 @@ class ZMLocalNotificationTests: MessagingTest {
...
+                       ]]
+               ],
+               "type": "conversation.member-join"
+               ]
+           return ZMUpdateEvent(fromEventStreamPayload: payload as ZMTransportData, uuid: nonce)!
+       }
+
+    func createMemberLeaveUpdateEvent(_ nonce: UUID, conversationID: UUID, senderID: UUID = UUID.create()) -> ZMUpdateEvent {
+           
+//           let user2ID = UUID.create()
```
typfel, line 199: Commented out code

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests.swift
```diff
@@ -160,4 +160,66 @@ class ZMLocalNotificationTests: MessagingTest {
...
+               "time": NSDate().transportString(),
+               "data": [
+                "user_ids": [selfUser.remoteIdentifier.transportString()],
+               ],
+               "type": "conversation.member-leave"
+               ]
+           return ZMUpdateEvent(fromEventStreamPayload: payload as ZMTransportData, uuid: nonce)!
+       }
+    
+    func createMessageTimerUpdateEvent(_ nonce: UUID, conversationID: UUID, senderID: UUID = UUID.create()) -> ZMUpdateEvent {
```
typfel, line 213: Forgot to add a test where we use this helper method?

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests.swift
```diff
@@ -160,4 +160,66 @@ class ZMLocalNotificationTests: MessagingTest {
...
+            "from": senderID.transportString(),
+            "time": Date().transportString(),
+            "data": ["text": try? genericMessage.serializedData().base64String()],
+            "type": "conversation.otr-message-add"
+        ]
+        
+        return ZMUpdateEvent(fromEventStreamPayload: payload as ZMTransportData, uuid: nonce)!
+    }
+    
+    func createMemberJoinUpdateEvent(_ nonce: UUID, conversationID: UUID, senderID: UUID = UUID.create()) -> ZMUpdateEvent {
```
typfel, line 177: Forgot to add a test where we use this helper method?

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests_Event.swift
```diff
@@ -514,4 +501,39 @@ final class ZMLocalNotificationTests_Event: ZMLocalNotificationTests {
         XCTAssertNotNil(note)
         XCTAssertEqual(note!.title, "Super User")
     }
+    
+    // MARK: - Create text local notification from update events
+    
+    func testThatItCreatesATextNotification() {
```
typfel, line 507: I would expect more tests here.

typfel:
We should also remove the old code that this is replacing, which creates ZMLocalNotifications from a `ZMMessage`. Maybe it's also possible to adapt/re-use the existing tests we have for that.

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests_Event.swift
typfel: Maybe the old tests can be re-used/adapted?

Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification+Events.swift
KaterinaWire: Fixed

Source/Notifications/Push notifications/Notification Types/LocalNotificationContentType.swift
KaterinaWire: Fixed

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests.swift
KaterinaWire: Removed

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests.swift
KaterinaWire: Added

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests.swift
KaterinaWire: Added

Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification+Events.swift
```diff
@@ -269,3 +279,101 @@ private class NewUserEventNotificationBuilder: EventNotificationBuilder {
...
+    }
+    
+    override func shouldCreateNotification() -> Bool {
+        // we don't want to create notifications when other people join or leave conversation
+        guard let dataPayload = (event.payload as NSDictionary).dictionary(forKey: "data"),
+            let userIds = dataPayload["user_ids"] as? [String] else {
+                return false
+        }
+        let users = userIds.compactMap({ UUID.init(uuidString: $0)}).compactMap({ZMUser(remoteID: $0, createIfNeeded: true, in: moc)})
+        let forSelf = users.count == 1 && users.first!.isSelfUser
```
typfel, line 369: Let's move this to an extension on the `ZMUpdateEvent`?

```
extension ZMUpdateEvent {
   var userIDs: [UUID]
}
```

and then we could do `event.usersIDs.contains(ZMUser.selfUser(in: moc).remoteIdentifier)`. This way we wouldn't create user objects just to check if the self user is present.

Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification+Events.swift
```diff
@@ -269,3 +279,101 @@ private class NewUserEventNotificationBuilder: EventNotificationBuilder {
...
+
+        switch contentType {
+        case .participantsAdded, .participantsRemoved:
+            return forSelf ? super.shouldCreateNotification() : false
+        default:
+            return super.shouldCreateNotification()
+        }
```
typfel, line 376: I now realise that my suggestions was not maybe so great since it complicates the flow to call super. What about something like this?

```suggestion
        switch contentType {
        case .participantsAdded, .participantsRemoved where concernsSelfUser == false
            return false
        default:
            break
        }
        
        return super.shouldCreateNotification()
```

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests_Event.swift
KaterinaWire: Done

Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification+Events.swift
KaterinaWire: Done

KaterinaWire:
> We should also remove the old code that this is replacing, which creates ZMLocalNotifications from a `ZMMessage`. Maybe it's also possible to adapt/re-use the existing tests we have for that.

Done

Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification+Events.swift
KaterinaWire: Done

Source/Notifications/Push notifications/Notification Types/LocalNotificationContentType.swift
```diff
@@ -103,13 +104,97 @@ public enum LocalNotificationContentType : Equatable {
...
+                return .fileUpload
+            }
+        default:
+           return .undefined
+        }
+    }
+    
 }
 
 public func ==(rhs: LocalNotificationContentType, lhs: LocalNotificationContentType) -> Bool {
```
typfel, line 193: If `LocalNotificationContentType` only has primitive associated types we could probably just annotate it with `Equatable` and remove this function.

Source/Synchronization/Transcoders/ZMConversationTranscoder.swift
```diff
@@ -46,8 +46,8 @@ extension ZMConversationTranscoder {
     
     @objc(appendSystemMessageForUpdateEvent:inConversation:)
     public func appendSystemMessage(for event: ZMUpdateEvent, conversation: ZMConversation) {
-        if let systemMessage = ZMSystemMessage.createOrUpdate(from: event, in: self.managedObjectContext) {
-            self.localNotificationDispatcher.process(systemMessage)
+        if let _ = ZMSystemMessage.createOrUpdate(from: event, in: self.managedObjectContext) {
+            self.localNotificationDispatcher.process(event)
```
typfel, line 50: This forwarding to the `localNotificationDispatcher` is already happening here https://github.com/wireapp/wire-ios-sync-engine/blob/b9cddd15fe16b68383d287f80f2c5b4d54c6c5e6/Source/Synchronization/ZMSyncStrategy%2BEventProcessing.swift#L57

Source/Synchronization/Transcoders/ZMConversationTranscoder.swift
```diff
@@ -242,8 +242,8 @@ extension ZMConversationTranscoder {
         if let timestamp = event.timeStamp(), let conversation = conversation {
             // system message should reflect the synced timer value, not local
             let timer = conversation.hasSyncedDestructionTimeout ? conversation.messageDestructionTimeoutValue : 0
-            let message = conversation.appendMessageTimerUpdateMessage(fromUser: user, timer: timer, timestamp: timestamp)
-            localNotificationDispatcher.process(message)
+            let _ = conversation.appendMessageTimerUpdateMessage(fromUser: user, timer: timer, timestamp: timestamp)
+            localNotificationDispatcher.process(event)
```
typfel, line 246: Same as above.

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests_Message.swift
```diff
@@ -179,15 +182,14 @@ class ZMLocalNotificationTests_Message : ZMLocalNotificationTests {
...
         
         // given
-        let message = oneOnOneConversation.append(text: "Hello Hello!") as! ZMOTRMessage
-        message.serverTimestamp = Date.distantFuture
-        message.sender = sender
+        let event = createUpdateEvent(UUID.create(), conversationID: oneOnOneConversation.remoteIdentifier!, genericMessage: GenericMessage(content: Text(content: "Hello Hello!"), nonce: UUID.create()), senderID: sender.remoteIdentifier)
         
-        let note = ZMLocalNotification(message: message)!
+        let note = ZMLocalNotification(event: event, conversation: oneOnOneConversation, managedObjectContext: syncMOC)
+        //(message: message)!
```
typfel, line 188: Comment

Source/Notifications/Push notifications/LocalNotificationDispatcher+Messages.swift
```diff
@@ -23,35 +23,53 @@ extension LocalNotificationDispatcher: PushMessageHandler {
 
     // Processes ZMOTRMessages and ZMSystemMessages
     @objc(processMessage:) public func process(_ message: ZMMessage) {
```
typfel, line 25: We shouldn't we delete this method completely?

Source/Notifications/Push notifications/LocalNotificationDispatcher+Messages.swift
```diff
@@ -23,35 +23,53 @@ extension LocalNotificationDispatcher: PushMessageHandler {
...
+        //                else if let message = message as? ZMSystemMessage {
+        //                    note = ZMLocalNotification(systemMessage: message)
+        //                }
+        //
+        //                note.apply(scheduleLocalNotification)
+        //                note.apply(messageNotifications.addObject)
     }
     
     // Process ZMGenericMessage that have "invisible" as in they don't create a message themselves
     @objc(processEvent:) public func process(_ event: ZMUpdateEvent) {
```
typfel, line 47: This could be move/integrated into `ZMLocalNotification+Events.swift`?

Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification+Messages.swift
```diff
@@ -17,133 +17,6 @@
...
-        }
-        
-    }
-
-}
-
-
 // MARK: - Failed Messages
 
 extension ZMLocalNotification {
```
typfel, line 22: How about renaming this file to `ZMLocalNotification+ExpiredMessages` since that's its sole purpose now.

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests_SystemMessage.swift
```diff
@@ -24,38 +24,36 @@ class ZMLocalNotificationTests_SystemMessage : ZMLocalNotificationTests {
...
-        systemMessage.sender = aSender
-        systemMessage.visibleInConversation = conversation
-        return ZMLocalNotification(systemMessage: systemMessage)
+        let event = createMemberLeaveUpdateEvent(UUID.create(), conversationID: conversation.remoteIdentifier!, users: Array(otherUsers), senderID: aSender.remoteIdentifier)
+        
+        return ZMLocalNotification(event: event, conversation: conversation, managedObjectContext: syncMOC)
     }
     
-    // MARK: - Tests
+     // MARK: - Tests
```
typfel, line 38: Extra space

Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification+Events.swift
```diff
@@ -269,3 +277,101 @@ private class NewUserEventNotificationBuilder: EventNotificationBuilder {
...
+        return notificationType.messageBodyText(sender: sender, conversation: conversation).trimmingCharacters(in: .whitespaces)
+    }
+    
+    override var notificationType: LocalNotificationType {
+        return LocalNotificationType.message(contentType)
+    }
+    
+    override func shouldCreateNotification() -> Bool {
+        // we don't want to create notifications when other people join or leave conversation
+        if event.userIDs.isEmpty {
```
typfel, line 364: Why is this if statement necessary?

KaterinaWire: For the `.conversationMessageTimerUpdate` type of event. This type of event does not contain userIDs

KaterinaWire: But then we will skip to the default section and will call `super.shouldCreateNotification()`. So we don't really need this `if statement`.

typfel: Right but `.conversationMessageTimerUpdate` ends up in the `default` case which breaks and then it goes on to `return super.shouldCreateNotification()`

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests_Message.swift
KaterinaWire: Fixed

Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification+Messages.swift
KaterinaWire: Done

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests_SystemMessage.swift
KaterinaWire: Fixed

Source/Notifications/Push notifications/Notification Types/LocalNotificationContentType.swift
KaterinaWire: Removed

Source/Notifications/Push notifications/LocalNotificationDispatcher+Messages.swift
KaterinaWire: Removed

Source/Notifications/Push notifications/LocalNotificationDispatcher+Messages.swift
KaterinaWire: What do you mean?

typfel: Nevermind was confused and didn't realise that this works on `LocalNotificationDispatcher` and not on `ZMLocalNotification`.

PR #1271, opened 2020-07-22
[Fix] provide public interface to set conference calling property

David-Henner:
## What's new in this PR?

UI needs a way to set the `useConferenceCalling` property on `WireCallCenterV3`.
This is done similarly to the way `useConstantBitRate` is set.

### DIFF
--- a/Source/Data Model/NSManagedObjectContext+CTCallCenter.swift
+++ b/Source/Data Model/NSManagedObjectContext+CTCallCenter.swift
@@ -55,4 +55,19 @@ public extension NSManagedObjectContext {
         
     }
     
+    private static let ConferenceCallingKey = "ConferenceCallingKey"
+    
+    var zm_useConferenceCalling : Bool {
+        
+        get {
+            precondition(zm_isUserInterfaceContext, "zm_useConferenceCalling can only be accessed on the ui context")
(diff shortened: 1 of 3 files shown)


Source/Data Model/NSManagedObjectContext+CTCallCenter.swift
```diff
@@ -55,4 +55,19 @@ public extension NSManagedObjectContext {
...
+    var zm_useConferenceCalling : Bool {
+        
+        get {
+            precondition(zm_isUserInterfaceContext, "zm_useConferenceCalling can only be accessed on the ui context")
+            return userInfo[NSManagedObjectContext.ConferenceCallingKey] as? Bool ?? false
+        }
+        
+        set {
+            precondition(zm_isUserInterfaceContext, "zm_useConferenceCalling can only be accessed on the ui context")
+            userInfo[NSManagedObjectContext.ConferenceCallingKey] = newValue
```
typfel, line 69: Note that `NSManagedObjectContext.UserInfo` is not persisted.

David-Henner: That's right, we don't need to persist that property on the sync engine

David-Henner:
@zenkins release minor

PR #1272, opened 2020-07-24
[Feature] Only process update events when the app is in the foreground

typfel:
## What's new in this PR?

When we enable EAR we can only process update events when the application is running in the foreground because only then will we have access to the database key, which allows us to insert messages. Therefore I've changed it so that we only decrypt and store update events when running in the background, the event processing will happen when the app goes into foreground *or* when the database is unlocked.

In the future we might allow some events to processed background which does not need to insert messages, like conversation events.

## Notes

Sending delivery receipts is still something we need to do in background which will break now, since that's currently happening when events are processed. This will be implemented in a separate PR though.

The `ZMSyncStrategy` tests are in a bad state where the tests doesn't always reflect what's happening anymore after refactorings. For example there's a test named `testThatItAsksClientMessageTranscoderToDecryptUpdateEvents` even though the `ClientMEssageTranscoder` is not responsible for decrypting events. We should review the ZMSyncStrategy tests and remove the OCMock dependency to enable a Swift conversation.

### DIFF
--- a/Source/Synchronization/ZMUpdateEventsBuffer.m
+++ b/Source/Synchronization/ZMUpdateEventsBuffer.m
@@ -18,10 +18,11 @@
 
 @import WireTransport;
 #import "ZMUpdateEventsBuffer.h"
+#import <WireSyncEngine/WireSyncEngine-Swift.h>
 
 @interface ZMUpdateEventsBuffer ()
 
-@property (nonatomic, readonly, weak) id<ZMUpdateEventConsumer> consumer;
+@property (nonatomic, readonly, weak) id<UpdateEventProcessor> updateEventProcessor;
 @property (nonatomic, readonly) NSMutableArray *bufferedEvents;
 
 @end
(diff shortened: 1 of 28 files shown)


Source/UserSession/ZMUserSession.swift
```diff
@@ -413,7 +414,7 @@ extension ZMUserSession: ZMNetworkStateDelegate {
         }
     }
     
-    private func updateNetworkState() {
+    internal func updateNetworkState() {
```
billypchan, line 417:
```suggestion
    func updateNetworkState() {
```

Source/Synchronization/ZMUpdateEventsBuffer.m
```diff
@@ -48,25 +49,10 @@ - (void)addUpdateEvent:(ZMUpdateEvent *)event
 
 - (void)processAllEventsInBuffer
 {
-    [self.consumer consumeUpdateEvents:self.bufferedEvents];
+    [self.updateEventProcessor storeUpdateEvents:self.bufferedEvents ignoreBuffer:YES];
     [self.bufferedEvents removeAllObjects];
 }
 
-- (void)discardAllUpdateEvents
```
typfel, line 55: These methods were not used so I deleted them.

PR #1273, opened 2020-07-27
[Feature] Encrypt event updates

KaterinaWire:
## What's new in this PR?
- Add `encryptedPayload` property to the `StoredUpdateEvent` entity.
- Encrypt the payload and save it to the database.
- Decrypt store events with a private key


typfel:
> Do we need a database migration?

A light weight migration should be enough I think. The `encryptedPayload` field should be `nil` by default.

> Do we need to remove the payload property from the StoredUpdateEvent entity?

EAR is an option so will still support non-encrypted update events.

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -107,6 +107,8 @@ extension EventDecoder {
     /// - parameter events The new events that should be decrypted and stored in the database.
     /// - parameter startingAtIndex The startIndex to be used for the incrementing sortIndex of the stored events.
     fileprivate func storeEvents(_ events: [ZMUpdateEvent], startingAtIndex startIndex: Int64) {
+        let account = Account(userName: "", userIdentifier: ZMUser.selfUser(in: self.syncMOC).remoteIdentifier)
+        let encryptionkeys = try? EncryptionKeys.createKeys(for: account)
```
typfel, line 111: We should not create new encryption keys here. We should get the public key from from the static method ` publicKey(for account: Account)`

KaterinaWire:
> > Do we need a database migration?
> 
> A light weight migration should be enough I think. The `encryptedPayload` field should be `nil` by default.

I've made `encryptedPayload` as an Optional.

> > Do we need to remove the payload property from the StoredUpdateEvent entity?
> 
> EAR is an option so will still support non-encrypted update events.

Source/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -48,6 +49,32 @@ public final class StoredUpdateEvent: NSManagedObject {
...
     }
     
+    /// Maps a passed in `ZMUpdateEvent` to a `StoredUpdateEvent` which is persisted in a database
+    /// - Parameters:
+    ///   - event: received events
+    ///   - managedObjectContext: current managedObjectContext
+    ///   - index: the passed in `index` is used to enumerate events to be able to fetch and sort them later on in the order they were received
+    ///   - publicKey: the publicKey which will be used to encrypt update events
+    /// - Returns: storedEvent which will be persisted in a database
+    public static func encryptAndCreate(_ event: ZMUpdateEvent, managedObjectContext: NSManagedObjectContext, index: Int64, publicKey: SecKey?) -> StoredUpdateEvent? {
```
typfel, line 59: Shouldn't this replacing the existing `create` method?

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -96,17 +97,20 @@ extension EventDecoder {
...
     }
     
     /// Decrypts and stores the decrypted events as `StoreUpdateEvent` in the event database.
     /// The encryption context is only closed after the events have been stored, which ensures
     /// they can be decrypted again in case of a crash.
     /// - parameter events The new events that should be decrypted and stored in the database.
     /// - parameter startingAtIndex The startIndex to be used for the incrementing sortIndex of the stored events.
-    fileprivate func storeEvents(_ events: [ZMUpdateEvent], startingAtIndex startIndex: Int64) {
+    /// - parameter encryptionKeys  The keys to be used to encrypt update events
+    fileprivate func storeEvents(_ events: [ZMUpdateEvent], startingAtIndex startIndex: Int64, with encryptionKeys: EncryptionKeys?) {
```
typfel, line 112: This method can't take `EncryptionKeys` as a parameter since `EncryptionKeys` are only available in the foreground and we want to stop events in the background.

Source/Synchronization/Decoding/ZMEventModel.xcdatamodeld/ZMEventModel.xcdatamodel/contents
```diff
@@ -1,19 +1,15 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="11232" systemVersion="15G31" minimumToolsVersion="Xcode 7.0" sourceLanguage="Objective-C" userDefinedModelVersionIdentifier="">
```
typfel, line 2: I think you need to add new data model version instead of modifying the existing one, otherwise the migration won't work.

Tests/Source/Synchronization/StoreUpdateEventTests.swift
```diff
@@ -292,3 +304,123 @@ class StoreUpdateEventTests: MessagingTest {
         
     }
 }
+
+// MARK: - Encrypting / Decrypting events using public / private keys
+
+extension StoreUpdateEventTests {
+    func testThatItEncryptsEventIfThePublicKeyIsNotNill() throws {
```
typfel, line 311:
```suggestion
    func testThatItEncryptsEventIfThePublicKeyIsNotNil() throws {
```

typfel: Same typo exists in many places

Tests/Source/Synchronization/StoreUpdateEventTests.swift
```diff
@@ -292,3 +304,123 @@ class StoreUpdateEventTests: MessagingTest {
...
+            XCTAssertEqual(storedEvent.debugInformation, event.debugInformation)
+            XCTAssertNil(storedEvent.payload)
+            XCTAssertEqual(storedEvent.isTransient, event.isTransient)
+            XCTAssertEqual(storedEvent.source, Int16(event.source.rawValue))
+            XCTAssertEqual(storedEvent.sortIndex, 2)
+            XCTAssertEqual(storedEvent.uuidString, event.uuid?.transportString())
+            XCTAssertNotNil(storedEvent.encryptedPayload)
+            
+        // then
+            let convertedEvents = StoredUpdateEvent.eventsFromStoredEvents([storedEvent], encryptionKeys: encryptionKeys)
```
typfel, line 384: I would move the `when` here because `eventsFromStoredEvents()` is the method we are testing in the test. What happens before is just setup (`given`)

Tests/Source/Synchronization/StoreUpdateEventTests.swift
```diff
@@ -292,3 +304,123 @@ class StoreUpdateEventTests: MessagingTest {
...
+            XCTAssertEqual(storedEvent.debugInformation, event.debugInformation)
+            XCTAssertEqual(storedEvent.payload, event.payload as NSDictionary)
+            XCTAssertEqual(storedEvent.isTransient, event.isTransient)
+            XCTAssertEqual(storedEvent.source, Int16(event.source.rawValue))
+            XCTAssertEqual(storedEvent.sortIndex, 2)
+            XCTAssertEqual(storedEvent.uuidString, event.uuid?.transportString())
+            XCTAssertNil(storedEvent.encryptedPayload)
+            
+        // then
+            let convertedEvents = StoredUpdateEvent.eventsFromStoredEvents([storedEvent], encryptionKeys: nil)
```
typfel, line 415: Same as previous comment

Tests/Source/Synchronization/StoreUpdateEventTests.swift
```diff
@@ -292,3 +304,123 @@ class StoreUpdateEventTests: MessagingTest {
...
+            XCTAssertEqual(convertedEvents.first!.isTransient, event.isTransient)
+            XCTAssertEqual(convertedEvents.first!.source, event.source)
+            XCTAssertEqual(convertedEvents.first!.uuid?.transportString(), event.uuid?.transportString())
+            
+        } else {
+            XCTFail("Did not create storedEvent")
+        }
+    }
+    
+    func testThatItConvertsStoreEventToTheUpdateEventIfThePrivateKeyIsNill() throws {
```
typfel, line 396: How about a test for this scenario:
We try to convert an encrypted stored event without supplying a private key.

Source/Synchronization/Decoding/StoreUpdateEvent.swift
KaterinaWire: I removed `create` method and replaced it everywhere with `encryptAndCreate`

Source/Synchronization/Decoding/EventDecoder.swift
KaterinaWire: Yes, you are right. I changed it. Now this method does not expect `EncryptionKeys` as a parameter.

Tests/Source/Synchronization/StoreUpdateEventTests.swift
KaterinaWire: Fixed

Tests/Source/Synchronization/StoreUpdateEventTests.swift
KaterinaWire: Fixed

Tests/Source/Synchronization/StoreUpdateEventTests.swift
KaterinaWire: Added `testThatItCanNotConvertEncryptedStoredEventWithoutPrivateKey()`

Source/Synchronization/Decoding/ZMEventModel.xcdatamodeld/ZMEventModel.xcdatamodel/contents
KaterinaWire: Added new version

Source/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -107,6 +108,10 @@ extension EventDecoder {
     /// - parameter events The new events that should be decrypted and stored in the database.
     /// - parameter startingAtIndex The startIndex to be used for the incrementing sortIndex of the stored events.
     fileprivate func storeEvents(_ events: [ZMUpdateEvent], startingAtIndex startIndex: Int64) {
+        var publicKey: SecKey?
+        if let account = fetchSelectedAccount() {
```
typfel, line 112: We should not fetch currently selected account since we might operate on multiple user session in parallel. I'd say we should pass in the SecKey as a parameter or we might need to change `publicKey(for: account)` to maybe take a userID instead.

KaterinaWire: Yes, using the selected account is not correct. I decided to use `accountManager.accounts`and select the necessary account using `selfUser.remoteIdentifier.`

typfel:
Was looking at the scheme changes and noticed that both `payload` and `encryptedPayload` are of type Transformable. Maybe it would cleaner to add a `isEncrypted` boolean field instead of the `encryptedPayload`, because that would make it impossible to have an invalid state where both the encrypted and non-encrypted payload is available simultaneously.

Source/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -27,24 +27,38 @@ public final class StoredUpdateEvent: NSManagedObject {
...
         storedEvent.uuidString = event.uuid?.transportString()
+        guard let publicKey = publicKey,
+            let data = try? JSONSerialization.data(withJSONObject: event.payload, options: []) else {
+                storedEvent.payload = event.payload as NSDictionary
+                return storedEvent
+        }
+        storedEvent.encryptedPayload = SecKeyCreateEncryptedData(publicKey,
+                                                                 .eciesEncryptionCofactorX963SHA256AESGCM,
+                                                                 data as CFData,
+                                                                 nil)
```
johnxnguyen, line 61: As @typfel suggests, instead of creating a separate `encryptedPayload` property, you could continue to store the payload (whether it's encrypted or not) in the existing `payload` property. To differentiate between encrypted and non-encrypted payloads, you can add a new boolean property called `isEncrypted` and set the appropriate value.

For example:

```swift
if let publicKey = publicKey {
    storedEvent.payload = /* encrypt payload here */
    storedEvent.isEncrypted = true
} else {
    storedEvent.payload = event.payload
    storedEvent.isEncrypted = false
}

return storedEvent
```

Source/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -69,13 +83,26 @@ public final class StoredUpdateEvent: NSManagedObject {
...
+                let encryptedPayload = $0.encryptedPayload,
+                let decryptedData = SecKeyCreateDecryptedData(encryptionKeys.privateKey,
+                                                     .eciesEncryptionCofactorX963SHA256AESGCM,
+                                                     encryptedPayload,
+                                                     nil) {
+                decryptedPayload = try? JSONSerialization.jsonObject(with: decryptedData as Data, options: []) as? NSDictionary
+            }
+            guard let payload = decryptedPayload ?? $0.payload else {
+                    return nil
+            }
```
johnxnguyen, line 103: Likewise, here you can check a combination of the encryption keys and the `isEncrypted` property to determine how to proceed:

```swift
var payload = $0.payload

if $0.isEncrypted {
    guard let keys = encryptionKeys else {
        // We can't decrypt, handle this
    }

    payload = /* decrypt payload here */
}
```

KaterinaWire: I've made updates according to your suggestions.

Source/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -27,27 +27,47 @@ public final class StoredUpdateEvent: NSManagedObject {
...
         storedEvent.uuidString = event.uuid?.transportString()
+        storedEvent.payload = encryptIfNeeded(eventPayload: event.payload as NSDictionary, publicKey: publicKey)
+        storedEvent.isEncrypted = publicKey != nil
+        
         return storedEvent
     }
     
+    private static func encryptIfNeeded(eventPayload: NSDictionary, publicKey: SecKey?) -> NSDictionary {
+        guard let key = publicKey,
+            let data = try? JSONSerialization.data(withJSONObject: eventPayload, options: []),
```
johnxnguyen, line 61: I would expect that if serialization failed and we couldn't encrypt, then we shouldn't store the unencrypted payload. Is this correct?

Source/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -27,27 +27,47 @@ public final class StoredUpdateEvent: NSManagedObject {
...
+            let encryptedData = SecKeyCreateEncryptedData(key,
+                                                          .eciesEncryptionCofactorX963SHA256AESGCM,
+                                                          data as CFData,
+                                                          nil) else {
+                                                            return eventPayload
+        }
+        return NSDictionary(dictionary: ["encryptedPayload": encryptedData])
```
johnxnguyen, line 68: It would be a good to store this key as a private class variable or static variable, together with a comment about its purpose.

KaterinaWire: Done

Source/Synchronization/Decoding/StoreUpdateEvent.swift
KaterinaWire: Yes, you are right. We should store the unencrypted payload if there's no publicKey.

Source/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -24,6 +24,9 @@ public final class StoredUpdateEvent: NSManagedObject {
     
     static let entityName =  "StoredUpdateEvent"
     static let SortIndexKey = "sortIndex"
+    /// The key under which the event payload is encrypted by the public key.
+    static let encryptedPayloadKey = "encryptedPayload"
```
johnxnguyen, line 28: This can be private.

KaterinaWire: No, because I use it in tests.

KaterinaWire: Made this property as internal.

PR #1274, opened 2020-07-28
[Feature] Add isConferenceCall property to VoiceChannel

johnxnguyen:
## What's new in this PR?

When we receive an incoming call, AVS tells us what kind of call it is, whether it is a legacy group call or a conference call. This information is useful to the UI in that we can selectively remove some restrictions (such as video group limits) for conference calls only.

This PR uses this information to set the `isConferenceCall` property in the `CallProperties` protocol. This is then propagated to the `VoiceChannel` which is the main calling object used from the UI.

### DIFF
--- a/Source/Calling/AVSBridging.swift
+++ b/Source/Calling/AVSBridging.swift
@@ -166,4 +166,12 @@ extension AVSWrapper {
         return 0
     }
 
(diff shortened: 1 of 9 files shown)


johnxnguyen:
@zenkins release major

PR #1275, opened 2020-07-28
[Chore] Remove call center configuration

johnxnguyen:
## What's new in this PR?

This PR removes the `WireCallCenterConfiguration`, which was used to enable / disable the conference calling feature. This functionality is no longer needed since the feature flag is manipulated through the public beta toggle in UI.


Source/Calling/WireCallCenterV3.swift
```diff
@@ -370,7 +367,7 @@ extension WireCallCenterV3 {
         endAllCalls(exluding: conversationId)
         
         let callType: AVSCallType
-        if conversation.localParticipants.count > videoParticipantsLimit {
+        if !useConferenceCalling && conversation.localParticipants.count > legacyVideoParticipantsLimit {
```
David-Henner, line 370: could this condition be refactored into one method ?

johnxnguyen: Done

johnxnguyen:
@zenkins release major

PR #1276, opened 2020-07-29
[Feature] Integrate new delivery receipts strategy

typfel:
## What's new in this PR?

Integrate the `DeliveryReceiptsRequestStrategy` to re-enable delivery receipts sending.

Delete the `BackgroundAPNSConfirmationStatus` which purpose was to open a background activity while there  are messages which needs to be confirmed as delivered. This is redundant since we open a background activity for every request that we schedule.

Update after `ZMUpdateEvent` accessor methods were converted to properties.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/235
- [x] https://github.com/wireapp/wire-ios-data-model/pull/1006


typfel:
> Looks good but you have some failed tests.

Updated, also realised I could remove the `BackgroundAPNSConfirmationStatus`

PR #1277, opened 2020-08-06
[Feature]: Disable/Enable Cut, Save, Copy, Paste Flag

marcomaddy1982:
## What's new in this PR?

from the session_manager confirmation file it is not possible to enable and disable the cut, save, copy and paste action in the conversation

## Dependency

JIRA Ticket: https://wearezeta.atlassian.net/browse/ZIOS-13657
build-configuration: https://github.com/wireapp/wire-ios-build-configuration/pull/26

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -788,6 +788,10 @@ public final class SessionManager : NSObject, SessionManagerType {
     @objc public var isUserSessionActive: Bool {
         return activeUserSession != nil
     }
+    
+    public var isCopyAndPasteEnable: Bool {
+        return configuration.enableCopyAndPaste
+    }
 
     func updateProfileImage(imageData: Data) {
         activeUserSession?.enqueue {
@@ -820,7 +824,7 @@ public final class SessionManager : NSObject, SessionManagerType {
         }
     }
     
-    public var useConstantBitRateAudio : Bool = false {
+    public var useConstantBitRateAudio: Bool = false {
         didSet {
             activeUserSession?.useConstantBitRateAudio = useConstantBitRateAudio
         }
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -789,8 +789,8 @@ public final class SessionManager : NSObject, SessionManagerType {
         return activeUserSession != nil
     }
     
-    public var isCopyAndPasteEnable: Bool {
-        return configuration.enableCopyAndPaste
+    public var isDisabledClipboard: Bool {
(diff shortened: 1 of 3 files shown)


Source/SessionManager/SessionManager.swift
```diff
@@ -788,6 +788,10 @@ public final class SessionManager : NSObject, SessionManagerType {
     @objc public var isUserSessionActive: Bool {
         return activeUserSession != nil
     }
+    
+    public var isCopyAndPasteEnable: Bool {
+        return configuration.enableCopyAndPaste
+    }
```
johnxnguyen, line 794: Since there will be some more flags coming, I wonder if it makes sense to encapsulate the flags into a type so that we can do something like:

```swift
SessionManager.shared.featureFlags.disableClipboard

File: Source/SessionManager/SessionManager.swift [hunk 2]
```diff
@@ -789,8 +789,8 @@ public final class SessionManager : NSObject, SessionManagerType {
         return activeUserSession != nil
     }
     
-    public var isCopyAndPasteEnable: Bool {
-        return configuration.enableCopyAndPaste
+    public var isDisabledClipboard: Bool {
```
johnxnguyen, line 792: Suggestion: `isClipboardDisabled`

Source/SessionManager/SessionManagerConfiguration.swift
```diff
@@ -63,10 +63,10 @@ public class SessionManagerConfiguration: NSObject, NSCopying, Codable {
     /// The default value of this property is `nil`, i.e. threshold is ignored
     public var failedPasswordThresholdBeforeWipe: Int?
     
-    /// If `enableCopyAndPaste` is false the copy & pasting functionality is disabled in the conversation screen.
+    /// If `disableClipboard` is true the cut, save, copy & paste functionality is disabled in the conversation screen.
```
johnxnguyen, line 66: I think `save` should have its only flag because it doesn't have anything to do with the clipboard.

Source/SessionManager/SessionManagerConfiguration.swift
```diff
@@ -63,10 +63,10 @@ public class SessionManagerConfiguration: NSObject, NSCopying, Codable {
...
     ///
-    /// The default value of this property is `true`.
-    public var enableCopyAndPaste: Bool
+    /// The default value of this property is `false`.
+    public var disableClipboard: Bool
```
johnxnguyen, line 69: In sticking to the style guide, the name should be read like a noun or a description: 

```swift
public var isClipboardDisabled: Bool
```

PR #1278, opened 2020-08-06
[Feature]: Hide User Profile Camera Roll Configuration Flag

marcomaddy1982:
What's new in this PR?

from the session_manager configuration file it is now possible to show or hide the camera roll button in the user profile

Dependency

build-configuration: https://github.com/wireapp/wire-ios-build-configuration/pull/27
JIRA Ticket: https://wearezeta.atlassian.net/browse/ZIOS-13659

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -794,6 +794,10 @@ public final class SessionManager : NSObject, SessionManagerType {
             self.activeUserSession?.userProfileImage?.updateImage(imageData: imageData)
         }
     }
+    
+    public var isCameraRollOnUserProfileEnabled: Bool {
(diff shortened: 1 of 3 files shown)


Source/SessionManager/SessionManager.swift
```diff
@@ -794,6 +794,10 @@ public final class SessionManager : NSObject, SessionManagerType {
             self.activeUserSession?.userProfileImage?.updateImage(imageData: imageData)
         }
     }
+    
+    public var isCameraRollOnUserProfileEnabled: Bool {
+        return configuration.showCameraRollOnUserProfile
+    }
```
johnxnguyen, line 800: As I mentioned in another PR, I think we should consider to encapsulate these flags into a single structure so as not to bloat the session manager.

PR #1280, opened 2020-08-07
expose wipeDatabase for deleting all accounts

billypchan:
## What's new in this PR?

expose `SessionManager.deleteAllAccounts` and add new reason `ZMAccountDeletedReasonDatabaseWipped`

### DIFF
--- a/Source/Public/NSError+ZMUserSession.h
+++ b/Source/Public/NSError+ZMUserSession.h
@@ -28,7 +28,9 @@ typedef NS_ENUM(NSUInteger, ZMAccountDeletedReason) {
     /// The user account was deleted because the session expired
     ZMAccountDeletedReasonSessionExpired,
     /// The user account was deleted because the limit of failed password attempts was reached
-    ZMAccountDeletedReasonFailedPasswordLimitReached
+    ZMAccountDeletedReasonFailedPasswordLimitReached,
+    /// The user account was deleted because user chose to wipe the database (when he/she forgot the passcode)
+    ZMAccountDeletedReasonDatabaseWipped
 };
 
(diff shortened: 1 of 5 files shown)


Source/Public/NSError+ZMUserSession.h
```diff
@@ -28,7 +28,9 @@ typedef NS_ENUM(NSUInteger, ZMAccountDeletedReason) {
     /// The user account was deleted because the session expired
     ZMAccountDeletedReasonSessionExpired,
     /// The user account was deleted because the limit of failed password attempts was reached
-    ZMAccountDeletedReasonFailedPasswordLimitReached
+    ZMAccountDeletedReasonFailedPasswordLimitReached,
+    /// The user account was deleted because user chose to wipe the database (when he/she forgot the passcode)
+    ZMAccountDeletedReasonDatabaseWipped
```
johnxnguyen, line 33:
```suggestion
    ZMAccountDeletedReasonDatabaseWiped
```

Source/SessionManager/SessionManager.swift
```diff
@@ -524,7 +524,7 @@ public final class SessionManager : NSObject, SessionManagerType {
         delete(account: account, reason: .userInitiated)
     }
     
-    fileprivate func deleteAllAccounts(reason: ZMAccountDeletedReason) {
+    public func deleteAllAccounts(reason: ZMAccountDeletedReason) {
```
johnxnguyen, line 527: To avoid misusing the this method, I think it would be better to expose access to it through a new method, such as

```swift
public func wipeDatabase() {
    deleteAllAccounts(reason: .databaseWiped)
}
```

billypchan:
@zenkins release minor

PR #1286, opened 2020-08-17
[Feature] Encryption at rest

typfel:
## What's new in this PR?

Add support for encrypting the message data when its persisted to the database which can only be decrypted after the user authenticates inside the application.

---

- Split event processing into two phases. In the first phase we download, decrypt and store the events. We also do some minimal event processing for sending delivery receipts and creating notifications. The second phase happens when the application runs in the foreground, this is where the main event processing occurs like inserting new messages.
- Change the `LocalNotificationDispatcher` to create all notifications directly from events instead looking at inserted messages.
- Modify the event processing logic to take EAR into account
## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1014
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/237


Source/UserSession/SyncStatus.swift
```diff
@@ -117,6 +117,12 @@ extension Notification.Name {
         return currentSyncPhase.isSyncing
     }
     
+    var encryptionKeys: EncryptionKeys? { // TODO jacob move somewhere more appropriate
```
typfel, line 120: @johnxnguyen shouldn't we delete this property and retrieve the databaseKey from the manger object context? We need to move the `DatabaseEncryptionLockNotification` though.

typfel: I made a PR for this https://github.com/wireapp/wire-ios-sync-engine/pull/1287

johnxnguyen:
@zenkins release major

PR #1288, opened 2020-08-19
[Fix]: NSValueTransformer Warning

marcomaddy1982:
### Issues

At runtime we have the following warning:

CoreData: warning: Property 'payload' on Entity 'StoredUpdateEvent' is using nil or an insecure NSValueTransformer. Please switch to using "NSSecureUnarchiveFromData" or a subclass of NSSecureUnarchiveFromDataTransformer instead.

This is because using NSValueTransformer for core data model attribute is insecure and deprecated.

More info: https://www.kairadiagne.com/2020/01/13/nssecurecoding-and-transformable-properties-in-core-data.html

### Solutions

Created a new model version and all the payload transformable attribute (properties) is now subclass of NSSecureUnarchiveFromData

## Dependencies

data-model: https://github.com/wireapp/wire-ios-data-model/pull/1017


johnxnguyen (commented):
Is `NSDictionary` a plist type? If so, you don't need to create your own subclass of `NSSecureUnarchiveFromDataTransformer`.

marcomaddy1982:
> Is `NSDictionary` a plist type? If so, you don't need to create your own subclass of `NSSecureUnarchiveFromDataTransformer`.

Actually I was thinking the same

johnxnguyen (approved):

marcomaddy1982:
@zenkins release minor

PR #1289, opened 2020-08-19
[Fix] Don't create notifications for confirmation messages

johnxnguyen:
### Issues

User notifications (stating "New message") appear when no new message is visible.

### Causes

The notification is generated for a confirmation message (which is not visible in the conversation). 

### Solutions

Don't generate notifications for confirmation messages.

### DIFF
--- a/Source/Notifications/Push notifications/Notification Types/LocalNotificationContentType.swift
+++ b/Source/Notifications/Push notifications/Notification Types/LocalNotificationContentType.swift
@@ -183,6 +183,8 @@ public enum LocalNotificationContentType : Equatable {
             default:
                 return .fileUpload
             }
+        case .confirmation:
+            return nil
         default:
            return .undefined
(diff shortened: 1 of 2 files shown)


Source/Notifications/Push notifications/Notification Types/LocalNotificationContentType.swift
```diff
@@ -183,6 +183,8 @@ public enum LocalNotificationContentType : Equatable {
             default:
                 return .fileUpload
             }
+        case .confirmation:
+            return nil
```
johnxnguyen, line 187: A notification type is required to create an instance of `ZMLocalNotification`, so returning `nil` here means we never create the notification for confirmations.

KaterinaWire (approved):

johnxnguyen:
@zenkins release patch

PR #1292, opened 2020-08-19
[Fix] Push notifications are not scheduled when the app is in the background

typfel:
### Issues

Push notifications are not scheduled when the app is in the background.

### Causes

Events are only forwarded to the `LocalNotificationDispatcher` during the main event processing, which since EAR was introduced is only happening when the app is in the foreground.

### Solutions

Implement `processEventsWhileInBackground` in `LocalNotificationDispatcher` in order to make it receive events during the background phase of events processing.


billypchan (approved):

typfel:
@zenkins release major

PR #1293, opened 2020-08-20
[Fix] Don't display notifications of unknown messages

johnxnguyen:
### Issues

User notifications stating "new message" are shown, but there are no new messages visible in the app.

### Causes

User notifications are shown for messages that we do not understand. This can happen either because we don't explicitly check for `GenericMessage` content types that produce invisible messages, or we think there is a new message but can't get an instance of the `GenericMessage`.

### Solutions

It doesn't make sense to show such notifications so remove the `.undefined` case of `LocalNotificationContentType` and the related code.

## Notes

This PR also refactors some notification code, such as how `LocalNotificationContentType` are created (moved from static functions to initializers) as well as how the `NotificationBuilder` is initialized.


KaterinaWire (approved):

typfel (approved):

johnxnguyen:
@zenkins release minor

PR #1296, opened 2020-08-24
[Chore] Hide user notification actions when EAR is enabled

johnxnguyen:
### Issues

The user notification actions for replying to and liking a message do not work when encryption at rest is enabled.
### Causes

When encryption at rest is enabled, the database is locked while the app is in the background. For this reason, we can not perform these actions.

### Solutions

Hide the actions when encryption at rest is enabled.

## Notes

Encryption at rest is enabled per account. Therefore the actions are only hidden for user notification associated with accounts in which the the feature is enabled.

### DIFF
--- a/Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Configuration.swift
+++ b/Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Configuration.swift
@@ -18,61 +18,13 @@
 
 import Foundation
 
-extension PushNotificationCategory {
-    func addMuteIfNeeded(hasTeam: Bool) -> PushNotificationCategory {
-        guard !hasTeam else {
-            return self
-        }
-        
-        switch self {
-        case .conversation:
-            return .conversationWithMute
-        case .conversationWithLike:
-            return .conversationWithLikeAndMute
-        default:
-            return self
-        }
-    }
-}
 
 extension LocalNotificationType {
-    
-    func category(hasTeam: Bool) -> String {
-        let category: PushNotificationCategory
-        
-        switch self {
-        case .calling(let callState):
-            switch (callState) {
-            case .incoming:
-                category = .incomingCall
-            case .terminating(reason: .timeout):
-                category = .missedCall
-            default :
-                category = PushNotificationCategory.conversation.addMuteIfNeeded(hasTeam: hasTeam)
-            }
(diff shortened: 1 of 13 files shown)


Source/Notifications/Push notifications/Notification Types/PushNotificationCategory.swift
```diff
@@ -33,6 +33,8 @@ enum PushNotificationCategory: String, CaseIterable {
     case conversationWithLikeAndMute = "conversationCategoryWithLikeAndMute"
     case connect = "connectCategory"
     case alert = "alertCategory"
+    case conversationUnderEncryptionAtRest = "conversationUnderEncryptionAtRestCategory"
+    case conversationUnderEncryptionAtRestWithMute = "conversationUnderEncryptionAtRestCategoryWithMute"
```
typfel, line 37:
```suggestion
    case conversationUnderEncryptionAtRestWithMute = "conversationUnderEncryptionAtRestWithMuteCategory"
```

Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Configuration.swift
```diff
@@ -99,3 +51,80 @@ extension LocalNotificationType {
...
+        switch self {
+        case .conversation, .conversationWithLike:
+            return .conversationUnderEncryptionAtRest
+        case .conversationWithMute, .conversationWithLikeAndMute:
+            return .conversationUnderEncryptionAtRestWithMute
+        default:
+            return self
+        }
+    }
+}
```
typfel, line 130: This is not very easy to read but I understand that it was already setup this way before.

typfel (approved):

Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Configuration.swift
typfel: Maybe create a refactoring ticket?

johnxnguyen: Yeah I know, to be honest there's a lot of refactoring that I would like to do regarding notifications. I'll make a ticket 👍

johnxnguyen:
@zenkins release minor

PR #1297, opened 2020-08-24
[Fix] Avoid sending encrypted messages while fetching the notification stream

typfel:
### Issues

It was possible that the `CallingRequestStrategy` could start sending encrypted messages before we've finished fetching and decrypting all encrypted events in the notification stream.

### Causes

The `CallingRequestStrategy` was not a subclass of AbstractRequestStrategy and therefore weren't following any rules about when it's allowed to make requests.

### Solutions

Make `CallingRequestStrategy` an `AbstractRequestStrategy` and configure it not run during quickSync phase, which is the phase where we download and decrypt events.

In order to express this the ZMStrategyConfigurationOption was expanded to differentiate between:

slow syncing
quick syncing
waiting for the web socket connection.

## Dependencies
Needs releases with:

- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/239


johnxnguyen (approved):

typfel:
@zenkins release major

PR #1298, opened 2020-08-26
[Refactor] Remove files moved to request strategy

KaterinaWire:
## What's new in this PR?
To be able to convert an update event to a localNotification in Notification engine, we need to have ZMLocalNotification in the Request strategy. In this PR ZMLocalNotification and all required classes are removed.

## Dependencies
wireapp/wire-ios-request-strategy/pull/240


billypchan (approved):

Source/Notifications/Push notifications/LocalNotificationDispatcher.swift
```diff
@@ -19,18 +19,18 @@
 import Foundation
 import UserNotifications
 
-/// Creates and cancels local notifications
+// Creates and cancels local notifications
 @objcMembers public class LocalNotificationDispatcher: NSObject {
```
typfel, line 23: The changes in this file can be reverted I guess since it was never moved

typfel (changes requested):

Source/Notifications/Push notifications/LocalNotificationDispatcher.swift
KaterinaWire: Done

PR #1299, opened 2020-08-26
[Fix] label refetch not being reset after fetching labels

typfel:
### Issues

The labels would get re-fetched over an over again.

### Causes

If the `needsToRefetchLabels` property was ever set to true if would never be set to false and thus we would download the labels over and over again. 

The `needsToRefetchLabels` property is only set to true in a hot fix patch which was added when the folder feature was released.

### Solutions

Reset `needsToRefetchLabels` to false after fetching the labels.


johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1300, opened 2020-08-28
[Fix] Delete account when self client is deleted remotely

typfel:
### Issues

When EAR is enabled and your self client gets deleted you'll get logged out, If you log back in again your messages will not be readable.

### Causes

The message encryption is tied to your self client ID.

### Solutions

Delete the account/database if your client gets deleted remotely.


johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1302, opened 2020-09-01
[Refactor]  Create APNS token

KaterinaWire:
## What's new in this PR?

Sending to BE APNS or VoIP token depends on the client's OS.

## Dependencies
wireapp/wire-ios-data-model/pull/1024
wireapp/wire-ios-data-model/pull/1027
wireapp/wire-ios-utilities/pull/96


Source/SessionManager/SessionManager.swift
```diff
@@ -426,17 +432,38 @@ public final class SessionManager : NSObject, SessionManagerType {
...
         postLoginAuthenticationToken = PostLoginAuthenticationNotification.addObserver(self, queue: self.groupQueue)
         callCenterObserverToken = WireCallCenterV3.addGlobalCallStateObserver(observer: self)
         
         checkJailbreakIfNeeded()
     }
     
+    /// For iOS 13 or above we should use Non voIP push notifications. We should migrate the push token when upgrading the client OS or app version.
+    private func registerOrMigratePushToken() {
+        if #available(iOS 13.0, *) {
+            NotificationCenter.default.addObserver(forName: SessionManager.registerPushTokenNotificationName, object: nil, queue: nil) { (note) in
```
typfel, line 446: Why do we need this notification? We get notified about any new APNS token via the [UIApplicationDelegate](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622958-application). UI implementing this delegate but it could forward it to the `SessionManager`.

Source/SessionManager/SessionManager.swift
```diff
@@ -201,6 +206,7 @@ public final class SessionManager : NSObject, SessionManagerType {
     public weak var showContentDelegate: ShowContentDelegate?
     public weak var foregroundNotificationResponder: ForegroundNotificationResponder?
     public weak var switchingDelegate: SessionManagerSwitchingDelegate?
+    public var deviceAPNSToken: Data?
```
typfel, line 209: Why do we need to store the `deviceAPNSToken` in the SessionManager? It's persisted in the user client, no?

typfel (changes requested):

Source/UserSession/ZMUserSession+Push.swift
```diff
@@ -142,7 +141,11 @@ extension ZMUserSession {
             guard let selfClient = ZMUser.selfUser(in: syncMOC).selfClient() else { return }
             guard let pushToken = selfClient.pushToken else {
                 // If we don't have any push token, then try to register it again
-                self.sessionManager?.updatePushToken(for: self)
+                if #available(iOS 13.0, *) {
+                    self.sessionManager?.reRegisterPushToken()
+                } else {
+                    self.sessionManager?.updatePushToken(for: self)
+                }
```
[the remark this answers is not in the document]
KaterinaWire: Yes, you are right. I did

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -146,12 +146,23 @@ extension SessionManager: PKPushRegistryDelegate {
...
+                }
+            } else {
+                if let token = self.pushRegistry.pushToken(for: .voIP) {
+                    session.setPushKitToken(token)
+                }
             }
         }
     }
     
+    public func reRegisterPushToken() {
```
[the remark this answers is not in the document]
KaterinaWire: It's not necessary now. 
Removed.

Source/SessionManager/SessionManager.swift
KaterinaWire: I replaced it with SessionManager method.

Source/SessionManager/SessionManager.swift
KaterinaWire: Removed it

Source/SessionManager/SessionManager.swift
```diff
@@ -426,17 +426,28 @@ public final class SessionManager : NSObject, SessionManagerType {
...
         checkJailbreakIfNeeded()
     }
     
+    /// For iOS 13 or above we should use Non voIP push notifications. We should migrate the push token when upgrading the client OS or app version.
+    private func registerOrMigratePushToken() {
+        if #available(iOS 13.0, *) {
+            // register for Non voIP push notifications
+            if !application.isRegisteredForRemoteNotifications {
+                self.application.registerForRemoteNotifications()
+            }
```
johnxnguyen, line 443: Do we first need to delete existing PushKit tokens before registering for APNS?

Source/UserSession/ZMUserSession+Push.swift
```diff
@@ -74,11 +74,10 @@ struct PushTokenMetadata {
...
     var transportType: String {
-        if isSandbox {
-            return "APNS_VOIP_SANDBOX"
-        }
-        else {
-            return "APNS_VOIP"
+         if #available(iOS 13.0, *) {
+             return isSandbox ? "APNS_SANDBOX" : "APNS"
+         } else {
+            return isSandbox ? "APNS_VOIP_SANDBOX" : "APNS_VOIP"
```
johnxnguyen: To avoid repeated OS version checks, I suggest to encode more information in `PushTokenMetaData` and / or elsewhere about what kind of push token we have.

For example:

```swift
// Can be .voip or .standard
let tokenType: PushTokenType
```

You can then use this to simplify code, such as:

```swift
var transportType: String {
    switch tokenType {
        case .standard where isSandbox:
            return "APNS_SANDBOX"
        ...
    }
}
```

johnxnguyen: Or you could the transport type to the token type:

```swift
enum PushTokenType {
    case standard, voip

    var transportType: String {
        switch self {
            case .standard: return "APNS"
            case .voip: return "APNS_VOIP"
        }
    }
}

...

var transportType: String {
    let result = tokenType.transportType
    return isSandbox ? transportType + "_SANDBOX" ? transportType
}
```

Source/SessionManager/SessionManager.swift
typfel: The API for deleting push token is on swagger as `DELETE push/token/{pid}}` https://staging-nginz-https.zinfra.io/swagger-ui/

I can see two ways forward

1. We register new APNS token immediately, which I think means we'll lose the reference to any previous VoIP token (or can we hold a reference to multiple tokens?). In order to delete any still existing VOIP tokens we would need perform a `GET push/tokens` to get the ID of the existing VoIP token followed by a DELETE request.

2. If we run iOS 13 and we have a VoIP token initially just mark it to be deleted, once the token has been successfully deleted we register for new APNS token.

Source/UserSession/ZMUserSession+Push.swift
```diff
@@ -102,7 +101,7 @@ extension ZMUserSession {
         NotificationCenter.default.addObserver(self, selector: #selector(ZMUserSession.registerCurrentPushToken), name: ZMUserSession.registerCurrentPushTokenNotificationName, object: nil)
     }
 
-    func setPushKitToken(_ data: Data) {
+    public func setPushKitToken(_ data: Data) {
```
typfel, line 104: Why did you mark this as public? Do access it in the tests?

KaterinaWire: Yes

Source/UserSession/ZMUserSession+Push.swift
KaterinaWire: Changed

Source/SessionManager/SessionManager.swift
KaterinaWire: I'm deleting existing PushKit tokens like this:
1. After launch app we check iOS. If it's 13 (or above) we `registerForRemoteNotifications(). `
2. When we receive a response with a device token, we set it to the self client.
3. We call the `validatePushToken()` method, which compares the push token registered on the backend with the local one. If the client has iOS 13 and a voip token for this client, we send a request to remove the voip token.
The `validatePushToken()` method we also use to reset a push token from the Settings.

Source/Synchronization/Strategies/PushTokenStrategy.swift
```diff
@@ -236,3 +250,33 @@ extension PushTokenStrategy : ZMEventConsumer {
     }
 }
 
+// MARK: -  From Hex String to Data
+
+extension String {
+  /// A data representation of the hexadecimal bytes in this string.
+  public func hexDecodedData() -> Data {
```
johnxnguyen, line 257: I suggest to make this an initializer for `Data` and move it to WireUtilities. E.g. `Data(fromHexEncodedString: String)`

johnxnguyen: It would be fallible too if the string isn't a valid hex string.

Source/Synchronization/Strategies/PushTokenStrategy.swift
```diff
@@ -148,12 +148,26 @@ extension PushTokenStrategy : ZMUpstreamTranscoder {
...
-            let current = tokens.filter { $0.client == client.remoteIdentifier }
-
-            if current.count == 1 && // We found one token
-                current[0].token == pushToken.deviceTokenString // It matches what we have locally
+            let clientTokens = tokens.filter { $0.client == client.remoteIdentifier }
+            
+            let voipToken = clientTokens.filter { $0.transport.contains(PushTokenType.voip.transportType) }
+            if #available(iOS 13.0, *),
+                !voipToken.isEmpty {
+                let syncMOC = managedObjectContext.zm_sync!
```
typfel, line 158: The `managedObjectContext` is already the sync context, all strategies operate on the sync context. Since we can be sure that we already run on the SE context there's also no need to perform `syncMOC.performGroupedBlock()`

Source/Synchronization/Strategies/PushTokenStrategy.swift
```diff
@@ -148,12 +148,26 @@ extension PushTokenStrategy : ZMUpstreamTranscoder {
...
+                syncMOC.performGroupedBlock {
+                    let token = PushToken(deviceToken: voipToken[0].token.hexDecodedData(),
```
typfel, line 160: I would suggest to make a PushToken extension here in SE with a convenience initialiser which takes a `PushTokenPayload` as an argument.

**or**

Add a method on `PushTokenPayload` for constructing a `Pushtoken`, like `func createPushToken() -> PushToken`. We use this pattern here https://github.com/wireapp/wire-ios-sync-engine/blob/671d7b052600884b144b79bdb5460774ae268f1d/Source/Synchronization/Strategies/PermissionsDownloadRequestStrategy.swift#L57

Source/Synchronization/Strategies/PushTokenStrategy.swift
```diff
@@ -148,12 +148,26 @@ extension PushTokenStrategy : ZMUpstreamTranscoder {
...
+                                          appIdentifier: voipToken[0].app,
+                                          transportType: voipToken[0].transport,
+                                          isRegistered: true)
+                    client.pushToken = token.markToDelete()
+                    syncMOC.saveOrRollback()
```
typfel, line 165: Shouldn't we return early here? Because we wouldn't want to call `resetFlags()`

Source/Synchronization/Strategies/PushTokenStrategy.swift
typfel, line 257: We already have this functionality in https://github.com/wireapp/wire-ios-utilities/blob/develop/Source/Public/Data%2BZMSCrypto.swift

typfel (changes requested):

Source/UserSession/ZMUserSession+Push.swift
```diff
@@ -102,8 +111,8 @@ extension ZMUserSession {
         NotificationCenter.default.addObserver(self, selector: #selector(ZMUserSession.registerCurrentPushToken), name: ZMUserSession.registerCurrentPushTokenNotificationName, object: nil)
     }
 
-    func setPushKitToken(_ data: Data) {
-        let metadata = PushTokenMetadata.current
+    public func setPushKitToken(_ data: Data, tokenType: PushTokenType) {
```
johnxnguyen, line 114: "PushKitToken" is referring to the old PushKit way of getting the (voip) token. 

I think it would be nice if we had a method like:

```swift
    public func setPushToken(_ token: PushToken)
```

Perhaps we could add a token type property to `PushToken`. I think it would make it a bit clearer rather than checking OS versions and code paths.

Source/Synchronization/Strategies/PushTokenStrategy.swift
johnxnguyen: Where? I only see the encoding function.

typfel: Oh missed that it decoding hex.

KaterinaWire: We have `zmHexEncodedString()` in Data extension that converts data to the hexEncodedString. What if to keep the same pattern and create `zmHexDecodedData()` method in WireUtilities?

Source/Synchronization/Strategies/PushTokenStrategy.swift
KaterinaWire: Removed

Source/Synchronization/Strategies/PushTokenStrategy.swift
KaterinaWire: Added `createPushToken()` for the `PushTokenPayload`.

Source/Synchronization/Strategies/PushTokenStrategy.swift
KaterinaWire: No, because if we return after `syncMOC.saveOrRollback()` client will have a voipToken as a push token. And when we get a response after the delete request, client.pushToken will be set to nil.
So, we need to continue and set a standard token for userClient. 
`resetFlags()` will be applied for the APNS token.

typfel: But don't we need the voipToken as a push token in order to perform the deletion? I thought it works like this:

1. `markToDownload()` on existing push token, this will trigger a GET request to fetch tokens on BE.
2. Check if backend has VOIP token registered for our client and if so set it as our local PushToken and call markToDelete()
3. Now the PushTokenStrategy will delete the voip token from BE and locally
4. We re-register for a normal a token in order to populate the local PushToken with an APNS token.

Now if replace our local PushToken immediately with a regular token after calling `markToDelete()` how would the `PushTokenStrategy` know which token to delete?

Source/Synchronization/Strategies/PushTokenStrategy.swift
```diff
@@ -148,12 +148,20 @@ extension PushTokenStrategy : ZMUpstreamTranscoder {
...
-
+            
             // Find tokens belonging to self client
-            let current = tokens.filter { $0.client == client.remoteIdentifier }
-
-            if current.count == 1 && // We found one token
-                current[0].token == pushToken.deviceTokenString // It matches what we have locally
+            let clientTokens = tokens.filter { $0.client == client.remoteIdentifier }
+            
+            let voipTokens = clientTokens.filter { $0.transport.contains(PushTokenType.voip.transportType) }
```
typfel, line 155: Since you only care about the first VoIP token you could do:

```suggestion
            let voipToken = clientTokens.first(where: { $0.transport.contains(PushTokenType.voip.transportType) })
```

Source/UserSession/ZMUserSession+Push.swift
typfel, line 114: We don't need this to be public in order to access it in Swift tests, If you add `@testable import WireSyncEngine` in  your test file you'll get access to all internal methods.

KaterinaWire: I added `setPushToken(_ token: PushToken) `method, but now every time I call this method I need to create a push token object. Is it fine?

KaterinaWire: Fixed

Source/Synchronization/Strategies/PushTokenStrategy.swift
KaterinaWire: Yes, it works according to this flow.
We have in response method for delete request:
```
 // The token might have changed in the meantime, check if it's still up for deletion
            if let token = client.pushToken, token.isMarkedForDeletion {
                client.pushToken = nil
            }
            return false
```

We remove voip token from BE, but locally we already have another token.

Source/SessionManager/SessionManager.swift
```diff
@@ -438,9 +438,7 @@ public final class SessionManager : NSObject, SessionManagerType {
     private func registerOrMigratePushToken() {
         if #available(iOS 13.0, *) {
             // register for Non voIP push notifications
-            if !application.isRegisteredForRemoteNotifications {
-                self.application.registerForRemoteNotifications()
-            }
+            self.application.registerForRemoteNotifications()
```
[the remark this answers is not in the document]
johnxnguyen: Yep makes sense.

Source/Synchronization/Strategies/PushTokenStrategy.swift
johnxnguyen: You could do both quite easily, just use one implementation and wrap it in the other. Also a quick note, don't forget to test this implementation.

Source/UserSession/ZMUserSession+Push.swift
johnxnguyen: Yes that's how I imagine it, but how many times do you need to call this method? There should be only two different places, one for PushKit tokens, and one for APNS tokens, right?

Source/UserSession/ZMUserSession+Push.swift
```diff
@@ -41,59 +41,6 @@ extension Dictionary {
     }
 }
 
-struct PushTokenMetadata {
-    let isSandbox: Bool
```
johnxnguyen, line 45: Is it necessary to move this and `BuildType` to Data Model?

Source/UserSession/ZMUserSession+Push.swift
johnxnguyen: Another thing you can do is create convenience initializers or static builders for each token type, e.g. `PushToken.createVOIPToken(from: data)` and `PushToken.createAPNSToken(from: data)`.

Source/UserSession/ZMUserSession+Push.swift
KaterinaWire: I added `type: PushTokenType` property to the `PushToken` and changed a public initializer:
```
public init(deviceToken: Data, tokenType: PushTokenType, isRegistered: Bool = false) {
        let metadata = PushTokenMetadata.current(for: tokenType)
        self.init(deviceToken: deviceToken,
                  appIdentifier: metadata.appIdentifier,
                  transportType: metadata.transportType,
                  type: tokenType,
                  isRegistered: isRegistered,
                  isMarkedForDeletion: false,
                  isMarkedForDownload: false)
    }
```
So, we don't need to send as a parameter `appIdentifier` and `transportType`.

johnxnguyen: You can define the new initializer in Sync Engine through an extension, that way you don't need to move all that code to Data Model.

You do need to move `PushTokenType`, though I think it now makes more sense to make it a nested type like `PushToken.TokenType`.

Source/Synchronization/Strategies/PushTokenStrategy.swift
KaterinaWire: Done. I opened a PR in WireUtilities.

Source/UserSession/ZMUserSession+Push.swift
KaterinaWire: Done.
Move back `BuildType`.
Opened PR for DataModel.

Source/UserSession/ZMUserSession+Push.swift
KaterinaWire: Methods `PushToken.createVOIPToken(from: data) `and `PushToken.createAPNSToken(from: data`) are created.

Source/UserSession/ZMUserSession+Push.swift
KaterinaWire: Replaced this method with `setPushToken(_ token: PushToken)` without public access

Source/Synchronization/Strategies/PushTokenStrategy.swift
KaterinaWire: I've made the changes as we discussed during the call.
1. Delete `voip token` (if client has iOS 13) in the configuration method for user session.
2. Post a notification for the `.deleteToken` response that we should register for remote notifications.

johnxnguyen:
It would be helpful to have links to the other frameworks that also contain changes.

Source/SessionManager/SessionManager.swift
```diff
@@ -721,10 +728,20 @@ public final class SessionManager : NSObject, SessionManagerType {
...
         userSession.useConferenceCalling = useConferenceCalling
-        updatePushToken(for: userSession)
+        registerOrMigratePushToken(session: userSession)
         registerObservers(account: account, session: userSession)
     }
     
+    /// For iOS 13 or above we should use Non voIP push notifications. We should migrate (remove voip token and register APNS token ) the push token when upgrading the client OS or app version.
+    private func registerOrMigratePushToken(session userSession: ZMUserSession) {
+        if #available(iOS 13.0, *),
+            userSession.selfUserClient?.pushToken?.type == .voip {
```
johnxnguyen, line 738: When decoding the push token from the database, do we account for the fact there may not be a `type` property (since it was recently added?).

Tests/Source/Integration/LoginFlowTests+PushToken.swift
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    func testThatItRegistersThePushTokenWithTheBackend() {
+        // given
+        let deviceToken = "asdfasdf".data(using: .utf8)!
+        let deviceTokenAsHex = "6173646661736466"
+        XCTAssertTrue(self.login())
+        
+        // when
+        self.pushRegistry.updatePushToken(deviceToken)
```
johnxnguyen, line 41: `pushRegistry` here is an instance of `PushRegistryMock` which is a subclass of `PKPushRegistry`. You may need to update tests to account for the new APNS tokens.

Source/SessionManager/SessionManager.swift
KaterinaWire: https://github.com/wireapp/wire-ios-data-model/pull/1027

Tests/Source/Integration/LoginFlowTests+PushToken.swift
KaterinaWire: Yes, we don't need to use pushRegistry now. I removed this part.
This test already contain 
```
        let pushToken = PushToken.createAPNSToken(from: deviceToken)
        self.userSession?.setPushToken(pushToken)
```

Source/SessionManager/SessionManager.swift
```diff
@@ -426,17 +426,24 @@ public final class SessionManager : NSObject, SessionManagerType {
...
+        registerForVoipPushNotifications()
 
         postLoginAuthenticationToken = PostLoginAuthenticationNotification.addObserver(self, queue: self.groupQueue)
         callCenterObserverToken = WireCallCenterV3.addGlobalCallStateObserver(observer: self)
         
         checkJailbreakIfNeeded()
     }
     
+    ///  For iOS earlier than 13 we should register for voip push notifications
+    private func registerForVoipPushNotifications() {
```
johnxnguyen, line 438: Suggestion: `registerForVoipPushNotificationsIfNeeded`

johnxnguyen (approved):
Looking fine to me, but I would wait to get more feedback from @typfel .

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -221,3 +227,16 @@ extension SessionManager {
     }
 
 }
+
+extension SessionManager {
+    public func updateDeviceToken(_ deviceToken: Data) {
+        let pushToken = PushToken.createAPNSToken(from: deviceToken)
+        if let userSession = activeUserSession {
+            userSession.setPushToken(pushToken)
+        }
```
typfel, line 236: You can delete this.

`backgroundUserSessions` also include the active UserSession, even though the names doesn't indicate it.

Source/SessionManager/SessionManager.swift
```diff
@@ -721,10 +728,20 @@ public final class SessionManager : NSObject, SessionManagerType {
         backgroundUserSessions[account.userIdentifier] = userSession
         userSession.useConstantBitRateAudio = useConstantBitRateAudio
         userSession.useConferenceCalling = useConferenceCalling
-        updatePushToken(for: userSession)
+        registerOrMigratePushToken(session: userSession)
         registerObservers(account: account, session: userSession)
     }
     
+    /// For iOS 13 or above we should use Non voIP push notifications. We should migrate (remove voip token and register APNS token ) the push token when upgrading the client OS or app version.
+    private func registerOrMigratePushToken(session userSession: ZMUserSession) {
```
typfel, line 736:
```suggestion
    private func updateOrMigratePushToken(session userSession: ZMUserSession) {
```

typfel:
Looks good, just some minor changes suggested.

Source/SessionManager/SessionManager+Push.swift
KaterinaWire: Done

Source/SessionManager/SessionManager.swift
KaterinaWire: Done

typfel (approved):

PR #1303, opened 2020-09-04
[Feature] Wait for all background activities to end before locking the database

typfel:
### Issues

With EAR enabled it was possible to run into issues when you would send a message and put the application into the background. If the request to send the message wasn't created before the application transition to the background it would never get sent.

### Causes

The message wouldn't get sent because under EAR we would remove the database key the moment application transitions to the background at which point we would no longer have access to the message and therefore can't generate a request for it.

### Solutions

Keep the database key around until we've finished all background tasks.

## Dependencies
Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/167


billypchan (approved):

johnxnguyen (approved):

johnxnguyen (approved):

typfel:
@zenkins release minor

PR #1305, opened 2020-09-11
[Feature] Update badge count when processing notifications in the background

typfel:
### Issues

After we split the event processing into two phases the badge count is no longer updated when a push notification is received.

### Causes

The badge count is updated by calculating the number of conversation which has unread messages. The number of unread messages in a conversation is updated when we read or insert new messages in a conversation, but since we no longer insert messages in the database during the first event phase the unread message count doesn't get updated until open the app.

### Solutions

Update the unread count when displaying a notification for a message which should increase the badge count. This will only be an estimation of the unread count because this solution doesn't take into account that you might have read messages on a second device. The correct unread count will be calculated during the second event processing phase when the app is put into the foreground.
## Notes

- This solution should also work when displaying notifications using the Notification Service Extension.
- This solution currently doesn't update the badge count when you receive an incoming connection request, since that would require either refactoring how we count pending connections or start processing connection request events in the background.

### DIFF
--- a/Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification.swift
+++ b/Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification.swift
@@ -153,6 +153,68 @@ extension ZMLocalNotification {
     }
 }
 
+// MARK: - Unread Count
+
+extension ZMLocalNotification {
+            
+    func increaseEstimatedUnreadCount(on conversation: ZMConversation?) {
+        if shouldIncreaseUnreadCount {
+            conversation?.internalEstimatedUnreadCount += 1
+        }
+        
+        if shouldIncreaseUnreadMentionCount {
+            conversation?.internalEstimatedUnreadSelfMentionCount += 1
+        }
+        
+        if shouldIncreaseUnreadReplyCount {
+            conversation?.internalEstimatedUnreadSelfReplyCount += 1
+        }
+    }
+    
+    var shouldIncreaseUnreadCount: Bool {
+        switch type {
+        case .event(.connectionRequestPending):
+            return true
+        case .message(let contentType):
+            switch contentType {
+            case .messageTimerUpdate:
+                return false
+            default:
+                return true
+            }
+        default:
+            return false
+        }
+    }
+    
+    var shouldIncreaseUnreadMentionCount: Bool {
+        guard case LocalNotificationType.message(let contentType) = type else {
+            return false
+        }
+        
+        switch contentType {
+        case .text(_, isMention: true, isReply: _):
+            return true
+        default:
+            return false
+        }
+    }
+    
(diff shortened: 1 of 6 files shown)


Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification.swift
```diff
@@ -153,6 +153,73 @@ extension ZMLocalNotification {
...
+        }
+    }
+    
+}
+
+extension LocalNotificationType {
+    
+    var shouldIncreaseUnreadCount: Bool {
+        switch self {
+        case .message(let contentType):
```
johnxnguyen, line 181: A `guard case let` would be more readable and consistent with the other computed properties.

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests_UnreadCount.swift
```diff
@@ -0,0 +1,69 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+@testable import WireSyncEngine
+
+class ZMLocalNotificationTests_UnreadCount: ZMLocalNotificationTests {
+    
+    func testThatUnreadCountIsIncreased_ForContentTypes() {
+        let bar: [LocalNotificationContentType] = [.image,
```
johnxnguyen, line 25: What does `bar` mean?

johnxnguyen (changes requested):

Tests/Source/Notifications/PushNotifications/ZMLocalNotificationTests_UnreadCount.swift
typfel: Nothing, just forgot write a proper name while figuring how test this. I'll change it.

KaterinaWire (approved):

Source/Notifications/Push notifications/Notification Types/Content/ZMLocalNotification.swift
```diff
@@ -153,6 +153,73 @@ extension ZMLocalNotification {
...
 }
 
+// MARK: - Unread Count
+
+extension ZMLocalNotification {
+            
+    func increaseEstimatedUnreadCount(on conversation: ZMConversation?) {
+                
+        if type.shouldIncreaseUnreadCount {
+            conversation?.internalEstimatedUnreadCount += 1
```
KaterinaWire, line 163: Should we check if the conversation is muted or not?

typfel: If the conversation is muted we wouldn't generate a `ZMLocalNotification` in the first place as I understood the code.

johnxnguyen (approved):

typfel:
@zenkins release minor

PR #1306, opened 2020-09-14
Feature/debug commands

marcoconti83:
## What's new in this PR?

Adding support for debug commands (see [RFC](https://wearezeta.atlassian.net/wiki/spaces/~444846844/pages/39518232/RFC%3A+Public+debugging+console++for+mobile+apps?search_id=95fc9d4f-ee69-41d9-87b9-8fbf90d9ff71) and relevant [cryptobox PR](https://github.com/wireapp/wire-ios-cryptobox/pull/44).

### Testing

I tested this manually. Given the nature of the feature, which enables or disables debugging mode on other frameworks, testing this at this level will add little value but require extensive mocking. I think it's not worth it, considering this is not a user-facing feature.

### Implementation details

I'm not very happy with the initialization of the `debugCommands` in `ZMUserSession` achieved with a `var` variable (instead of let) and a late initialization at the end of the `init` method. But last time I checked there was no good way to initialize let variables that themselves need a reference to `self` to be initialized. If some appropriate language feature came along in newer Swift versions that I'm not aware of, I'd be happy update the PR accordingly.


Source/UserSession/ZMUserSession+Debugging.swift
```diff
@@ -0,0 +1,351 @@
...
+extension DebugCommand {
+    
+    /// Save any "state" that needs to be persisted. The state should
+    /// only contain types can serialized in user defaults.
+    func saveState(state: [String: Any]) {
+        self.userSession.savedDebugState[self.keyword] = state
+    }
+    
+    var savedState: [String: Any]? {
+        get { return self.userSession.savedDebugState[self.keyword] }
```
David-Henner, line 103:
```suggestion
        return userSession.savedDebugState[keyword]
```

Source/UserSession/ZMUserSession+Debugging.swift
```diff
@@ -0,0 +1,351 @@
...
+    /// when the user session is initialized
+    func restoreFromState()
+}
+
+extension DebugCommand {
+    
+    /// Save any "state" that needs to be persisted. The state should
+    /// only contain types can serialized in user defaults.
+    func saveState(state: [String: Any]) {
+        self.userSession.savedDebugState[self.keyword] = state
```
David-Henner, line 99:
```suggestion
        userSession.savedDebugState[keyword] = state
```

Source/UserSession/ZMUserSession+Debugging.swift
```diff
@@ -0,0 +1,351 @@
...
+/// The result of a debug command
+public enum DebugCommandResult {
+    /// The command was a success. There is a string to show to the user
+    case Success(info: String?)
+    /// The command was a success. There is a file to show to the user
+    case SuccessWithFile(file: NSURL)
+    /// The command failed
+    case Failure(error: String?)
+    /// The command was not recognized
+    case UnknownCommand
```
David-Henner, line 143: by convention enum cases should be lowercased

Source/UserSession/ZMUserSession+Debugging.swift
```diff
@@ -0,0 +1,351 @@
...
+}
+
+
+// MARK: - Commands List
```
David-Henner, line 147: Lonely comment here 😄

Source/UserSession/ZMUserSession+Debugging.swift
```diff
@@ -0,0 +1,351 @@
...
+        self.init(userId: user, clientId: client)
+    }
+}
+
+private class DebugCommandLogEncryption: DebugCommandMixin {
+    
+    var currentlyEnabledLogs: Set<EncryptionSessionIdentifier> = Set()
+    
+    private var usage: String {
+        get { "\(self.keyword) <add|remove|list> <sessionId|all>"}
```
David-Henner, line 168:
```suggestion
        "\(keyword) <add|remove|list> <sessionId|all>"
```

Source/UserSession/ZMUserSession+Debugging.swift
```diff
@@ -0,0 +1,351 @@
...
+            userSession: userSession
+        )
+    }
+    
+    override func execute(
+        arguments: [String],
+        onComplete: @escaping ((DebugCommandResult) -> ()))
+    {
+        defer {
+            self.saveEnabledLogs()
```
David-Henner, line 183: Reference to `self` isn't always necessary

Source/UserSession/ZMUserSession+Debugging.swift
```diff
@@ -0,0 +1,351 @@
...
+    override func execute(
+        arguments: [String],
+        onComplete: @escaping ((DebugCommandResult) -> ()))
+    {
+        guard let client = userSession.selfUserClient,
+            let user = userSession.selfUser as? ZMUser
+        else {
+            onComplete(.Failure(error: "No user"))
+            return
+        }
```
David-Henner, line 276:
```suggestion
        guard 
            let client = userSession.selfUserClient,
            let user = userSession.selfUser as? ZMUser
        else {
            onComplete(.Failure(error: "No user"))
            return
        }
```

johnxnguyen (commented):
Do you need to store a reference to the user session? One way to avoid the lateinit issue you mentioned would be to inject the user session into the command when needed, through the `execute` method for instance.

Source/UserSession/ZMUserSession+Debugging.swift
marcoconti83: This is a styling choice. I like to explicitly add `self`, but if the convention in the team is otherwise (I know @typfel doesn't like it :-) ) I'll remove it.

marcoconti83:
@David-Henner @johnxnguyen thanks for the comments, I addressed all of them.

marcoconti83:
> Do you need to store a reference to the user session? One way to avoid the lateinit issue you mentioned would be to inject the user session into the command when needed, through the `execute` method for instance.

It took passing the user session also to saving and loading the state, but it's done now! Please have a look again

Source/UserSession/ZMUserSession+Debugging.swift
David-Henner: Our [style guide](https://github.com/wireapp/ios-architecture/blob/master/style-guide/swift-style-guide.md#general--1) has a section about the use of `self`, best to remove it when not necessary

johnxnguyen (approved):

David-Henner (approved):

PR #1308, opened 2020-09-16
[Fix] Attempt to fix badge count not updating by explicitly saving

typfel:
### Issues

Badge count is not updated after receiving push notifications.

### Causes

From looking at logs I can see that we are not saving SE before attempting re-calculate the badge count, which means that any increases to estimated unread count of conversations will not be included.

### Solutions

Explicitly save the SE context after processing events in the background.

## Notes

I added a test for this unfortunately is still passing without this change due to a delayed saved which gets performed at the right time.

### DIFF
--- a/Source/Synchronization/ZMOperatonLoop+Background.swift
+++ b/Source/Synchronization/ZMOperatonLoop+Background.swift
@@ -34,21 +34,19 @@ fileprivate enum PushNotificationType: String {
 public extension ZMOperationLoop {
 
     @objc(fetchEventsFromPushChannelPayload:completionHandler:)
-    func fetchEvents(fromPushChannelPayload payload: [AnyHashable : Any], completionHandler: @escaping () -> Void) {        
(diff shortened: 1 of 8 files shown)


Source/Synchronization/ZMOperatonLoop+Background.swift
```diff
@@ -34,21 +34,19 @@ fileprivate enum PushNotificationType: String {
 public extension ZMOperationLoop {
 
     @objc(fetchEventsFromPushChannelPayload:completionHandler:)
-    func fetchEvents(fromPushChannelPayload payload: [AnyHashable : Any], completionHandler: @escaping () -> Void) {        
-        syncMOC.performGroupedBlock {
```
typfel, line 38: Removed an unnecessary `performGroupedBlock()` because we are always already on the SE context queue here.

johnxnguyen (approved):

typfel:
@zenkins release minor

PR #1309, opened 2020-09-17
[Feature] Support migration to and from encryption at rest

johnxnguyen:
## What's new in this PR?

This PR includes changes to support migrating to and from encryption at rest.

## Changes

Previously the feature was toggled by setting a value for `encryptMessagesAtRest` on an instance of `ZMUserSession`. This setter has been removed and replaced with a method since the operation can throw an error no if the migration can't be started.

The migration which potentially can take a long time happens on a temporary copy of the local store, only if the migration is successful will the migrated local store replace the existing local store. Since we don't want any changes to local store while the migration is in progress we are tearing down the user session and UI and display a migration screen.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/1030


Source/SessionManager/SessionManager+EncryptionAtRest.swift
```diff
@@ -0,0 +1,51 @@
...
+            }) { result in
+                switch result {
+                case .success():
+                    self?.loadSession(for: account, completion: { _ in })
+                case .failure(let error):
+                    Logging.EAR.debug("Failed to migrate account: \(error)")
+                }
+            }
+        })
+    }
```
johnxnguyen, line 49: Do I understand correctly the flow here?

- Set encryption at rest on the session manager
- The delegate is informed the account will be migrated
- The delegate responds to this somehow
- The delegate will invoke the `userSessionCanBeTornDown` closure
- The account is migrated

It would mean that we don't perform the migration if we don't have a delegate, which isn't immediately obvious. Do we rely on the delegate to perform the migration?

Source/SessionManager/SessionManager.swift
```diff
@@ -47,7 +47,7 @@ public typealias LaunchOptions = [UIApplication.LaunchOptionsKey : Any]
     func sessionManagerWillOpenAccount(_ account: Account,
                                        from selectedAccount: Account?,
                                        userSessionCanBeTornDown: @escaping () -> Void)
-    func sessionManagerWillMigrateAccount()
+    func sessionManagerWillMigrateAccount(userSessionCanBeTornDown: @escaping () -> Void)
```
johnxnguyen, line 50: I'm not sure I understand the naming `userSessionCanBeTornDown`. What is it suppose to signify?

Source/UserSession/ZMUserSession+EncryptionAtRest.swift
```diff
@@ -21,40 +21,60 @@ import LocalAuthentication
...
+            
+            do {
+                let account = Account(userName: "", userIdentifier: storeProvider.userIdentifier)
+                let encryptionKeys = try storeProvider.contextDirectory.encryptionKeysForSettingEncryptionAtRest(enabled: newValue, account: account)
+                try managedObjectContext.enableEncryptionAtRest(encryptionKeys: encryptionKeys, skipMigration: true)
+            } catch let error {
+                Logging.EAR.error("Failed to enable/disable encryption at rest: \(error)")
+            }
+            
+        }
```
johnxnguyen, line 77: Is this exposed only for testing purposes?

Source/SessionManager/SessionManager+EncryptionAtRest.swift
typfel: The flow is:

* Set encryption at rest on the user session
* User session prepares the encryption keys and delegates the migration to the session manager
* The session manager notifies the UI about migration so that it can tear down UI and show the migration screen
* `userSessionCanBeTornDown` is called when UI is torn down and then migration can starts

Yes, we rely on the delegate to perform the migration because the migration can not happen when the user session is running.

Source/SessionManager/SessionManager.swift
typfel: The name is copied from the existing methods which also have such a callback. It means that the UI no longer holds a reference to the user session and it is therefore safe to tear down the user session.

Source/UserSession/ZMUserSession+EncryptionAtRest.swift
typfel: No, it's also used when we can active EAR by default after creating an account.

marcomaddy1982 (approved):

typfel:
@johnxnguyen let me if I answered you questions or if you want to request some changes

Source/UserSession/ZMUserSession+EncryptionAtRest.swift
johnxnguyen: Maybe we could add a `skipMigration` argument to the `setEncryptionAtRest(enabled:)` method above and move line 72 into that method. This would reduce duplication and make it skipping migration an explicit choice.

typfel: Done

typfel (approved):

PR #1310, opened 2020-09-21
[Fix] Badge count not updating on receiving a push notification

typfel:
### Issues

Badge count is not updating when receiving a notification in the background.

### Causes

After the EAR refactoring we started to [calculate an estimated unread count](https://github.com/wireapp/wire-ios-sync-engine/pull/1305) when scheduling notifications. This change was relying on the fact this estimated unread count would get updated again when the app put into the foreground and we perform the full event processing and read unread count can be calculated.

This didn't work because the estimated unread count was updated too early because we calculate the unread count when a `ZMConversation` is awoken by a fetch. This was happening before we could update the badge count and was reverting estimated unread count since messages haven't been inserted into the database at this point.

### Solutions

Stop calculating the unread count on fetches and inserts. The fact we were doing this probably is a left over from a time when we were calculating unread counts in a different way. Nowadays we re-calculate unread count on very message insert and  every time a message is read, which should cover all cases when it needs to re-calculated.

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/1031

## Notes

I moved the updating of the badge count to the event processing, this way we will also update the badge count when we fetch messages due to a background fetch and not only when receiving a push.


Source/UserSession/ZMUserSession+Push.swift
```diff
@@ -100,6 +100,10 @@ extension ZMUserSession {
     public func registerForRegisteringPushTokenNotification() {
         NotificationCenter.default.addObserver(self, selector: #selector(ZMUserSession.registerCurrentPushToken), name: ZMUserSession.registerCurrentPushTokenNotificationName, object: nil)
     }
+    
+    public func registerForUpdateBadgeCountNotification() {
+        
+    }
```
johnxnguyen, line 106: Needs to be deleted?

typfel: Yes!

johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1311, opened 2020-09-23
[Fix] Crash when enabling EAR during login

typfel:
### Issues

App would crash when attempting to activate EAR during login

### Causes

When EAR is enabled by default we activate it when the user session is created, but If you have reached the maximum number of devices you will be asked to remove a device before we have a fully functional user session. Importantly the self user is not yet accessible, which causes the EAR activation to fail since it accesses the self user in order to look up the userID (remoteIdentifier). This results in a crash due to force unwrapping an optional (nil) value.

### Solutions

Look up the userID on the `storeProvider` instead

### DIFF
--- a/Source/UserSession/ZMUserSession+EncryptionAtRest.swift
+++ b/Source/UserSession/ZMUserSession+EncryptionAtRest.swift
@@ -26,7 +26,8 @@ extension ZMUserSession {
         
         set {
             do {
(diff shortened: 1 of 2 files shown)


marcomaddy1982 (approved):

KaterinaWire (approved):

typfel:
@zenkins release patch

PR #1312, opened 2020-09-23
[Chore] Expose user session protocol for the EAR interface

typfel:
## What's new in this PR?

Expose a user session protocol for the EAR feature, this will make it easier to write tests in the UI which interacts with the EAR functionality.


KaterinaWire (approved):

typfel:
@zenkins release minor

PR #1314, opened 2020-09-28
[Performance] Disable observation system during quick sync

johnxnguyen:
### Issues

When the app comes to the foreground after accumulating several update events, it can take a while to complete syncing.

### Causes

Previously update events were processed in the background when we received remote pushes. After implementing encryption at rest, this is no longer the case and most update events are now processed when the app returns to the foreground. It can easily happen that a device accumulates a lot of update events needing processing, which can lead to slow synchronization times.

### Solutions

A way we can speed this up is to disable the observation system during this event processing stage. The observation system is used to keep the UI up-to-date when events are processed, however it is quite an expensive and makes little sense during the quick sync phase, as it essentially "replays" the history of the update events as they are being processed. By disabling this system during quick sync, the UI will be updated in a single step once all buffered events have been processed.

## Notes

**JIRA:** https://wearezeta.atlassian.net/browse/ZIOS-13814

### DIFF
--- a/Source/UserSession/ZMUserSession.swift
+++ b/Source/UserSession/ZMUserSession.swift
@@ -459,6 +459,7 @@ extension ZMUserSession: ZMSyncStateDelegate {
     
     public func didStartQuickSync() {
         managedObjectContext.performGroupedBlock { [weak self] in
+            self?.notificationDispatcher.isDisabled = true
             self?.isPerformingSync = true
             self?.updateNetworkState()
         }
@@ -470,7 +471,9 @@ extension ZMUserSession: ZMSyncStateDelegate {
         syncStrategy.processAllEventsInBuffer()
         let hasMoreEventsToProcess = syncStrategy.processEventsAfterFinishingQuickSync()
         
+
         managedObjectContext.performGroupedBlock { [weak self] in
+            self?.notificationDispatcher.isDisabled = hasMoreEventsToProcess
             self?.isPerformingSync = hasMoreEventsToProcess
             self?.updateNetworkState()
             self?.notifyThirdPartyServices()
--- a/Tests/Source/UserSession/ZMUserSessionTests+Syncing.swift
+++ b/Tests/Source/UserSession/ZMUserSessionTests+Syncing.swift
@@ -118,6 +118,22 @@ class ZMUserSessionTests_Syncing: ZMUserSessionTestsBase {
     
     // MARK: Quick Sync
 
+    func testThatObserverSystemIsDisabledDuringQuickSync() {
+        // when
+        sut.didStartQuickSync()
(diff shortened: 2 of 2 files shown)


Source/UserSession/ZMUserSession.swift
```diff
@@ -470,7 +471,9 @@ extension ZMUserSession: ZMSyncStateDelegate {
         syncStrategy.processAllEventsInBuffer()
         let hasMoreEventsToProcess = syncStrategy.processEventsAfterFinishingQuickSync()
         
+
         managedObjectContext.performGroupedBlock { [weak self] in
+            self?.notificationDispatcher.isDisabled = hasMoreEventsToProcess
```
typfel, line 476: So setting this to `isDisabled = false` after the event processing fires all the existing observers with a flag the thing needs to reloaded or that all properties changes?

typfel (approved):

Source/UserSession/ZMUserSession.swift
johnxnguyen: Setting it to false will start the observation again, which calls `startObserving()` for all change consumers. There appears to be only two observers, the `ConversationListObserverCenter` and the `SearchUserObserverCenter`. When the conversation list starts observing it fetches its data, but the search user observer center does nothing.

I realize now that we need to communicate the need to reload explicitly, since the observation starts every time we go to the foreground and in these cases we don't necessarily have to reload the data every time.

johnxnguyen: @typfel hmm, looking at this a bit more I'm wondering now if it is necessary. 

```swift
@objc public protocol ChangeInfoConsumer : NSObjectProtocol {
    func objectsDidChange(changes: [ClassIdentifier: [ObjectChangeInfo]])
    func startObserving()
    func stopObserving()
}
```

The naming of `startObserving` and `stopObserving` is a bit confusing, they should really be named `didStartObserving` and `didStopObserving`. Since these methods are call backs to notify that observation started or stopped, I think it's sufficient for the change consumer to decide if it needs to reload its data or not.

What do you think?

typfel: Yes, they only need to decide if they need to reload the data.

johnxnguyen: Ok, so I'll leave it as is, but I will change the names to make it the intent clearer.

typfel: Ok, I think we should review a bit how the observer system work as a separate task. Seems like it's not that robust when started & stopped and since most observers will not be notified that something might have changed.

johnxnguyen: As discussed in 1:1, I'll merge this without further changes and address this issue as a separate task.

johnxnguyen:
@zenkins release patch

PR #1315, opened 2020-09-30
[Performance] Calculating unread messages

KaterinaWire:
## What's new in this PR?

### Issues

We re-calculate the unread message count on every message insert or timestamp change. This is wasteful in a scenario where we receive x messages in the same conversation, we would then calculate the unread count x times when it's enough to do it once after inserting all the messages.
### Solutions

Mark a conversation as needed to update the unread message counter. After creating or updating all messages from update events, update the unread message counter for all conversations.

## Dependencies

wireapp/wire-ios-data-model/pull/1035
https://wearezeta.atlassian.net/browse/ZIOS-13812

TODO: write tests


Source/Synchronization/ZMSyncStrategy+EventProcessing.swift
```diff
@@ -114,12 +114,22 @@ extension ZMSyncStrategy: UpdateEventProcessor {
...
     }
-     
+    
+    ///
+    private func calculateUnreadMessagesForConversationsFrom(_ decryptedUpdateEvents: [ZMUpdateEvent]) {
+        let conversationUUIDs = decryptedUpdateEvents.compactMap { $0.conversationUUID }
+        let conversations: [ZMConversation?] = conversationUUIDs
+            .map { ZMConversation.fetch(withRemoteIdentifier: $0, in: syncMOC) }
+            .filter { $0?.needsToCalculateUnreadMessages == true }
+        conversations.forEach { $0?.calculateLastUnreadMessages() }
```
johnxnguyen, line 130: I suggest to make put these in a `Set` to remove duplicates, for instance:

```swift
let conversationIds = Set(decryptedUpdateEvents.compactMap(\.conversationUUID)
```

You should also make use of lazy chaining to avoid enumerating the collection several times:

```swift
let conversations = conversationIds.lazy
  .compactMap { ZMConversation.fetch... }
  .filter(\.needsToCalculateUnreadMessages)
  .forEach { $0.calculateLastUnreadMessages() }
```

johnxnguyen (changes requested):

Source/Synchronization/ZMSyncStrategy+EventProcessing.swift
KaterinaWire: Updated

Source/Synchronization/ZMSyncStrategy+EventProcessing.swift
```diff
@@ -114,12 +114,22 @@ extension ZMSyncStrategy: UpdateEventProcessor {
...
             syncMOC.saveOrRollback()
             
             Logging.eventProcessing.debug("Events processed in \(-date.timeIntervalSinceNow): \(self.eventProcessingTracker?.debugDescription ?? "")")
         }
     }
-     
+    
+    ///Fetch all conversation from the decryptedUpdateEvents array thatare marked as needsToCalculateUnreadMessages and calculate unread messages for them
+    private func calculateUnreadMessagesForConversationsFrom(_ decryptedUpdateEvents: [ZMUpdateEvent]) {
+        let conversationIds = Set(decryptedUpdateEvents.compactMap(\.conversationUUID))
```
typfel, line 126: I think it's bit risky to assume that we only touch conversation which referenced by the update events. For example last read update are sent in the self conversation but actually update the lastReadTimestamp is a different conversation.

I think it's safer to do fetch request for all conversation where `needsToCalculateUnreadCount` is `true`

typfel (changes requested):

Source/Synchronization/ZMSyncStrategy+EventProcessing.swift
```diff
@@ -114,12 +114,20 @@ extension ZMSyncStrategy: UpdateEventProcessor {
...
+            calculateUnreadMessagesForConversationsFrom(decryptedUpdateEvents)
             syncMOC.saveOrRollback()
             
             Logging.eventProcessing.debug("Events processed in \(-date.timeIntervalSinceNow): \(self.eventProcessingTracker?.debugDescription ?? "")")
         }
     }
-     
+    
+    ///Fetch all conversation that are marked as needsToCalculateUnreadMessages and calculate unread messages for them
+    private func calculateUnreadMessagesForConversationsFrom(_ decryptedUpdateEvents: [ZMUpdateEvent]) {
```
typfel, line 125: `decryptedUpdateEvents` argument can be removed.

typfel, line 125: I think we should move this method to data model, It could be class method on `ZMConversation`.

typfel (approved):

KaterinaWire:
@zenkins release minor

PR #1318, opened 2020-10-06
[Fix] Do not check the file extension

KaterinaWire:
### Issues
When you sent a backup file through the email, the Gmail app changes the file extension.(xxx.ios_wbu to xxx.ios-wbu).

### Causes
There is the Gmail app issue.
https://support.google.com/mail/thread/15716995?hl=en

### Solutions
Do not check the file extension and accept both file extensions.

## Dependencies

wireapp/wire-ios/pull/4634

### DIFF
--- a/Source/SessionManager/SessionManager+Backup.swift
+++ b/Source/SessionManager/SessionManager+Backup.swift
@@ -93,9 +93,6 @@ extension SessionManager {
         }
 
         guard let userId = unauthenticatedSession?.authenticationStatus.authenticatedUserIdentifier else { return completion(.failure(BackupError.notAuthenticated)) }
-
-        // Verify the imported file has the correct file extension.
-        guard location.pathExtension == BackupMetadata.fileExtension else { return completion(.failure(BackupError.invalidFileExtension)) }
         
         SessionManager.workerQueue.async(group: dispatchGroup) { [weak self] in
             guard let `self` = self else { return }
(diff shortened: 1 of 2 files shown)


marcomaddy1982 (approved):

johnxnguyen (changes requested):
I think we should keep the validation of the file extension, just extend it to now accept the two different versions.

KaterinaWire:
> I think we should keep the validation of the file extension, just extend it to now accept the two different versions.

But we never use this verification, because on the UI side we only allow certain file types to be selected.

johnxnguyen:
> But we never use this verification, because on the UI side we only allow certain file types to be selected.

@KaterinaWire indeed, but this is an assumption that verification happens outside of the framework. The sync engine doesn't know about the UI and therefore shouldn't make any assumptions about how the framework is used. It seems like a trivial thing because we know that only the UI uses the sync engine, but even small assumptions like this can lead to nasty bugs.

KaterinaWire:
> > But we never use this verification, because on the UI side we only allow certain file types to be selected.
> 
> @KaterinaWire indeed, but this is an assumption that verification happens outside of the framework. The sync engine doesn't know about the UI and therefore shouldn't make any assumptions about how the framework is used. It seems like a trivial thing because we know that only the UI uses the sync engine, but even small assumptions like this can lead to nasty bugs.

Yes, it makes sense. The verification is restored and modified.

johnxnguyen (changes requested):
It would be good to add a test for the new behavior.

KaterinaWire:
> It would be good to add a test for the new behavior.

@johnxnguyen Added test

johnxnguyen (approved):

KaterinaWire:
@zenkins release patch

PR #1319, opened 2020-10-06
[Feature] call degradation

David-Henner:
## What's new in this PR?

- New behaviour for dropping calls when a security degradation occurs.
- Allow a degraded incoming call to be joined

#### Current behaviour 
Drop the call if the security level of **the conversation** degrades. 

Degradation may occur in following scenario:
_GIVEN_ **all conversation participants** are verified 
_WHEN_ one or more participants become unverified
_THEN_ conversation security degrades

#### New behaviour 
Drop the call if the security of **any call participant** degrades.

Degradation may occur in following scenario:
_GIVEN_ **any call participant** is verified 
_WHEN_ a verified participant becomes unverified
_THEN_ call security degrades

### DIFF
--- a/Source/Calling/VoiceChannelV3.swift
+++ b/Source/Calling/VoiceChannelV3.swift
@@ -180,12 +180,10 @@ extension VoiceChannelV3 : CallActionsInternal {
         var joined = false
         
         switch state {
-        case .incoming(video: _, shouldRing: _, degraded: let degraded):
-            if !degraded {
-                joined = callCenter?.answerCall(conversation: conversation, video: video) ?? false
-            }
+        case .incoming(video: _, shouldRing: _, degraded: _):
+            joined = callCenter?.answerCall(conversation: conversation, video: video) ?? false
         default:
-            joined = self.callCenter?.startCall(conversation: conversation, video: video) ?? false
+            joined = callCenter?.startCall(conversation: conversation, video: video) ?? false
         }
         
         return joined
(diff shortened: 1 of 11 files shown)


Source/Calling/VoiceChannelV3.swift
```diff
@@ -122,7 +122,19 @@ public class VoiceChannelV3 : NSObject, VoiceChannel {
 
         return callCenter.isConferenceCall(conversationId: remoteIdentifier)
     }
-    
+
+    public var firstDegradedUser: ZMUser? {
+        if let conversationId = conversation?.remoteIdentifier,
```
marcomaddy1982, line 127: Probably here using the guard statement is more "natural" but feel free to leave as it is if you like more.

Source/Calling/WireCallCenterV3+Events.swift
```diff
@@ -326,6 +326,18 @@ extension WireCallCenterV3 {
     }
 }
 
+extension WireCallCenterV3 {
```
marcomaddy1982, line 329: The class is already pretty huge so for a better organisation I would add or a praga mark here before the `extension` definition or create a separate file extension: it is up to you choose the way you believe is better.

marcomaddy1982:
There are two little comments focused on the styling and not on the implementation logic so for me it is super good also as it is. I will approve it. 👍

marcomaddy1982 (approved):

David-Henner:
@zenkins release major

PR #1321, opened 2020-10-09
[Refactor]: Quit Dispatching User Session State When Blacklisted or Jailbroken

marcomaddy1982:
What's new in this PR?

### Context
When the application version is `blacklisted` or the device is `jailbroken` we would never move from that state and we would never have a transition to any other state in the `UI`.

Before the changes of the `AppState` and `AppStateCalculator` introduced with the PR https://github.com/wireapp/wire-ios/pull/4632/files, the behavior that the user should not leave a blacklisted or jailbroken state was previously handled on the UI side. With this PR we will move the logic in the sync-engine layer. So when the application version is `blacklisted` or the device is `jailbroken`  in order to avoid any other communication of the `SessionManager` state change with the `UI`, we set the `SessionManager` `delegate` to nil , in this manner no call back to the UI is triggered anymore.


johnxnguyen:
Do you mean to merge this into another branch? The pr for the UI project isn't targeting develop.

Source/SessionManager/SessionManager.swift
```diff
@@ -341,6 +341,10 @@ public final class SessionManager : NSObject, SessionManagerType {
                     if blacklisted {
                         self.isAppVersionBlacklisted = true
                         self.delegate?.sessionManagerDidBlacklistCurrentVersion()
+                        // When the application version is blacklisted we would never have a
+                        // transition to any other state in the UI, so we won't inform it
+                        // anymore setting the delegate to nil.
+                        self.delegate = nil
```
typfel, line 347:
```suggestion
                        // When the application version is blacklisted we don't want have a
                        // transition to any other state in the UI, so we won't inform it
                        // anymore by setting the delegate to nil.
                        self.delegate = nil
```

johnxnguyen:
Also just some notes about the PR template:

1. I wouldn't tag this as a [Feature], it's part of a refactor so [Refactor] would be more appropriate.
2. This PR doesn't depend on the PR for the UI project, it's the other way around.
3. Providing a little more contextual information in the description will help the reader understand the reason / need for these changes. For example, I would also mention that the behavior that we should not leave a blacklisted state was previously handled on the UI side, but this is no longer the case. Therefore we enforce it on sync engine side (hence this PR).

Addressing these three points will make it easier for anyone to review a PR, not just those that are familiar with the changes.

marcomaddy1982:
> Do you mean to merge this into another branch? The pr for the UI project isn't targeting develop.

Yep I thought the same.

marcomaddy1982:
> Also just some notes about the PR template:
> 
> 1. I wouldn't tag this as a [Feature], it's part of a refactor so [Refactor] would be more appropriate.
> 2. This PR doesn't depend on the PR for the UI project, it's the other way around.
> 3. Providing a little more contextual information in the description will help the reader understand the reason / need for these changes. For example, I would also mention that the behavior that we should not leave a blacklisted state was previously handled on the UI side, but this is no longer the case. Therefore we enforce it on sync engine side (hence this PR).
> 
> Addressing these three points will make it easier for anyone to review a PR, not just those that are familiar with the changes.

Check now and let me know if now it is more clear, please.

marcomaddy1982:
> Do you mean to merge this into another branch? The pr for the UI project isn't targeting develop.

Done

johnxnguyen:
> Check now and let me know if now it is more clear, please.

@marcomaddy1982 perfetto!

typfel (approved):

Source/SessionManager/SessionManager.swift
```diff
@@ -341,6 +341,10 @@ public final class SessionManager : NSObject, SessionManagerType {
                     if blacklisted {
                         self.isAppVersionBlacklisted = true
                         self.delegate?.sessionManagerDidBlacklistCurrentVersion()
+                        // When the application version is blacklisted we don't want have a
+                        // transition to any other state in the UI, so we won't inform it
+                        // anymore setting by the delegate to nil.
```
typfel, line 346:
```suggestion
                        // anymore by setting the delegate to nil.
```

Source/SessionManager/SessionManager.swift
```diff
@@ -846,6 +850,10 @@ public final class SessionManager : NSObject, SessionManagerType {
             }
             
             self.delegate?.sessionManagerDidBlacklistJailbrokenDevice()
+            // When the device is jailbroken we don't want have a
+            // transition to any other state in the UI, so we won't inform it
+            // anymore setting by the delegate to nil.
```
typfel, line 855:
```suggestion
            // anymore by setting the delegate to nil.
```

PR #1323, opened 2020-10-13
[Fix] Save the incoming call in the pendingCallKitCall property

KaterinaWire:
### Issues
When the applock is enable and client accepts a call through CallKit, the client has no visual elements about the current call.

### Solutions
Store the incoming call in the `pendingCallKitCall` property and join the call when the client unlocks the app.

## Dependencies

https://wearezeta.atlassian.net/browse/ZIOS-13792

### DIFF
--- a/Source/Calling/CallKitManager.swift
+++ b/Source/Calling/CallKitManager.swift
@@ -66,6 +66,7 @@ protocol CallKitManagerDelegate: class {
 @objc
 public class CallKitManager: NSObject {
     
+    public var pendingCallKitCall: (() -> Void)?
     fileprivate let provider: CXProvider
     fileprivate let callController: CXCallController
     fileprivate weak var delegate: CallKitManagerDelegate?
@@ -416,8 +417,10 @@ extension CallKitManager : CXProviderDelegate {
             action.fail()
         }
         
-        if call.conversation.voiceChannel?.join(video: false) != true {
-            action.fail()
+        pendingCallKitCall = {
+            if call.conversation.voiceChannel?.join(video: false) != true {
+                action.fail()
+            }
         }
     }
     


Source/Calling/CallKitManager.swift
```diff
@@ -66,6 +66,7 @@ protocol CallKitManagerDelegate: class {
 @objc
 public class CallKitManager: NSObject {
     
+    public var pendingCallKitCall: (() -> Void)?
```
johnxnguyen, line 69: Maybe better to name this `pendingCallKitAnswerAction` or something like that, to express that executing it will answer a call.

johnxnguyen (changes requested):
Please add tests.

Source/Calling/CallKitManager.swift
KaterinaWire: Done.
And I added tests.

Source/Calling/CallKitManager.swift
```diff
@@ -66,7 +66,7 @@ protocol CallKitManagerDelegate: class {
 @objc
 public class CallKitManager: NSObject {
     
-    public var pendingCallKitCall: (() -> Void)?
+    public var pendingCallKitAnswerAction: (() -> Void)?
```
johnxnguyen, line 69: I'm thinking now it might be better to make this private and instead expose a method or two that will take care of executing the closure and resetting it to `nil` afterwards. What do you think?

KaterinaWire: Yes, I think it's a better solution. I have doubts about when we should reset `pendingCallKitAnswerAction` to `nil`.
In the current version I do it in the `EndCallAction`. But if I do something like this:
```
public func handleCallKitAnswerAction() {
        pendingCallKitAnswerAction?()
        pendingCallKitAnswerAction = nil
    }
```
I'm not sure, will it cancel the call immediately?

johnxnguyen: Invoking the closure will just attempt to answer the call. Setting the closure to `nil` will have no effect on the call, we're just "consuming" the closure because it should only be executed once. Your implementation is what I had in mind, but rename the method to something like `answerPendingCallIfNeeded()`.

johnxnguyen (changes requested):
We should also include a test that the answer action is fulfilled when we execute it.

typfel:
I see a potential issue with this solution. 

One of the advantages of CallKit is that you don't need to open the app in order to answer and join a call. So we now effectively disable this behavior if you have App Lock enabled?

I think users would expect a call to connect from the lock screen and I'm afraid it will look like the call just never connects.

Did we agree with product on this behavior?

KaterinaWire:
> I see a potential issue with this solution.
> 
> One of the advantages of CallKit is that you don't need to open the app in order to answer and join a call. So we now effectively disable this behavior if you have App Lock enabled?
> 
> I think users would expect a call to connect from the lock screen and I'm afraid it will look like the call just never connects.
> 
> Did we agree with product on this behavior?

@typfel There is an issue posted by Victor
https://wearezeta.atlassian.net/browse/ZIOS-13792
and I discussed it with the design team. Eva also approved it. I will tag you in the `Design&iOS` chat where it was discussed.

Tests/Source/Calling/CallKitManagerTests.swift
```diff
@@ -175,6 +175,7 @@ class CallKitManagerTest: DatabaseTest {
     var mockWireCallCenterV3: WireCallCenterV3Mock!
     var mockTransportSession: MockTransportSession!
     var mockCallKitManagerDelegate: MockCallKitManagerDelegate!
+    var pendingCallKitAnswerAction: (() -> Void)?
```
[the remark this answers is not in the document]
KaterinaWire: Removed it. It was a temp code.

Source/Calling/CallKitManager.swift
```diff
@@ -66,6 +66,7 @@ protocol CallKitManagerDelegate: class {
 @objc
 public class CallKitManager: NSObject {
     
+    internal var pendingCallKitAnswerAction: (() -> Void)?
```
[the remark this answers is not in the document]
KaterinaWire: Done

KaterinaWire:
> We should also include a test that the answer action is fulfilled when we execute it.

During the call we found out that there is existing test that can check it: 

> We should also include a test that the answer action is fulfilled when we execute it.

During the call we found out that there is an existing test that can check it: `testThatCallAnswerActionIsFulfilledWhenCallIsEstablished`

PR #1324, opened 2020-10-13
[Chore] remove conference calling flag

David-Henner:
## What's new in this PR?

- Remove the conference calling flag and update calling logic accordingly

### DIFF
--- a/Source/Calling/WireCallCenterV3.swift
+++ b/Source/Calling/WireCallCenterV3.swift
@@ -368,8 +368,10 @@ extension WireCallCenterV3 {
         
         endAllCalls(exluding: conversationId)
         
-        let callType = self.callType(startedWithVideo: video)
-        
+        let callType = self.callType(for: conversation,
+                                     startedWithVideo: video,
+                                     isConferenceCall: isConferenceCall(conversationId: conversationId))
+
         if !video {
             setVideoState(conversationId: conversationId, videoState: VideoState.stopped)
         }
@@ -412,7 +414,9 @@ extension WireCallCenterV3 {
             conversationType = .oneToOne
         }
 
-        let callType = self.callType(startedWithVideo: video)
+        let callType = self.callType(for: conversation,
+                                     startedWithVideo: video,
+                                     isConferenceCall: conversationType == .conference)
 
         let started = avsWrapper.startCall(conversationId: conversationId,
                                            callType: callType,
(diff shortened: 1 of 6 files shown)


johnxnguyen (changes requested):
Don't forget that the client can still answer legacy calls, so you need to provide update the callsnapshot flag `isConferenceCall` when you answer a call and start a call. This will also affect the `callType` property because if you answer a legacy call you need to adhere to the group video limit. You'll need to assert these in tests too.

Otherwise 👍

David-Henner:
> Don't forget that the client can still answer legacy calls, so you need to provide update the callsnapshot flag `isConferenceCall` when you answer a call and start a call. This will also affect the `callType` property because if you answer a legacy call you need to adhere to the group video limit. You'll need to assert these in tests too.

I understand, but looking at the previous `callType` implementation:

```
    private func callType(for conversation: ZMConversation, startedWithVideo: Bool) -> AVSCallType {
         if !useConferenceCalling && conversation.localParticipants.count > legacyVideoParticipantsLimit {	
             return .audioOnly		
         } else {		
             return startedWithVideo ? .video : .normal		
         }		
     }
```

We see that it checks if `useConferenceCalling` is `false` before looking at the legacy call limit.

That flag was always set on the client side only, so if an incoming call was using legacy but the client enabled conference calling, we would still answer with either `.video` or `.normal`. Only when starting a call we would use `.audioOnly` if conference calling wasn't enabled.

Now, `useConferenceCalling` is always true and, in the future, removing the `if - else` statement shouldn't alter behaviour.

However, If it's necessary to answer an incoming legacy call with `.audioOnly`, I think we could look at the incoming call snapshot to determine the type of call.

johnxnguyen:
> N

@David-Henner yes I know, I think that was a mistake in our initial implementation. Perhaps it's worth checking with product to be sure that we need to consider limits for legacy calls we answer. If not, then I guess we have no need to distinguish between legacy and conference calls now? Otherwise we should still retain the information that the incoming call is not a conference call.

David-Henner:
- Updated `callType` to take into account incoming legacy calls.
- Updated tests to test for legacy and conference calling

johnxnguyen (approved):

KaterinaWire (approved):

David-Henner:
@zenkins release major

PR #1325, opened 2020-10-16
[Fix] Don't generate notifications for edit messages

typfel:
### Issues

Message edits is generating notifications and therefore also increases the badge count, which doesn't go away if the original message has already been read.

### Causes

After the encryption at rest refactoring we are increasing the badge count (unread count) when we are scheduling a notification if the message is of a type which should increase the unread count. This unread count will later be re-calculated when the actual message is inserted during the second phase of the event processing.

However for message edits we don't insert a new message but instead modify an existing message and therefore we also don't re-calculate the unread count, which means that unread count will stay until another message arrives in the conversation.

### Solutions

1. Re-calculate the unread count also when applying message edits.
2. Stop displaying notifications for message edits.
3. Don't increase the unread count for message edits

I've implemented solution 2 because: 
- It  will actually align us with the behaviour on the desktop app. 
- Since EAR we no longer remove the original notification when a message gets edited, which means that a message will show up multiple times in notification list.


johnxnguyen (approved):

typfel:
Waiting for approval by product.

typfel:
Approved by product.

typfel:
@zenkins release patch

PR #1327, opened 2020-10-22
[Refactoring]: Unifying Delegates for UI presentation

marcomaddy1982:
## What's new in this PR?

The scope of the following PR is unify the delegates of the the `sync-engine` that inform the `UI` layer that something should be presented or pushed on the screen after a DeepLinks or a Push Notification. 
Before the `sync-engine` used two different delegates: `URLActionDelegate` and `ShowContentDelegate`. This two delegates are now merged in a unique one called `PresentationDelegate`. Having just one delegates will be easier to the `UI` implement it without spreading all around the code method to perform the routing related to the DeepLinks and Push Notification action.


Source/SessionManager/PresentationDelegate.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  PresentationDelegate.swift
+//  WireSyncEngine-ios
+//
+//  Created by Marco Maddalena on 22/10/2020.
+//  Copyright © 2020 Zeta Project Gmbh. All rights reserved.
+//
```
johnxnguyen, line 7: Please use the Wire header.

johnxnguyen (changes requested):

johnxnguyen (approved):

PR #1330, opened 2020-10-27
[Refactor]: AppStateCalculator Tests

marcomaddy1982:
## What's new in this PR?

 In order to improve the testability of the `AppStateCalculator` (`wire-ios`) we re-organise the `SessionMagager` in a way that the `AppStateCalculator` will not perform `async` tasks and don't need to access to the `SessionManager` properties through a singletons (`SessionManager.share()`). To achieve this scope the following changes are made:

1) Moving `async` tasks made in the `AppStateCalculator` directly in the `sessionManager` and then inform the `UI` through the `SessionMagagerDelegate`.  For Instance:

```swift
func checkIfLoggedIn(userSession : ZMUserSession)
```

or 

```swift
let databaseEncryptionObserverToken = session.registerDatabaseLockedHandler({ [weak self] isDatabaseLocked in
             self?.delegate?.sessionManagerDidReportDatabaseLockChange(isLocked: session.isDatabaseLocked)
         })
```

2) Passing through the `SessionMagagerDelegate` all the necessary parameters needed to the `UI` without accessing them using a singleton.


Source/SessionManager/SessionManager.swift
```diff
@@ -35,7 +35,8 @@ public typealias LaunchOptions = [UIApplication.LaunchOptionsKey : Any]
 }
 
 @objc public protocol SessionActivationObserver: class {
-    func sessionManagerActivated(userSession : ZMUserSession)
+    func sessionManagerUserIsLoggedIn(isDatabaseLocked: Bool)
+    func sessionManagerRegisterDatabaseLocked(isDatabaseLocked: Bool)
```
typfel, line 39: This approach looks like it should work. I would change the names though to still reflect that the session manager operates on user sessions. Maybe it even make sense to not distinguish between the the database locked/unlock case.

```suggestion
    func sessionManagerChangedActiveUserSession(isDatabaseLocked: Bool)
```

Source/SessionManager/SessionManager.swift
```diff
@@ -515,6 +516,7 @@ public final class SessionManager : NSObject, SessionManagerType {
             let error = NSError(code: .addAccountRequested, userInfo: userInfo)
             if self?.activeUserSession == nil {
                 // If the user is already unauthenticated, we dont need to log out the current session
+                self?.databaseEncryptionObserverToken = nil
```
typfel, line 519: We have no active user session here, therefore `databaseEncryptionObserverToken` should already be `nil`.

Source/SessionManager/SessionManager.swift
```diff
@@ -587,6 +589,7 @@ public final class SessionManager : NSObject, SessionManagerType {
         
         self.createUnauthenticatedSession(accountId: deleteAccount ? nil : account.userIdentifier)
         
+        databaseEncryptionObserverToken = nil
```
typfel, line 592: I would couple nilling there `databaseEncryptionObserverToken ` to when we set the `activeUserSession` to `nil`. Maybe introduce a `tearDownActiveUserSession()` method?

Source/SessionManager/SessionManager.swift
```diff
@@ -633,14 +636,32 @@ public final class SessionManager : NSObject, SessionManagerType {
...
+        userSession.checkIfLoggedIn { [weak self] loggedIn in
+            guard loggedIn else {
+                return
+            }
+            self?.delegate?.sessionManagerUserIsLoggedIn(isDatabaseLocked: userSession.isDatabaseLocked)
+        }
+    }
+    
+    private var databaseEncryptionObserverToken: Any? = nil
+    func registerDatabaseLocked(userSession: ZMUserSession) {
```
typfel, line 659:
```suggestion
    func registerDatabaseLockedObserver(userSession: ZMUserSession) {
```

Source/SessionManager/SessionManager.swift
typfel: or

We register the database lock observer for all users sessions by doing it inside `registerObservers(account: Account, session: ZMUserSession)`. The observer is then coupled to the life cycle of users session. Then you would need to add a `guard` before calling `delegate.sessionManagerRegisterDatabaseLocked()`, which checks that it was the `activeUserSession` was locked/unlocked.

typfel (changes requested):

Source/SessionManager/SessionManager.swift
```diff
@@ -695,16 +708,19 @@ public final class SessionManager : NSObject, SessionManagerType {
...
         let connectionRequestObserver = ConversationListChangeInfo.add(observer: self, for: ZMConversationList.pendingConnectionConversations(inUserSession: session), userSession: session)
         let unreadCountObserver = NotificationInContext.addObserver(name: .AccountUnreadCountDidChangeNotification,
-                                                                    context: account)
-        { [weak self] note in
+                                                                    context: account) { [weak self] note in
             guard let account = note.context as? Account else { return }
             self?.accountManager.addOrUpdate(account)
         }
+        let databaseEncryptionObserverToken = session.registerDatabaseLockedHandler({ [weak self] isDatabaseLocked in
+            self?.delegate?.sessionManagerChangedActiveUserSession(isDatabaseLocked: session.isDatabaseLocked)
```
typfel, line 716: Now you need to guard that  `session == activeUserSession`

Source/SessionManager/SessionManager.swift
```diff
@@ -35,7 +35,8 @@ public typealias LaunchOptions = [UIApplication.LaunchOptionsKey : Any]
 }
 
 @objc public protocol SessionActivationObserver: class {
-    func sessionManagerActivated(userSession : ZMUserSession)
+    func sessionManagerWillChangeActiveUserSession(userSession: ZMUserSession?)
```
typfel, line 38: Why is `ZMUserSession` optional?

Source/SessionManager/SessionManager.swift
```diff
@@ -35,7 +35,8 @@ public typealias LaunchOptions = [UIApplication.LaunchOptionsKey : Any]
...
+    func sessionManagerChangedActiveUserSession(isDatabaseLocked: Bool)
```
typfel, line 39: It's tricky to find good names for this situation for these two states that the active user session can be in. Maybe something like this? We **change** the active user session when we switch between the account and we **update** the current active session when it's either authenticated or when the database is locked/unlocked.

```suggestion
    func sessionManagerDidChangeActiveUserSession(_ userSession: ZMUserSession)
    func sessionManagerDidUpdateActiveUserSession(isDatabaseLocked: Bool)
```

typfel: ping @johnxnguyen

typfel (changes requested):

Source/SessionManager/SessionManager.swift
marcomaddy1982: Yep probably we don't need it.

Source/SessionManager/SessionManager.swift
johnxnguyen: `sessionManagerDidUpdateActiveUserSession` is a little misleading because the session manager isn't doing anything to the active user session, it's just passing on information that the active user session database became locked / unlocked.

Perhaps something like

```swift
func activeUserSessionDatabaseLockDidChange(isLocked: Bool)
```

marcomaddy1982: yep, makes sense. Thanks a lot!!!

marcomaddy1982: I guess we added "sessionManager" at the beginning to be aligned with all the other methods of the delegate (see also `SessionManagerDelegate`) and because basically this call back is fired by the `SessionManager`.

marcomaddy1982: Anyway I will use you hint cause it is definitely more descriptive.

typfel: Even if the session manager is just delegating information I would still write it from the session managers perspective, this makes the delegate methods consistent.

`sessionManagerDidObserveDatabaseLockChange(isLocked: Bool)`
`sessionManagerDidReportDatabaseLockChange(isLocked: Bool)`
`sessionManagerDidReportDatabaseLockChangeForActiveUserSession(isLocked: Bool)`

marcomaddy1982: ok at this point I go for 

`sessionManagerDidReportDatabaseLockChange(isLocked: Bool)`

Source/SessionManager/SessionManager.swift
```diff
@@ -695,16 +722,19 @@ public final class SessionManager : NSObject, SessionManagerType {
...
         let connectionRequestObserver = ConversationListChangeInfo.add(observer: self, for: ZMConversationList.pendingConnectionConversations(inUserSession: session), userSession: session)
         let unreadCountObserver = NotificationInContext.addObserver(name: .AccountUnreadCountDidChangeNotification,
-                                                                    context: account)
-        { [weak self] note in
+                                                                    context: account) { [weak self] note in
             guard let account = note.context as? Account else { return }
             self?.accountManager.addOrUpdate(account)
         }
+        let databaseEncryptionObserverToken = session.registerDatabaseLockedHandler({ [weak self] isDatabaseLocked in
+            self?.delegate?.sessionManagerDidReportDatabaseLockChange(isLocked: session.isDatabaseLocked)
```
typfel, line 730: You are still missing the check that `session == activeUserSession`.

Source/SessionManager/SessionManager.swift
```diff
@@ -633,14 +645,29 @@ public final class SessionManager : NSObject, SessionManagerType {
...
         }
     }
     
+    func checkIfLoggedIn(userSession : ZMUserSession) {
+        delegate?.sessionManagerDidChangeActiveUserSession(userSession: userSession)
+        
+        userSession.checkIfLoggedIn { [weak self] loggedIn in
+            guard
+                loggedIn,
+                userSession == self?.activeUserSession
```
typfel, line 663: This is redundant since we only call `checkIfLoggedIn()` on the `activeUserSession`

Source/SessionManager/SessionManager.swift
```diff
@@ -633,14 +645,29 @@ public final class SessionManager : NSObject, SessionManagerType {
...
+            self.checkIfLoggedIn(userSession: session)
             
             // Configure user notifications if they weren't already previously configured.
             self.configureUserNotifications()
             self.processPendingURLAction()
         }
     }
     
+    func checkIfLoggedIn(userSession : ZMUserSession) {
+        delegate?.sessionManagerDidChangeActiveUserSession(userSession: userSession)
```
typfel, line 658: This doesn't fit the description of the function. I would move it out again.

typfel (changes requested):

typfel (approved):
Tests are failing though

PR #1331, opened 2020-10-28
[Fix] Process events when offline

typfel:
### Issues

If a user has received messages in the background via pushes, those won't be visible if the phone is offline when the user opens the application.

### Causes

We currently only perform the second phase of events processing after the quick sync has finished and if the device offline we will never finish the quick sync.

### Solutions

Also process  any previously received events directly when the app enters the foreground.

 We now have four code paths which ends up performing the second event processing phase:
1. When the app enter the foreground
2. When the quick sync is finished
3. When the database is unlocked (in the case that EAR is enabled)
4. After receiving events through the web socket.


johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1333, opened 2020-10-29
[Fix] Indicate syncing while processing events

typfel:
### Issues

Syncing indicator isn't shown when processing events received in the background.

### Causes

The syncing indicator is only shown when we perform the quick sync but we are now processing previously received events before the quick sync. If many events have been accumulating this processing can take a few seconds.

### Solutions

Indicate that we are syncing before processing previously received events.


johnxnguyen (approved):
When we set `isPerformingSync` to `true`, the `NotificationDispatcher` enters `economical` mode, so we may also get a little performance increase with these changes too.

typfel:
@zenkins release patch

PR #1336, opened 2020-11-09
[Refactor] SessionManager account migration

typfel:
## What's new in this PR?

UI currently has logic for determining if a call `sessionManagerWillMigrateAccount()` should display the migration screen or not. It's better to have this logic inside the SessionManager.

Changes:
- Only call `sessionManagerWillMigrateAccount()` when we load a session which will become the active session. That is don't call it when loading a session for processing a push notification.
- Make the `Account` parameter non-optional in the session loading methods
- Delete the `sessionManagerWillMigrateLegacyAccount` delegate method, use `sessionManagerWillMigrateAccount` use instead.

## Notes

This PR is part of a series refactoring necessary to enable migrations to and from EAR.

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -454,9 +454,12 @@ public final class SessionManager : NSObject, SessionManagerType {
                 in: sharedContainerURL,
                 migration: { [weak self] in self?.delegate?.sessionManagerWillMigrateAccount() },
                 completion: { [weak self] userIdentifier in
-                    guard let `self` = self, let userIdentifier = userIdentifier else { return }
-                    let account = self.migrateAccount(with: userIdentifier)
-                    self.selectInitialAccount(account, launchOptions: launchOptions)
+                    guard let strongSelf = self, let userIdentifier = userIdentifier else {
+                        self?.createUnauthenticatedSession()
+                        return
+                    }
+                    let account = strongSelf.migrateAccount(with: userIdentifier)
+                    self?.selectInitialAccount(account, launchOptions: launchOptions)
             })
         }
     }
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -640,7 +640,7 @@ public final class SessionManager : NSObject, SessionManagerType {
     }
     
     fileprivate func activateSession(for account: Account, completion: @escaping (ZMUserSession) -> Void) {
-        self.withSession(for: account) { session in
+        self.withSession(for: account, notifyAboutMigration: true) { session in
             self.activeUserSession = session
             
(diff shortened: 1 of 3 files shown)


Source/SessionManager/SessionManager.swift
```diff
@@ -618,16 +620,16 @@ public final class SessionManager : NSObject, SessionManagerType {
...
          - completion: called when session is loaded or when session fails to load
      */
-    func loadSession(for account: Account?, completion: @escaping (ZMUserSession?) -> Void) {
-        guard let authenticatedAccount = account, environment.isAuthenticated(authenticatedAccount) else {
+    func loadSession(for account: Account, completion: @escaping (ZMUserSession?) -> Void) {
+        guard environment.isAuthenticated(account) else {
             completion(nil)
             
-            if let account = account, configuration.wipeOnCookieInvalid {
+            if configuration.wipeOnCookieInvalid {
```
marcomaddy1982, line 627: probably it is bit verbose the code inside the `else` of the `guard` statement. What about using the `ternary operator` and create two separate private methods for the `if` case and `else` case:

```swift
    configuration.wipeOnCookieInvalid 
         ? method1()
         : method2()
}
```

or directly 

```swift
    configuration.wipeOnCookieInvalid ? method1() : method2()
}

File: Source/SessionManager/SessionManager.swift [hunk 2]
```diff
@@ -682,7 +686,11 @@ public final class SessionManager : NSObject, SessionManagerType {
                     applicationContainer: self.sharedContainerURL,
                     userIdentifier: account.userIdentifier,
                     dispatchGroup: self.dispatchGroup,
-                    migration: { [weak self] in self?.delegate?.sessionManagerWillMigrateAccount(account) },
+                    migration: { [weak self] in
+                        if notifyAboutMigration {
```
marcomaddy1982, line 690: I usually use `guard` statement in this case but it is just a personal preference. Fell free to do as you like more.

marcomaddy1982:
@typfel it is nice to a unique callback for the `migration` state to the `UI`. For me is good. There is just one style improvement we could do.

marcomaddy1982 (approved):

Source/SessionManager/SessionManager.swift
typfel: Yep, I agree it would make it more readable to break out the logic into its own method, but I'm not trying to refactor that code here. I was just changing the optionallity on `Account`.

I'd suggest cleaning up the session loading functions in a separate PR.

Source/SessionManager/SessionManager.swift
typfel: In this case I prefer an if statement since a guard force me to include `return` statement which would be slightly confusing.

KaterinaWire (approved):

typfel:
@zenkins release major

PR #1337, opened 2020-11-09
[Refactor] Add SessionManagerConfiguration option to enable EAR by default

typfel:
## What's new in this PR?

Add SessionManagerConfiguration option to enable EAR by default.

This change is necessary because when I introduce EAR migration I don't want to perform a migration EAR is enabled by default. That would require me to add new public API to the `ZMUserSession` which could easily be miss-used, I there decided move all of the EAR by default logic to SE.

## Notes

This PR is part of a series refactoring necessary to enable migrations to and from EAR.


johnxnguyen (approved):

KaterinaWire (approved):

typfel:
@zenkins release major

PR #1341, opened 2020-11-13
[Fix]: Inform the UI when an UnauthenticatedSession is created

marcomaddy1982:
### Issues

At the app launch when the user was unauthenticated the app remain stuck 

### Causes

Some changes were made and the `UI` was was not informed anymore by the `SessionManager` that the unauthenticated session was created and the `AppState` didn't change

### Solutions

Even if we noticed that the old solution to inform the UI when an UnauthenticatedSession is created smells a bit, the fastest solution to solve the issue is using the same approach and now the UI is informed using the  `sessionManagerDidFailToLogin` delegate method of the `SessionManager` and the `AppState` changes.

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -456,6 +456,11 @@ public final class SessionManager : NSObject, SessionManagerType {
                 completion: { [weak self] userIdentifier in
                     guard let strongSelf = self, let userIdentifier = userIdentifier else {
                         self?.createUnauthenticatedSession()
+                        let userInfo = self?.accountManager.selectedAccount?.loginCredentials?.dictionaryRepresentation
+                        let error = NSError(code: .accessTokenExpired, userInfo: userInfo)
+                        self?.delegate?.sessionManagerDidFailToLogin(account: self?.accountManager.selectedAccount,
+                                                                     from: self?.accountManager.selectedAccount,
+                                                                     error: error)
                         return
                     }
                     let account = strongSelf.migrateAccount(with: userIdentifier)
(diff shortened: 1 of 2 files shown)


Source/SessionManager/SessionManager.swift
```diff
@@ -456,6 +456,11 @@ public final class SessionManager : NSObject, SessionManagerType {
                 completion: { [weak self] userIdentifier in
                     guard let strongSelf = self, let userIdentifier = userIdentifier else {
                         self?.createUnauthenticatedSession()
+                        let userInfo = self?.accountManager.selectedAccount?.loginCredentials?.dictionaryRepresentation
```
typfel, line 459: We can be sure there's I not selected account in this code path so we can skip this part.

Source/SessionManager/SessionManager.swift
```diff
@@ -456,6 +456,11 @@ public final class SessionManager : NSObject, SessionManagerType {
...
+                        let error = NSError(code: .accessTokenExpired, userInfo: userInfo)
+                        self?.delegate?.sessionManagerDidFailToLogin(account: self?.accountManager.selectedAccount,
+                                                                     from: self?.accountManager.selectedAccount,
```
typfel, line 462: Also here we can assume that `selectedAccount` will be nil

typfel (changes requested):

typfel:
I also think we should improve the `testThatItCreatesUserSessionAndNotifiesDelegateIfStoreIsAvailable` test to verify that `sessionManagerDidFailToLogin` is called.

typfel (approved):

KaterinaWire (approved):

marcomaddy1982:
@zenkins release patch

PR #1342, opened 2020-11-16
[Feature]: EAR and export backup

marcomaddy1982:
## What's new in this PR?

as we mentioned in the data-model [PR](https://github.com/wireapp/wire-ios-data-model/pull/1068) for the `EAR and export backup` before to put the backup into an encrypted container using the password the user previously entered first we have to decrypt the EAR if enabled. In order to complete this operation we need to retrieve the  `EncryptionKeys`.

## Dependencies

data-model: https://github.com/wireapp/wire-ios-data-model/pull/1068


Source/SessionManager/SessionManager+Backup.swift
```diff
@@ -41,17 +52,34 @@ extension SessionManager {
...
     }
 
-    public func backupActiveAccount(password: String, completion: @escaping BackupResultClosure) {
+    public func backupActiveAccount(password: String, completion: @escaping BackupResultClosure) throws {
         guard let userId = accountManager.selectedAccount?.userIdentifier,
               let clientId = activeUserSession?.selfUserClient?.remoteIdentifier,
               let handle = activeUserSession.flatMap(ZMUser.selfUser)?.handle else { return completion(.failure(BackupError.noActiveAccount)) }
+        
+        var encryptionKeys: EncryptionKeys?
+        if activeUserSession?.storeProvider.contextDirectory.uiContext.encryptMessagesAtRest ?? false {
```
typfel, line 61:
```suggestion
        if activeUserSession?.encryptMessagesAtRest {
```

typfel, line 61: Let's add wrap the `activeUserSession` in the guard? There's no point in continuing without a user session.

Source/SessionManager/SessionManager+Backup.swift
```diff
@@ -22,6 +22,17 @@ import ZipArchive
 import WireUtilities
 import WireCryptobox
 
+enum EARBackupError: LocalizedError {
```
typfel, line 25: I'd prefer to add another case to the existing `BackupError`

Source/SessionManager/SessionManager+Backup.swift
```diff
@@ -41,17 +52,34 @@ extension SessionManager {
         case unknown
     }
 
-    public func backupActiveAccount(password: String, completion: @escaping BackupResultClosure) {
+    public func backupActiveAccount(password: String, completion: @escaping BackupResultClosure) throws {
```
typfel, line 55: Why is this method throwing now?

typfel (changes requested):

Source/SessionManager/SessionManager+Backup.swift
```diff
@@ -41,17 +52,34 @@ extension SessionManager {
...
+            }
+        }
 
         StorageStack.backupLocalStorage(
             accountIdentifier: userId,
             clientIdentifier: clientId,
             applicationContainer: sharedContainerURL,
             dispatchGroup: dispatchGroup,
-            completion: { [dispatchGroup] in SessionManager.handle(result: $0, password: password, accountId: userId, dispatchGroup: dispatchGroup, completion: completion, handle: handle) }
+            encryptionKeys: encryptionKeys,
```
typfel, line 74:
```suggestion
            encryptionKeys: activeUserSession.managedObjectContext.encryptionKeys,
```

Could also access the `encryptionKeys` like this and then you don't need deal with the error here.

typfel (approved):

johnxnguyen (approved):

Source/UserSession/ZMUserSession+EncryptionAtRest.swift
```diff
@@ -68,6 +68,18 @@ extension ZMUserSession: UserSessionEncryptionAtRestInterface {
...
+            
+            do {
+                let account = Account(userName: "", userIdentifier: storeProvider.userIdentifier)
+                let encryptionKeys = try storeProvider.contextDirectory.encryptionKeysForSettingEncryptionAtRest(enabled: newValue, account: account)
+                try managedObjectContext.enableEncryptionAtRest(encryptionKeys: encryptionKeys, skipMigration: true)
+            } catch let error {
+                Logging.EAR.error("Failed to enable/disable encryption at rest: \(error)")
+            }
+            
+        }
```
typfel, line 82: This should be deleted.

marcomaddy1982: Exactly, that was my doubt.

PR #1343, opened 2020-11-16
[Refactor] Remove OCMock dependency in ZMSyncStrategy tests

typfel:
## What's new in this PR?

The `ZMSyncStrategy` tests needs to be improved.
- We only want to test the sync strategy logic but that's hard to do in isolation since the ZMSyncStrategy creates all request strategies.
- `ZMSyncStrategy` tests rely on `OCMock` which doesn't work in Swift.

Changes:
- [x] Move request strategy creation out of ZMSyncStrategy
- [x] Remove any direct dependency on a specific request strategy 
- [x] Re-write tests not to depend on OCMock


typfel:
passed: *regression automation test*

Source/UserSession/ZMUserSession.swift
```diff
@@ -513,3 +519,16 @@ extension ZMUserSession: URLActionProcessor {
         urlActionProcessors?.forEach({ $0.process(urlAction: urlAction, delegate: delegate)} )
     }
 }
+
+// TODO jacob temporary solution while refactoring
```
David-Henner, line 523: is this intended to stay?

Tests/Source/Synchronization/MockRequestStrategyFactory.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  MockRequestStrategyFactory.swift
+//  WireSyncEngine-iOS-Tests
+//
+//  Created by Jacob Persson on 13.11.20.
+//  Copyright © 2020 Zeta Project Gmbh. All rights reserved.
+//
```
David-Henner, line 7: is this the right header ?

Source/UserSession/ZMUserSession.swift
typfel: Thanks, I'll re-evaluate this.

Tests/Source/Synchronization/MockRequestStrategyFactory.swift
typfel: Nope, I'll update

KaterinaWire (approved):

Source/UserSession/ZMUserSession.swift
typfel: I would keep it for now. IMO to properly fix this I would need to break out the update event processing from the `ZMSyncStrategy` which is too large of a refactoring for this PR.

Tests/Source/Synchronization/MockRequestStrategyFactory.swift
typfel: Done

David-Henner (approved):

typfel:
@zenkins release patch

PR #1348, opened 2020-11-30
[Refactor]: SessionManagerDelegate: improve sessionManagerDidFailToLogin method usage

marcomaddy1982:
## What's new in this PR?

the method of the `SessionManagerDelegate` (in the new version it is not there anymore)

```swift
func sessionManagerDidFailToLogin(account: Account?,
                                      error: Error)
```

 is currently called in situations which is unexpected, this means that we have been forced to add filtering logic in the `UI` to disregard these situations.

### Solutions

adding where necessary

```swift
let account = accountManager.account(with: accountId)
guard account == accountManager.selectedAccount else { return }
```

we ensure that the `sessionManagerDidFailToLogin` it is now called when it necessary and we don't need anymore to pass to the `UI` the `account: Account?` and the `selectedAccount: Account?`.

## Dependencies

JIRA: [SQCORE-215](https://wearezeta.atlassian.net/browse/SQCORE-215)

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -40,10 +40,7 @@ public typealias LaunchOptions = [UIApplication.LaunchOptionsKey : Any]
 }
 
 @objc public protocol SessionManagerDelegate : SessionActivationObserver {
-    func sessionManagerDidFailToLogin(error: Error)
-    func sessionManagerDidFailToFetchUserIdentifier()
-    func sessionManagerDidFailLoadSession(error: Error)
-    func sessionManagerDidFailToRegisterClient(error: Error)
+    func sessionManagerDidFailToLogin(error: Error?)
     func sessionManagerWillLogout(error : Error?, userSessionCanBeTornDown: (() -> Void)?)
     func sessionManagerWillOpenAccount(_ account: Account,
                                        from selectedAccount: Account?,
@@ -458,7 +455,7 @@ public final class SessionManager : NSObject, SessionManagerType {
                 completion: { [weak self] userIdentifier in
                     guard let strongSelf = self, let userIdentifier = userIdentifier else {
                         self?.createUnauthenticatedSession()
-                        self?.delegate?.sessionManagerDidFailToFetchUserIdentifier()
+                        self?.delegate?.sessionManagerDidFailToLogin(error: nil)
                         return
                     }
                     let account = strongSelf.migrateAccount(with: userIdentifier)
@@ -635,7 +632,7 @@ public final class SessionManager : NSObject, SessionManagerType {
                 guard account == accountManager.selectedAccount else { return }
                 let error = NSError(code: .accessTokenExpired,
                                     userInfo: account.loginCredentials?.dictionaryRepresentation)
-                delegate?.sessionManagerDidFailLoadSession(error: error)
+                delegate?.sessionManagerDidFailToLogin(error: error)
             }
             
             return
@@ -1071,7 +1068,7 @@ extension SessionManager: PostLoginAuthenticationObserver {
         
         let account = accountManager.account(with: accountId)
         guard account == accountManager.selectedAccount else { return }
-        delegate?.sessionManagerDidFailToRegisterClient(error: error)
+        delegate?.sessionManagerDidFailToLogin(error: error)
     }
     
     public func authenticationInvalidated(_ error: NSError, accountId: UUID) {
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -631,6 +631,8 @@ public final class SessionManager : NSObject, SessionManagerType {
                 delete(account: account, reason: .sessionExpired)
             } else {
                 createUnauthenticatedSession(accountId: account.userIdentifier)
+                
+                guard account == accountManager.selectedAccount else { return }
                 let error = NSError(code: .accessTokenExpired,
                                     userInfo: account.loginCredentials?.dictionaryRepresentation)
                 delegate?.sessionManagerDidFailLoadSession(error: error)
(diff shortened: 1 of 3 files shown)


typfel (changes requested):
AFAIK the issue with `sessionManagerDidFailToLogin()` is that it's also called for accounts which are not the active user session or the user session we are trying to login into. That's why we had to add the filtering logic in the UI code.

Splitting `sessionManagerDidFailToLogin()` up into several methods doesn't solve this issue though, they will still be called for none active accounts.

An example of such case is when a user has two accounts A and B. A is the active account at the moment but the user session for B is also loaded since a push notification has been received previously. If the cookie for account B gets deletes it will trigger `authenticationInvalidated()`. This causes `sessionManagerDidFailToLogin()` to be called but we don't want to show the logged out screen since we are displaying the logged in state for account A.

marcomaddy1982:
> AFAIK the issue with `sessionManagerDidFailToLogin()` is that it's also called for accounts which are not the active user session or the user session we are trying to login into. That's why we had to add the filtering logic in the UI code.
> 
> Splitting `sessionManagerDidFailToLogin()` up into several methods doesn't solve this issue though, they will still be called for none active accounts.
> 
> An example of such case is when a user has two accounts A and B. A is the active account at the moment but the user session for B is also loaded since a push notification has been received previously. If the cookie for account B gets deletes it will trigger `authenticationInvalidated()`. This causes `sessionManagerDidFailToLogin()` to be called but we don't want to show the logged out screen since we are displaying the logged in state for account A.

Ok, if I understood correctly what we should do is 

```swift
guard account == accountManager.selectedAccount else { return }
```

before the `sessionManagerDidFailToLogin` is called.
Anyway if you are agree I would keep the separation the separation of the `sessionManagerDidFailToLogin` in several methods.

typfel:
> Anyway if you are agree I would keep the separation the separation of the `sessionManagerDidFailToLogin` in several methods.

But what purpose this separation achieve? In the UI you perform the same operation in all cases. I would not make the API more complex if we don't have to.

> Ok, if I understood correctly what we should do is
>
> ```
> guard account == accountManager.selectedAccount else { return }
> ```

Yes, we need to do something like that and you also need verify if we need such checks in the other call sites.

marcomaddy1982:
> > Anyway if you are agree I would keep the separation the separation of the `sessionManagerDidFailToLogin` in several methods.
> 
> But what purpose this separation achieve? In the UI you perform the same operation in all cases. I would not make the API more complex if we don't have to.
> 
> > Ok, if I understood correctly what we should do is
> > ```
> > guard account == accountManager.selectedAccount else { return }
> > ```
> 
> Yes, we need to do something like that and you also need verify if we need such checks in the other call sites.

> But what purpose this separation achieve? In the UI you perform the same operation in all cases. I would not make the API more complex if we don't have to.

ok, no problem. I will revert to the old version that's not a problem. 

> Yes, we need to do something like that and you also need verify if we need such checks in the other call sites.
I already had a look yesterday but of course I can check it again.

marcomaddy1982:
@typfel your suggestion are applied, please have a look if there is some scenario that I have missed or not considered.

Source/SessionManager/SessionManager.swift
```diff
@@ -1098,9 +1096,9 @@ extension SessionManager: PostLoginAuthenticationObserver {
                 createUnauthenticatedSession(accountId: accountId)
             }
             
-            delegate?.sessionManagerDidFailToLogin(account: accountManager.account(with: accountId),
-                                                   from: accountManager.selectedAccount,
-                                                   error: error)
+            let account = accountManager.account(with: accountId)
+            guard account == accountManager.selectedAccount else { return }
```
typfel, line 1100: I think we need to do a similar check in `clientRegistrationDidFail()`

typfel (changes requested):

Source/SessionManager/SessionManager.swift
```diff
@@ -1071,21 +1066,24 @@ extension SessionManager: PostLoginAuthenticationObserver {
             createUnauthenticatedSession(accountId: accountId)
         }
         
-        delegate?.sessionManagerDidFailToLogin(account: accountManager.account(with: accountId),
-                                               from: accountManager.selectedAccount,
-                                               error: error)
+        let account = accountManager.account(with: accountId)
+        guard account == accountManager.selectedAccount else { return }
+        delegate?.sessionManagerDidFailToLogin(error: error)
```
marcomaddy1982, line 1071: @typfel I'm responding here to the other comment you made. It is already done in the `clientRegistrationDidFail()`. It is bit unclear the diff GitHub is showing for this change.

typfel: Ah I missed that 👍

Source/SessionManager/SessionManager.swift
```diff
@@ -633,11 +628,11 @@ public final class SessionManager : NSObject, SessionManagerType {
                 delete(account: account, reason: .sessionExpired)
             } else {
                 createUnauthenticatedSession(accountId: account.userIdentifier)
+                
+                guard account == accountManager.selectedAccount else { return }
```
typfel, line 632: Here we should not have this check because the selected account is updated afterwards in the `completion` block.  I think we assume that we always want to inform the UI in this scenario.

typfel (changes requested):

Source/SessionManager/SessionManager.swift
typfel: I guess gitub showed me an outdated diff.

Source/SessionManager/SessionManager.swift
marcomaddy1982: Done

typfel (approved):

Source/SessionManager/SessionManager.swift
```diff
@@ -457,10 +455,7 @@ public final class SessionManager : NSObject, SessionManagerType {
                 completion: { [weak self] userIdentifier in
                     guard let strongSelf = self, let userIdentifier = userIdentifier else {
                         self?.createUnauthenticatedSession()
-                        let error = NSError(code: .accessTokenExpired, userInfo: nil)
-                        self?.delegate?.sessionManagerDidFailToLogin(account: nil,
-                                                                     from: nil,
-                                                                     error: error)
+                        self?.delegate?.sessionManagerDidFailToLogin(error: nil)
```
KaterinaWire, line 458: Why `error` is nil here?

marcomaddy1982: We don't need there to pass the error and we won't show any alert to the user

KaterinaWire (approved):

marcomaddy1982:
@zenkins major release

PR #1356, opened 2020-12-08
[Fix] Delete encryption keys when migrating away from EAR

typfel:
### Issues

Messages are not received after migrating away from EAR

### Causes

EncryptionKeys are not deleted from key chain after migrating away from EAR.

### Solutions

Delete encryption keys from keychain


marcomaddy1982 (approved):

David-Henner (approved):

typfel:
@zenkins release patch

PR #1358, opened 2020-12-10
change CallParticipant.user from ZMUser to UserType

billypchan:
## What's new in this PR?

Change `CallParticipant`.`user` property type form `ZMUser` to `UserType` for easier mocking in UI project


Source/Calling/CallState.swift
```diff
@@ -27,11 +27,11 @@ private let zmLog = ZMSLog(tag: "calling")
 
 public struct CallParticipant: Hashable {
     
-    public let user: ZMUser
+    public let user: UserType
```
typfel, line 30: This causes lot of casting since we rely on having access to a `ZMUser` internally. Maybe an alternative would be introduce a `private let concreteUser: ZMUser` and turn `let user: UserType` into a computed property?

Do we use the initialiser in the UI?

billypchan: We created `CallParticipant` in `CallInfoConfigurationTests` in UI project. If we create `private let concreteUser: ZMUser` in `CallParticipant`, non `ZMUser` object(e.g. `MockUserType`) can not store in `CallParticipant` 

The casting issue is minimal, it only occursin hashing and comparing.

typfel: I don't like dirtying the code to support tests if it can be avoided.

How about this:

```
public init(user: UserType, userId: UUID, clientId: String, state: CallParticipantState) {
         self.user
         self.userId = userId
         self.clientId = clientId
         self.state = state
}
```

- That avoids casting in the `CallParticipant` 
- The trusted map can be implemented by mapping by userID instead, [UUID: Bool]
- We add `isTrusted` to the `UserType`

billypchan: I will add `isTrusted` in another PR.

Source/Calling/CallState.swift
```diff
@@ -27,25 +27,44 @@ private let zmLog = ZMSLog(tag: "calling")
...
     public let clientId: String
+    public let userId: UUID
     public let state: CallParticipantState
 
-    public init(user: ZMUser, clientId: String, state: CallParticipantState) {
+    
+    /// Init with seperated user and user id to allow CallParticipant is Hashable even user is not hashable
```
David-Henner, line 36: I think there's a typo here

billypchan: fixed

Source/Calling/CallState.swift
```diff
@@ -27,25 +27,57 @@ private let zmLog = ZMSLog(tag: "calling")
...
 
+    public static func == (lhs: CallParticipant, rhs: CallParticipant) -> Bool {
+        return lhs.userId == rhs.userId &&
+               lhs.clientId == rhs.clientId &&
+               lhs.state == rhs.state
+    }
+
     public func hash(into hasher: inout Hasher) {
-        hasher.combine(user.remoteIdentifier)
+        hasher.combine(state)
```
typfel, line 79: So `state` was added since it was previously already included in the auto generated `func ==`?

typfel (approved):

Source/Calling/CallState.swift
billypchan: yes, I forget to add state before.

Source/Calling/CallState.swift
```diff
@@ -27,25 +27,57 @@ private let zmLog = ZMSLog(tag: "calling")
...
+                clientId: String,
+                state: CallParticipantState) {
+        self.user = user
+        self.clientId = clientId
+        self.userId = user.remoteIdentifier
+        self.state = state
+    }
 
-    public init(user: ZMUser, clientId: String, state: CallParticipantState) {
+    /// Init with separated user and user id to allow CallParticipant is Hashable even user is not hashable
```
David-Henner, line 49:
```suggestion
    /// Init with separated user and user id to allow CallParticipant to be Hashable even though user is not Hashable
```

David-Henner (approved):

billypchan:
@zenkins release minor

johnxnguyen:
@billypchan this is actually an API breaker because you changed a public property from `ZMUser` to `UserType`, meaning that any code depending on that property needs to change.

billypchan:
@johnxnguyen I checked change to a wider type `UserType` would not break the UI project.

johnxnguyen:
@billypchan that's not true, going from `ZMUser` to `UserType` hides details. If some code expected a `ZMUser` but now gets `UserType`, it now needs to cast to `ZMUser`.

PR #1362, opened 2020-12-17
[Refacto] remove use custom code instead of account password

KaterinaWire:
## What's new in this PR?

Remove `useCustomCodeInsteadOfAccountPassword` because we don't use it anymore in the UI. The fallback to unlock Wire is always a custom passcode.

Needs releases with:

- [ https://github.com/wireapp/wire-ios-data-model/pull/1092 ]


Tests/Source/Data Model/ServiceUserTests.swift
```diff
@@ -51,6 +51,8 @@ final class DummyServiceUser: NSObject, ServiceUser {
     
     var isVerified: Bool = false
     
+    var isTrusted: Bool = true
```
johnxnguyen, line 54: Is this meant to be here?

typfel: This is due to this major change which was released as minor https://github.com/wireapp/wire-ios-data-model/pull/1089

typfel: The default value should probably be false though

typfel (approved):

Tests/Source/Data Model/ServiceUserTests.swift
KaterinaWire: Removed it and pulled changes from the develop

johnxnguyen (approved):

PR #1363, opened 2020-12-17
remove state from CallParticipant hash

billypchan:
### Issues

After https://github.com/wireapp/wire-ios-sync-engine/pull/1358 is merged, UI tests failed

### Causes

`state` is added to the hash calculation, which is not included before.

### Solutions

Remove the unnecessary hash components and make a test to describe this requirement.


typfel (approved):

David-Henner (approved):

billypchan:
@zenkins release patch

PR #1364, opened 2020-12-17
chore: bump components

billypchan:
bump for components
add new user type property form the mock user object


KaterinaWire (approved):

Tests/Source/Data Model/ServiceUserTests.swift
```diff
@@ -23,6 +23,8 @@ import XCTest
 
 final class DummyServiceUser: NSObject, ServiceUser {
     
+    var isTrusted: Bool = true
```
typfel, line 26: Shouldn't trusted be false?

typfel (changes requested):

Tests/Source/Data Model/ServiceUserTests.swift
billypchan: I ran the full test suite and it passes, but I can also set it to false.

typfel: Please do I think that reflects reality more

billypchan: tested with false and it is passed

typfel (approved):

billypchan:
@zenkins release patch

PR #1366, opened 2020-12-18
[Refactor] Move featureConfigRequestStrategy from the RequestStrategy

KaterinaWire:
## What's new in this PR?

Move `featureConfigRequestStrategy` from the RequestStrategy  and trigger it during slow sync.


Source/Synchronization/Strategies/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,233 @@
...
+
+    public var contextChangeTrackers: [ZMContextChangeTracker] {
+        return [fetchSingleConfigSync]
+    }
+
+    // MARK: - Init
+    @objc
+    public init(withManagedObjectContext managedObjectContext: NSManagedObjectContext,
+                         applicationStatus: ApplicationStatus,
+                         syncStatus: SyncStatus) {
```
johnxnguyen, line 57:
```suggestion
    public init(withManagedObjectContext managedObjectContext: NSManagedObjectContext,
                applicationStatus: ApplicationStatus,
                syncStatus: SyncStatus) {
```

Source/Synchronization/Strategies/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,233 @@
...
+            forName: Self.fetchAllConfigsTriggerNotification,
+            object: nil,
+            queue: nil,
+            using: { [weak self] _ in self?.needsToFetchAllConfigs = true }
+        )
+    }
+
+    // MARK: - Overrides
+    public override func nextRequestIfAllowed() -> ZMTransportRequest? {
+        guard syncStatus.currentSyncPhase == .fetchingFeatureConfigs else { return nil }
```
johnxnguyen, line 89: Right now this strategy will only produce a request during slow sync. 

What changed in this method? Did we have a guard at all?

Source/Synchronization/Strategies/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,233 @@
...
+
+            feature.needsToBeUpdatedFromBackend = false
+
+        } catch {
+            zmLog.error("Failed to process feature config response: \(error.localizedDescription)")
+        }
+        
+        if syncStatus.currentSyncPhase == .fetchingFeatureConfigs {
+            syncStatus.finishCurrentSyncPhase(phase: .fetchingFeatureConfigs)
+        }
```
johnxnguyen, line 137: This method is for processing the response of a single config, but the slow sync is really about fetching **all** configs. So we don't want to finish the sync phase here.

Source/Synchronization/Strategies/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    public func didReceive(_ response: ZMTransportResponse, forSingleRequest sync: ZMSingleRequestSync) {
+        guard
+            sync == fetchAllConfigsSync,
+            let team = team,
+            response.result == .success,
+            let responseData = response.rawData
+        else {
+            return
```
johnxnguyen, line 166: We should probably finish the sync phase here too because otherwise we risk to block the slow sync. Do we do a similar thing in other strategies?

johnxnguyen (changes requested):
Don't forget to test generating requests during the sync phase.

Source/Synchronization/Strategies/FeatureConfigRequestStrategy.swift
KaterinaWire: Removed

Source/Synchronization/Strategies/FeatureConfigRequestStrategy.swift
KaterinaWire: Put `syncStatus.finishCurrentSyncPhase(phase: .fetchingFeatureConfigs)` into the `defer` block

Source/Synchronization/Strategies/FeatureConfigRequestStrategy.swift
KaterinaWire: Updated

johnxnguyen (approved):

billypchan (approved):

billypchan (changes requested):
test failed.

PR #1367, opened 2020-12-28
[Refactor]: Removed SessionManager Async Creation Method

marcomaddy1982:
## What's new in this PR?

According to the [RFC](https://wearezeta.atlassian.net/wiki/spaces/IOS/pages/159646019/Simplify+API+for+Login+Registration) about the login and registration API one of the improvement we would like to add is removing the async creation of the `SessionManager` in the `AppRootRouter`. To achieve this goal some changes are made in the `SessionManager`:

- remove the method:

```swift
public static func create(
        appVersion: String,
        mediaManager: MediaManagerType,
        analytics: AnalyticsType?,
        delegate: SessionManagerDelegate?,
        presentationDelegate: PresentationDelegate?,
        application: ZMApplication,
        environment: BackendEnvironmentProvider,
        configuration: SessionManagerConfiguration,
        detector: JailbreakDetectorProtocol = JailbreakDetector(),
        completion: @escaping (SessionManager) -> Void
```

- expose the `pubic init` method 
- remove the logic that wait the protected data available to create the SessionManager (it will be moved in the UI).

## Dependencies

[JIRA](https://wearezeta.atlassian.net/browse/SQCORE-285)

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -344,7 +344,7 @@ public final class SessionManager : NSObject, SessionManagerType {
          environment: BackendEnvironmentProvider,
          configuration: SessionManagerConfiguration = SessionManagerConfiguration(),
          detector: JailbreakDetectorProtocol = JailbreakDetector()) {
-        
+
         SessionManager.enableLogsByEnvironmentVariable()
         self.environment = environment
         self.appVersion = appVersion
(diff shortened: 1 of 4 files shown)


johnxnguyen (approved):

Source/SessionManager/SessionManager.swift
```diff
@@ -389,14 +353,14 @@ public final class SessionManager : NSObject, SessionManagerType {
         self.dispatchGroup = dispatchGroup
         self.configuration = configuration.copy() as! SessionManagerConfiguration
         self.jailbreakDetector = detector
-
+
```
billypchan, line 356: I would suggest running my lint script to undo this indent space changes:

https://github.com/billypchan/iosGitHubTools/blob/master/scripts/lint.sh

billypchan (commented):
undo indent space change

Source/SessionManager/SessionManager.swift
marcomaddy1982: Thanks, done!!

billypchan (approved):

marcomaddy1982:
@zenkins release major

PR #1370, opened 2021-01-04
[Feature] Feature configurations

johnxnguyen:
## What's new in this PR?

This is the feature branch for _feature configurations_. Feature configuration is a mechanism that allows the client to receive configuration payloads from the backend used to adjust the behavior of features on the client. The first feature to use this configuration is the app lock.

In this branch we: 
- add the app lock controller to the user session.
- adjust the session manager configuration object
- hook up the new `FeatureConfigRequestStrategy`

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-request-strategy/pull/266
- [x] https://github.com/wireapp/wire-ios-data-model/pull/1095


KaterinaWire (approved):

David-Henner (approved):

KaterinaWire:
@zenkins release major

PR #1372, opened 2021-01-06
[Fix] Failed testThatItPostsNotification_WhenLogoutRequestSucceeds test

KaterinaWire:
### Issues
There's a failed `testThatItPostsNotification_WhenLogoutRequestSucceeds` test.

### Causes
In the `ZMUserSession` initialiser, we create a selfUser in `uiMoc` without the `remoteIdentifier` that is causing the test to fail.

### Solutions
Create a `selfUser` before initialise the userSession in `ZMUserSessionTestsBase`.


typfel (approved):
We are now accessing the syncMOC on the main thread in the tests but I saw that we are doing this all over the place in this test :)

I would accept it like this because I'm anyway refactoring and converting these tests to swift.

johnxnguyen (approved):

PR #1374, opened 2021-01-11
[Refactor] Move featureConfigRequestStrategy from the RequestStrategy

KaterinaWire:
## What's new in this PR?

Move featureConfigRequestStrategy from the RequestStrategy and trigger it during slow sync.

Needs releases with:

- [ ] Request strategy


Source/Synchronization/Strategies/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,204 @@
...
+extension FeatureConfigRequestStrategy: ZMSingleRequestTranscoder {
+
+    public func request(for sync: ZMSingleRequestSync) -> ZMTransportRequest? {
+        guard sync == fetchAllConfigsSync else { return nil }
+        return requestToFetchAllFeatureConfigs()
+    }
+
+    private func requestToFetchAllFeatureConfigs() -> ZMTransportRequest? {
+        guard let teamId = team?.remoteIdentifier?.transportString() else { return nil }
+        return ZMTransportRequest(getFromPath: "/teams/\(teamId)/features/\(Feature.AppLock.name.rawValue)")
```
johnxnguyen, line 139: I suggest leaving a comment here stating that the endpoint for all features is not active yet so that's why we're hitting the app lock endpoint.

Source/Synchronization/Strategies/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,204 @@
...
+        guard
+            sync == fetchAllConfigsSync,
+            let team = team,
+            response.result == .success,
+            let responseData = response.rawData else { return }
+
+        do {
+            let decoder = JSONDecoder()
+            let encoder = JSONEncoder()
+            let config = try decoder.decode(ConfigResponse<Feature.AppLock>.self, from: responseData)
```
johnxnguyen, line 157: Also it's helpful to add a comment here saying that we'll need to decode all configs when we start using the right endpoint.

Tests/Source/Synchronization/Strategies/FeatureConfigRequestStrategyTests.swift
```diff
@@ -0,0 +1,180 @@
...
+            let request = self.sut.nextRequestIfAllowed()
+
+            // then
+            XCTAssertNil(request)
+            return
+        }
+    }
+
+    // MARK: - All configurations
+    func test_ItGeneratesARequest_ToFetchAllConfigs() {
```
johnxnguyen, line 100: Does this fail now because we changed the endpoint?

johnxnguyen (commented):
Perhaps we should also test that when the response completes that we update the sync status.

johnxnguyen:
This PR is adding support for slow sync so I would rename the PR.

PR #1375, opened 2021-01-11
[Fix] Restart slow sync when upgrading in order to restore lost connections

typfel:
## What's new in this PR?

Restart the slow sync when upgrading in order to recover any potentially lost connections caused by https://github.com/wireapp/wire-ios-data-model/pull/1107


johnxnguyen (approved):

David-Henner (approved):

typfel:
@zenkins release patch

PR #1378, opened 2021-01-19
[Refactor] Remove OCMock from ZMUserSession tests (Part 2)

typfel:
Re-opened because https://github.com/wireapp/wire-ios-sync-engine/pull/1365 was automatically closed when the target branch got deleted.


typfel:
@zenkins release major

PR #1379, opened 2021-01-20
[Fix] Old accounts getting logged out on slow sync

typfel:
### Issues

For a period time it was possible for a personal account to join and leave a team. Users who joined a team during this period will get logged out if we perform a slow sync or when they attempt to restore a backup from this time period. 

### Causes

If these users never lost their history they will still have an old `Team` object in their database. When we synchronise the users's team during the slow sync we will discover that are they longer a part of team, indicated by `binding = 0`.

```
{
            binding = 0;
            creator = "8se82348-6b2c-471e-a438-9e6a26040075";
            icon = default;
            id = "02250469-873d-4a40-25fb-46b8fd44f681";
            name = Team Name;
}
```

Today a user will be deleted if they leave a team but that wasn't the case previously, we don't distinguish these old team users from todays team users and delete the account locally if we discover that user is no longer a member of team.

### Solutions

Don't delete an account if a user is no longer a member of a team, just delete the team object. New team user accounts will still be deleted when they leave a team, since their clients will also be deleted which will also trigger an account to be deleted.

### Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1119


WireSyncEngine.xcodeproj/xcshareddata/xcschemes/WireSyncEngine.xcscheme
```diff
@@ -122,7 +122,7 @@
          <EnvironmentVariable
             key = "ZMLOG_TAGS"
             value = "MockTransportRequests"
-            isEnabled = "NO">
```
marcomaddy1982, line 125: Maybe, should it be this reverted?

marcomaddy1982:
@typfel a part for the only comment I have done the rest it fine for me. I will approve it 👍. I approve the PR

marcomaddy1982 (approved):

WireSyncEngine.xcodeproj/xcshareddata/xcschemes/WireSyncEngine.xcscheme
typfel: Yes, it should. Thanks 👍

johnxnguyen (approved):

typfel:
@zenkins release major

PR #1382, opened 2021-01-25
[Refactor] User session lock

johnxnguyen:
## What's new in this PR?

This PR contains changes to support a new `locked` app state, as well as some API changes from data model.

### Changes

- Introduce a `Lock` to the user session, which is used to communicate whether / how the user session is locked. This will be used in the UI to determine if the app needs to move to the `locked` app state (and show the app lock UI).
- Report potential changes to the lock through delegate callbacks from the `AppLockController`. Each session has its own app lock controller, so when that controller "opens", the user session will know about it through a delegate callback. The user session then informs its delegate (the `SessionManager`) that it became unlocked. The session manager will then inform its delegate (`AppStateCalculator` in UI project) that the active user session changed, which will then trigger a new calculation of the app state.
- Postpone some actions until after the session is unlocked. These include setting up notifications and processing url actions.
- Begin the app lock timer when the active user session changes or the app goes to the background.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/276
- [x] https://github.com/wireapp/wire-ios-data-model/pull/1121


Source/SessionManager/SessionManager.swift
```diff
@@ -605,28 +604,28 @@ public final class SessionManager : NSObject, SessionManagerType {
...
+            guard
+                loggedIn,
+                session.lock == .none
+            else {
                 return
             }
-            self?.delegate?.sessionManagerDidReportDatabaseLockChange(isLocked: userSession.isDatabaseLocked)
+
+            self?.configureUserNotifications()
+            self?.processPendingURLAction()
```
marcomaddy1982, line 625: About the `processPendingURLAction` method: there are some `URLAction` that are performed also when the user is not logged in. Indeed in the `URLAction`  object you can find the `requiresAuthentication: Bool` property. So here it seems that the pending action url is performed just if the user is logged in. My question is what happen now to that pending action that must be excused when the user is not login, f.i .`startCompanyLogin(code: UUID)`?

Source/UserSession/ZMUserSession.Lock.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Whether the session is currently locked.
+
+    var isLocked: Bool {
+        return lock != .none
+    }
+
+    /// The current session lock, if any.
+
+    var lock: Lock? {
```
marcomaddy1982, line 45: Instead of return an optional we could return .none instead of the nil? What do you think about?

Source/SessionManager/SessionManager.swift
```diff
@@ -692,10 +691,12 @@ public final class SessionManager : NSObject, SessionManagerType {
             guard let account = note.context as? Account else { return }
             self?.accountManager.addOrUpdate(account)
         }
-        let databaseEncryptionObserverToken = session.registerDatabaseLockedHandler({ [weak self] isDatabaseLocked in
+
+        let databaseEncryptionObserverToken = session.registerDatabaseLockedHandler { [weak self] _ in
             guard session == self?.activeUserSession else { return }
-            self?.delegate?.sessionManagerDidReportDatabaseLockChange(isLocked: session.isDatabaseLocked)
-        })
+            self?.delegate?.sessionManagerDidChangeActiveUserSession(userSession: session)
```
marcomaddy1982, line 697: IMHO we could use `sessionManagerDidChangeActiveUserSession` delegate method instead of having an ad hoc one but maybe @typfel could have a look too here and see what is his point of view. :-)

Source/SessionManager/SessionManager.swift
johnxnguyen: Ah I wasn't aware. I'm not sure why we have this method to check if we're logged in, because I would expect if we're not logged in then we wouldn't be activating a user session. I'm guessing we should handle url actions that don't depend on an active user session to be handled somewhere else, probably further up the chain. I'll need to take a look though and get more familiar with these actions.

Source/SessionManager/SessionManager.swift
johnxnguyen: I'm not sure I understand. What is the ad hoc one you're referring to?

Source/UserSession/ZMUserSession.Lock.swift
johnxnguyen: Do you mean to have a specific `none` case? Or to return the `none` case that belongs to `Optional`?

Source/SessionManager/SessionManager.swift
marcomaddy1982: If this can help you you can use this custom URL wire://start-sso/wire-5977c2d2-aa60-4657-bad8-4e4ed08e483a to debug a case where there is no user logged in.

johnxnguyen: [Here](https://github.com/wireapp/wire-ios-sync-engine/blob/086c56f7a884df9a95ed9d86165ec7068ad1219c/Source/SessionManager/SessionManager%2BURLActions.swift#L53) the session manager first becomes aware of the url action. Actions that don't require authentication are processed on an unauthenticated session.

But I think I see another issue, which is that if we have an action that requires authentication and there exists already an active user session, then we process that action right away. We should be careful to not process any actions in a locked state.

Source/SessionManager/SessionManager.swift
marcomaddy1982: Sorry If I wasn't clear. So what I wanted to say was: for me is fine using here the `sessionManagerDidChangeActiveUserSession` delegate method instead of the `sessionManagerDidReportDatabaseLockChange` but maybe @typfel has a different idea cause he developed the part regarding the database locking.

Source/UserSession/ZMUserSession.Lock.swift
marcomaddy1982: To have a specific `none` case and return it instead of the `nil` so the `Lock` object could be not nil. Anyway this is just a thought, feel free to leave as it is if you prefer.

Source/SessionManager/SessionManager.swift
johnxnguyen: To clarify how I'm using `sessionManagerDidChangeActiveUserSession`: a change in the lock state (whether does to an app lock timeout or a database lock) is a change to the active session, and this change can be queried by the delegate by looking at `userSession.lock`.

Source/SessionManager/SessionManager.swift
marcomaddy1982: Yes that's true, I didn't think about to this aspect. Good point. 👍

Source/SessionManager/SessionManager.swift
marcomaddy1982: Ok so it is how I thought, thanks for clarifying.

Source/SessionManager/SessionManager.swift
```diff
@@ -605,28 +604,28 @@ public final class SessionManager : NSObject, SessionManagerType {
...
+
+    func performPostUnlockActionsIfPossible(for session: ZMUserSession) {
+        session.checkIfLoggedIn { [weak self] loggedIn in
+            guard
+                loggedIn,
+                session.lock == .none
+            else {
                 return
             }
-            self?.delegate?.sessionManagerDidReportDatabaseLockChange(isLocked: userSession.isDatabaseLocked)
```
johnxnguyen, line 626: @marcomaddy1982 or @typfel do you know what the significance of this `checkIfLoggedIn` method is? Is it just a sanity check to ensure we wouldn't report a database lock if we happened to be logged out?

marcomaddy1982: @johnxnguyen  I honestly also had some doubt about `checkIfLoggedIn` but from how it is used I deduced that it is use for the reason that you just mentioned.

marcomaddy1982:
@johnxnguyen some tests are failing (I'm sure you noticed, anyway I mentioned :-) )

Source/UserSession/ZMUserSession.Lock.swift
johnxnguyen: Actually I initially implemented it as a non-optional but I felt that with the name it was so apparent that there could be no lock:

```swift
public protocol UserSessionAppLockInterface {
    var lock: ZMUserSession.Lock? { get }
}
```

vs

```swift
public protocol UserSessionAppLockInterface {
    var lock: ZMUserSession.Lock { get }
}
```

In the first declaration, it's clear there may not be a lock. But in the second, it kind of looks like there's always a lock. 

I know it's kind of obvious that there shouldn't always be a lock present, but I'd rather it be super clear even when you're just skim reading, rather than having to think more about it.

marcomaddy1982: You are right, it makes sense, let's keep as it is. Thanks

Source/SessionManager/SessionManager.swift
```diff
@@ -605,28 +604,20 @@ public final class SessionManager : NSObject, SessionManagerType {
...
-    func checkIfLoggedIn(userSession : ZMUserSession) {
-        userSession.checkIfLoggedIn { [weak self] loggedIn in
-            guard loggedIn else {
-                return
-            }
-            self?.delegate?.sessionManagerDidReportDatabaseLockChange(isLocked: userSession.isDatabaseLocked)
-        }
+
+    func performPostUnlockActionsIfPossible(for session: ZMUserSession) {
+        guard session.lock == .none else { return }
```
marcomaddy1982, line 616: Nice 👍!!

Source/SessionManager/SessionManager.swift
johnxnguyen: I removed the check for whether we're logged in because this check happens at a higher level (where I linked to before).

Source/SessionManager/SessionManager.swift
johnxnguyen: Ah yeah, it's a nice little thing that .none is the same as `nil` :)

marcomaddy1982 (approved):

KaterinaWire (approved):

KaterinaWire:
@johnxnguyen Everything is fine for me.

Source/SessionManager/SessionManager.swift
typfel: `userSession.checkIfLoggedIn` is checking two things:
- That we have a valid cookie
- That we have registered a client

I can imagine we wouldn't want to notify that the database is locked before we have a client because then the user is still on the login screen.

typfel: Can't tell if this is issue or not without debugging/running the app.

Source/SessionManager/SessionManager.swift
```diff
@@ -605,28 +610,20 @@ public final class SessionManager : NSObject, SessionManagerType {
...
+        configureUserNotifications()
```
typfel, line 623: Why is important to this after we've unlocked the session? Wouldn't this prevent us from presenting push notifications in the background?

johnxnguyen: Good catch, this doesn't need to be here.

johnxnguyen: I moved this to the the closure of `activateSession`.

Source/SessionManager/SessionManager.swift
johnxnguyen: When is it possible that we have a user session but are not logged in? When the client gets deleted remotely or the cookie expires?

johnxnguyen: So it turns out this `checkIfLoggedIn` method is important because it prevents transitioning to the authenticated state before having to delete a client. The result is the user appears to log in (they see the conversation list) but soon get logged out. After a subsequent log in they need to delete a client.

So it means that there are two places where the UI is informed that it should move to an authenticated state, one is through `sessionManagerDidChangeActiveUserSession`, and the other is through the `AuthenticationCoordinatorDelegate` method `userAuthenticationDidComplete(addedAccount:)`.

@marcomaddy1982 I know you're working on refactoring this at the moment. I'm curious though what the solution to this is. How / when would the UI be informed that the authentication flow is competed?

marcomaddy1982: @johnxnguyen first of all sorry for the delay in answering, I've missed your comment yesterday. 
Regarding the part of the `userAuthenticationDidComplete(addedAccount:)` till now nothing is changed. The reason is the following: with the refactoring we will introduce a `LoginInDelegate` which will inform the `AuthenticationCoordinator` when a client is registered through the `func clientRegistrationDidSucceed(accountId : UUID)` callback, at that point the transition to the `.authenticated` case of the `UI` is still not done cause  there is another step that will not be covered in my refactoring: the `SyncPhase`. Indeed when the `SyncPhase` is completed the `userAuthenticationDidComplete(addedAccount:)` is called.

marcomaddy1982 (approved):

johnxnguyen:
@zenkins release major

PR #1385, opened 2021-01-28
[Refactor] Remove the AccountStatus

typfel:
## What's new in this PR?

Remove the `AccountStatus` and any code related to it.

We don't need the `AccountStatus` anymore since we decided not to insert the `.reactivatedDevice` and `.usingNewDevice` system messages anymore. These system messages are not unnecessary now that we delete all account data when you lose your client, which means that you can't start using a new client or re-activate an old one.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1129


marcomaddy1982 (approved):

David-Henner (approved):

typfel:
@zenkins release major

PR #1389, opened 2021-02-03
[Refactor] App lock config sync

johnxnguyen:
## What's new in this PR?

These changes are part of a fix for a crash at launch.

### Changes

In  https://github.com/wireapp/wire-ios-data-model/pull/1135/ we replace lazy creation of `Feature` instances to eager creation. In other words, we must explicitly create a default instance for all feature configurations (right now there's only app lock).

In this PR we explicitly create the default instance for the app lock feature config and immediately request an update from the backend.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1135
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/279


Source/UserSession/ZMUserSession.swift
```diff
@@ -504,6 +504,18 @@ extension ZMUserSession: ZMSyncStateDelegate {
         managedObjectContext.performGroupedBlock { [weak self] in
             self?.notifyThirdPartyServices()
         }
+
+        syncManagedObjectContext.performGroupedBlock { [weak self] in
```
typfel, line 508: You can assume that `didFinishQuickSync` is always called on the sync context.

Source/UserSession/ZMUserSession.swift
```diff
@@ -504,6 +504,18 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
+            guard
+                let `self` = self,
+                let team = ZMUser.selfUser(in: self.syncManagedObjectContext).team
+            else {
+                return
+            }
+
+            Feature.createDefaultInstanceIfNeeded(name: .appLock, team: team, context: self.syncManagedObjectContext)
+            team.enqueueBackendRefresh(for: .appLock)
```
typfel, line 517: I think is also enqueued form the UI when the application becomes active and when the registration flow is completed. Can they be removed now?

typfel (changes requested):

Source/UserSession/ZMUserSession.swift
johnxnguyen: Yes I've removed it from the app root router.

Source/UserSession/ZMUserSession.swift
johnxnguyen: Fixed in f866d0525217ce5ad546d5170c64602869a2f531

KaterinaWire (approved):

marcomaddy1982 (approved):

johnxnguyen:
@zenkins release major

PR #1391, opened 2021-02-05
[Feature] integrate 'smoothed' active speakers to calling logic

David-Henner:
## What's new in this PR?

AVS modified the active speaker change payload. 

#### Previous Payload
```
"audio_levels": [
  {
    "userid": "3f49da1d-0d52-4696-9ef3-0dd181383e8a",
    "clientid": "24cc758f602fb1f4",
    "audio_level": 100
  }
]
```

`audio_level` referred to the **real time** audio level of the active speaker.

#### New Payload
```
"audio_levels": [
  {
    "userid": "3f49da1d-0d52-4696-9ef3-0dd181383e8a",
    "clientid": "24cc758f602fb1f4",
    "audio_level": 100,
    "audio_level_now": 100
  }
]
```

`audio_level` now refers to the audio level of the active speaker **smoothed over time.**

### Changes

- Updated `AVSActiveSpeakerChange.ActiveSpeaker` to align with avs payload
- Updated the call center's method to get a list of participants allowing the consumer to specify which kind of call participants will be returned. The kind is either `all` or `smoothedActiveSpeakers`. `all` returns all the participants in the call and flags the real time active speakers as active. `smoothedActiveSpeakers` returns only the active speakers smoothed over time.


Source/Calling/CallParticipantsKind.swift
```diff
@@ -0,0 +1,37 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public enum CallParticipantsKind {
+    case all
+    case smoothedActiveSpeakers
+}
```
johnxnguyen, line 24: Perhaps it might be more expressive to have a type like:

```swift
public enum CallParticipantKind {

  /// Applies to all participants.
  case any

  /// Filters out only active speakers louder than threshold.
  case activeSpeaker(smoothed: Bool, threshold: Int)

}
```

Then you could simplify other calls like:


```swift
func participants(ofKind kind: CallParticipantKind) -> [CallParticipant]

let allParticipants = participants(ofKind: .all)
let activeSpeakers = participants(ofKind: .activeSpeaker(smoothed: false, threshold: 60)
let smoothedActiveSpeakers = participants(ofKind: .activeSpeaker(smoothed: true, threshold: 60)


```

David-Henner: I agree with you on the idea of making it more expressive, however, the suggestion you made doesn't quite fit the scenarios we have. I'm sure the misunderstanding comes from the lack of clarity of the code, so I'd love to find something more explicit.

In essence, we only need the endpoint to provide the consumer with two distinct kinds of lists.
- A list of all the call participants. Including the real time active speakers, which are flagged with `isActive` = `true`
- A list of the smoothed active speakers only. 

The threshold can apply to both lists. 

Maybe renaming the `CallParticipantsKind` enum into `CallParticipantsListKind` would help slightly.

johnxnguyen: Ah ok, got it. The name `CallParticipantsListKind` definitely helps so I would do that.

Tests/Source/Calling/WireCallCenterV3Tests.swift
```diff
@@ -1341,19 +1356,58 @@ extension WireCallCenterV3Tests {
...
         let activeSpeakersAmount = participants.filter { $0.isActiveSpeaker }.count
-        XCTAssertTrue(activeSpeakersAmount == 1)
+        assertionBlock?(participants, activeSpeakersAmount)
+    }
+    
+    func testThatCallParticipants_LimitsActiveSpeakersCorrectly() {
+        testCallParticipants(activeSpeakerKind: .realTime, participantsKind: .all, limit: 1) {
+            participants, activeSpeakerAmount in
+            
+            XCTAssertTrue(activeSpeakerAmount == 1)
```
johnxnguyen, line 1377: Is there a particular reason to use `XCTAsserTrue`? Using `XCTAssertEquals` would provide more information when it fails.

Source/Calling/CallParticipantsKind.swift
David-Henner: done in 3b695c1

Tests/Source/Calling/WireCallCenterV3Tests.swift
David-Henner: No specific reason, updated it to `XCTAssertEqual` here b9a0c59

johnxnguyen (approved):

typfel (approved):

David-Henner:
@zenkins release major

PR #1395, opened 2021-02-09
[Fix] Opening conversation from notification after switching accounts

typfel:
### Issues

If you tap on a message notification from account A while having account  B loaded, the conversation with message wouldn't be shown after switching to account A.

### Causes

The delegate call asking to display the the message is being called before the view controller for account A has been loaded. The cause for this is two fold:

1. The call to `self.select(account)` is using the wrong completion block, `tearDownCompletion` instead of `completion`.
2. The `self.select(account)` completion block is called before we've checked that the account is logged in. 

### Solutions

1. Explicitly specify the completion block when calling `select(account)`
2. Call the completion block after the login check completes.

## Notes

I will remove the asynchronous `checkIfLoggedIn` method which makes the logic unnecessarily complex, this will happen in separate PR though.


marcomaddy1982 (approved):

marcomaddy1982:
@johnxnguyen maybe this PR is interesting for you.

marcomaddy1982:
@typfel there are some tests fails, maybe they are just flaky test.

typfel:
> @typfel there are some tests fails, maybe they are just flaky test.

Or I broke something :) I'll check.

johnxnguyen (approved):
I would be happy to see `checkIfLoggedIn` go away, so I'm curious how you're planning to do it.

typfel:
> > @typfel there are some tests fails, maybe they are just flaky test.
> 
> Or I broke something :) I'll check.

The tests which failed were trying to load an existing account without a self client. Since we don't consider a session to be logged in without a self client we never call the completion block and the tests never completes.

Now internally in the session manager it might make sense to load a session without a session so I think best way to fix this is to also call the completion handler in the case when the session is not logged in.

johnxnguyen:
> the completion block and the tests never completes.

@typfel ahh I think this is why tests on my branch also timeout.

typfel:
> I would be happy to see `checkIfLoggedIn` go away, so I'm curious how you're planning to do it.

https://github.com/wireapp/wire-ios-sync-engine/pull/1396

marcomaddy1982 (approved):

typfel:
@zenkins release patch

PR #1396, opened 2021-02-10
[Refactor] Remove the need for the async method checkIfLoggedIn

typfel:
## What's new in this PR?

Replace the async method `checkIfLoggedIn()` with a `isLoggedIn` property which is safe to call from the main queue. 

The async `checkIfLoggedIn()` existed because we needed to query the `ZMClientRegistrationStatus` if the self client exists or not, and the status object is only safe to access from the sync context. We can avoid this by using the already existing method `ZMClientRegistrationStatus.needsToRegisterClient(in:)` which allows us to query if the self client exists from any context.


marcomaddy1982:
@typfel some tests are failing.

johnxnguyen (approved):

marcomaddy1982 (approved):

typfel:
@zenkins release patch

PR #1406, opened 2021-02-23
[Refactor] active speaker state

David-Henner:
## What's new in this PR?

Updated the `CallParticipant`'s property `isActiveSpeaker: Bool` into `activeSpeakerState: ActiveSpeakerState`.

This allow the `.active` case of `ActiveSpeakerState` enum to carry an associated value for the instantaneous audio level, which should be used to pulse the participant's microphone / show the active speaker frame.

### DIFF
--- a/Source/Calling/CallParticipantsListKind.swift
+++ b/Source/Calling/CallParticipantsListKind.swift
@@ -36,7 +36,7 @@ extension CallParticipantsListKind {
         switch self {
         case .all where activeSpeaker.audioLevelNow > 0,
              .smoothedActiveSpeakers where activeSpeaker.audioLevel > 0:
-            return .active(audioLevels: activeSpeaker.audioLevels)
+            return .active(audioLevelNow: activeSpeaker.audioLevelNow)
         default:
             return .inactive
         }
--- a/Source/Calling/CallParticipantsListKind.swift
+++ b/Source/Calling/CallParticipantsListKind.swift
@@ -32,7 +32,7 @@ public enum CallParticipantsListKind {
 
 extension CallParticipantsListKind {
 
-    func stateOf(activeSpeaker: AVSActiveSpeakersChange.ActiveSpeaker) -> ActiveSpeakerState {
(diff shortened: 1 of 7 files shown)


Source/Calling/CallState.swift
```diff
@@ -146,6 +146,17 @@ public enum MicrophoneState: Int32, Codable {
     case muted = 1
 }
 
+/**
+ * The speaking activity state of a participant in the call.
+ */
+
+public enum ActiveSpeakerState: Equatable {
```
marcomaddy1982, line 153: I would just call it `SpeakerState`, I mean it could be `.active` or ` .inactive`

David-Henner: I'm just concerned it could be confused with the device's loud speaker

David-Henner: otherwise I agree that it would be better

marcomaddy1982:
@David-Henner there is a super small comment about naming that you could just ignore if you don't like the proposed change. A par that the PR for me is fine 👍 and I'll approve as well.

marcomaddy1982 (approved):

Source/Calling/CallParticipantsListKind.swift
```diff
@@ -31,13 +31,14 @@ public enum CallParticipantsListKind {
 }
 
 extension CallParticipantsListKind {
-    
-    func isActive(activeSpeaker: AVSActiveSpeakersChange.ActiveSpeaker) -> Bool {
+
+    func stateOf(activeSpeaker: AVSActiveSpeakersChange.ActiveSpeaker) -> ActiveSpeakerState {
```
johnxnguyen, line 35: As per the [Swift API design guidelines](https://swift.org/documentation/api-design-guidelines/#argument-labels) I would rename this to either:

-  `state(ofActiveSpeaker activeSpeaker:)`, or
- `stateOfActiveSpeaker(_ activeSpeaker)`

johnxnguyen (approved):

David-Henner:
@zenkins release major

PR #1407, opened 2021-02-23
[Fix] Edited message was incorrectly deleted from notification center

typfel:
### Issues

When message is edited the original message notification is deleted from the Notification Center.

### Causes

This was introduced by https://github.com/wireapp/wire-ios-sync-engine/pull/1325, where we stopped inserting notifications for edits but we didn't remove the logic for removing the original notification.

### Solutions

Stop removing the notifications from Notification Center on edits.


marcomaddy1982 (approved):

KaterinaWire (approved):

marcomaddy1982:
@zenkins release patch

PR #1408, opened 2021-02-24
[Fix] Syncing indicator not shown while on a very bad network

typfel:
### Issues

Network state is not reflecting that the app is waiting for sync to finish when on a very bad network.

### Causes

When bring the app back from the background the following happens:

1. Set network state to `synchronizing`
2. Process events
3. Set network state to `online`
4. Wait for push channel to establish
5. Set network state to `synchronizing`

Now if you are on really bad network the push channel will not be established and we won't indicate that the app is actually waiting to be synchronised.

### Solutions

Consider the app to be `synchronizing` when the push channel is not established.

### Dependencies

- [ ] https://github.com/wireapp/wire-ios-request-strategy/pull/287


johnxnguyen (approved):

David-Henner (approved):

typfel:
Tests some tests were failing because of a valid reason which I fixed in https://github.com/wireapp/wire-ios-request-strategy/pull/287

typfel:
@zenkins release patch

PR #1411, opened 2021-03-01
[Fix]: Handle gracefully when device runs out of disk space

marcomaddy1982:
### Issues

See the [issue](https://wearezeta.atlassian.net/browse/SQCORE-429)

When there was a problem to load the database in case we run out of disk space (or for any other error), the database was dropped. This means that the user was losing all the previous conversation

### Causes

Now a screen similar to the `blacklisted` or `jailbroken` is shown but with a different message. Moreover the alert contains two buttons:

- Device Settings -> Which will show the application settings
- Delete Database -> When the user will choose it another confirmation alert will be show. If the user confirms the database will be removed and the login screen will be shown.

## Dependencies

data-model: https://github.com/wireapp/wire-ios-data-model/pull/1144


Source/SessionManager/SessionManager.swift
```diff
@@ -46,6 +46,7 @@ public protocol SessionManagerDelegate: SessionActivationObserver {
                                        from selectedAccount: Account?,
                                        userSessionCanBeTornDown: @escaping () -> Void)
     func sessionManagerWillMigrateAccount(userSessionCanBeTornDown: @escaping () -> Void)
+    func sessionManagerDidFailLoadDatabase()
```
typfel, line 49:
```suggestion
    func sessionManagerDidFailToLoadDatabase()
```

KaterinaWire (approved):

Source/SessionManager/SessionManager.swift
```diff
@@ -434,6 +435,17 @@ public final class SessionManager : NSObject, SessionManagerType {
         }
     }
 
+    public func removeDatabaseFromDisk() {
+        guard let account = accountManager.selectedAccount else {
+            return
+        }
+        LocalStoreProvider.removeDatabaseFromDisk(accountIdentifier: account.userIdentifier,
```
typfel, line 442: Why do we need `LocalStoreProvider.removeDatabaseFromDisk`? Because I think `delete(account: account)` will also attempt to remove the database.

marcomaddy1982: Ok i was not sure that delete(account: account) will also attempts to remove the database. At this point we don't need `removeDatabaseFromDisk`

marcomaddy1982: Done

typfel (approved):

KaterinaWire (approved):

typfel (approved):

KaterinaWire (approved):

marcomaddy1982:
@zenkins release major

PR #1412, opened 2021-03-02
[fix]: SQCORE-511 Logout is not working when switch account.

marcomaddy1982:
## What's new in this PR?

for ore info about the issue see the bug ticket [here](https://wearezeta.atlassian.net/browse/SQCORE-511). Anyway this happens always, not just when the user has the maximum of the devices registered as written in the ticket.

### Causes

when we inform the `UI` with 

```swift
if session.isLoggedIn {            
      self.delegate?.sessionManagerDidReportLockChange(forSession: session)
}
```

we still didn't select the account. The account selection is done in the `completion(session)`

### Solutions

move the `completion(session)` before we inform the UI through the `delegate`. 

## Notes

I have tried many scenarios and it seems to work. Anyway it would be nice if who is reviewing this PR will think a bit if there is some edge cases on which this solution doesn't work.

### DIFF
--- a/Source/SessionManager/SessionManager.swift
+++ b/Source/SessionManager/SessionManager.swift
@@ -618,14 +618,14 @@ public final class SessionManager : NSObject, SessionManagerType {
             self.delegate?.sessionManagerDidChangeActiveUserSession(userSession: session)
             self.configureUserNotifications()
 
+            completion(session)
+            
             // If the user isn't logged in it's because they still need
             // to complete the login flow, which will be handle elsewhere.
             if session.isLoggedIn {            
                 self.delegate?.sessionManagerDidReportLockChange(forSession: session)
                 self.performPostUnlockActionsIfPossible(for: session)
             }
-            
-            completion(session)
         }
     }
 


Source/SessionManager/SessionManager.swift
```diff
@@ -618,14 +618,14 @@ public final class SessionManager : NSObject, SessionManagerType {
             self.delegate?.sessionManagerDidChangeActiveUserSession(userSession: session)
             self.configureUserNotifications()
 
+            completion(session)
+
```
johnxnguyen, line 622: It quite important that this is called before we report the database lock change, but it's not obvious at all. I think it has moved around a bit recently, also because it's reasonable to expect a completion block to be executed after all other code.

Do we always select the account in this closure? Could we explicitly select the account in this method?

johnxnguyen: In any case, I think is where the completion block was originally called (according to the git history in this file a few commits back), so it's probably the right place to call it.

marcomaddy1982: the completion is one of the `loadSession`  method for this particular scenario:

```swift
self?.loadSession(for: account) { [weak self] session in
                    self?.isSelectingAccount = false
                    
                    if let session = session {
                        self?.accountManager.select(account)
                        completion?(session)
                    }
                }
```

I don't think we can move the selection inside cause in the other case we load the session we do something like:

```swift
loadSession(for: account) { [weak self] session in
            guard let `self` = self, let session = session else { return }
            self.updateCurrentAccount(in: session.managedObjectContext)
            session.application(self.application, didFinishLaunching: launchOptions)
        }
```

Anyway I don't know if I'm missing something.

marcomaddy1982:
@johnxnguyen anyway there are some test failing and honestly i don't know if they are related to this change. It could be.

Source/SessionManager/SessionManager.swift
marcomaddy1982: @johnxnguyen anyway there are some test failing and honestly i don't know if they are related to this change. So maybe I am missing something :-)

johnxnguyen: I guess the question is whether activating a session is synonymous with activating an account. My first thought is yes, but it would need some consideration.

johnxnguyen: So is the freeze because when we tap "log out" we're activating the previous account and try to load the UI for that account, but previous account isn't selected yet?

marcomaddy1982: Exactly. In the AppRootRouter the guard of this method return cause `SessionManager.shared?.accountManager.selectedAccount` is `nil`.

```swift
private func showAuthenticated(isComingFromRegistration: Bool, completion: @escaping () -> Void) {
        guard
            let selectedAccount = SessionManager.shared?.accountManager.selectedAccount,
            let authenticatedRouter = buildAuthenticatedRouter(account: selectedAccount,
                                                               isComingFromRegistration: isComingFromRegistration)
        else {
            return
        }

        self.authenticatedRouter = authenticatedRouter

        rootViewController.set(childViewController: authenticatedRouter.viewController,
                               completion: completion)
    } 
```

johnxnguyen: We should try to make this an impossible state; if we are trying to show ui for the authenticated app state then we should have a selected account. It's probably out of scope of this PR, but I think it's worth investigating.

marcomaddy1982: Maybe tomorrow @typfel know something more about. Anyway I continue too.

typfel: Right, I guess I broke this when trying to fix opening å conversation via push notification.

We definitely need to clean up this logic it's impossible to reason about it is today. I'll create. refactoring ticket.

typfel (approved):

typfel:
Failing tests looked unrelated. I restarted the tests.

johnxnguyen (approved):

marcomaddy1982:
@zenkins release patch

PR #1413, opened 2021-03-02
[Feature] mute when answering call in group conversation

David-Henner:
## What's new in this PR?

- mute user when answering call in group conversation

### DIFF
--- a/Source/Calling/WireCallCenterV3.swift
+++ b/Source/Calling/WireCallCenterV3.swift
@@ -422,7 +422,11 @@ extension WireCallCenterV3 {
             if previousSnapshot != nil {
                 callSnapshots[conversationId] = previousSnapshot!.update(with: callState)
             }
-            
+
+            if conversation.conversationType == .group {
+                muted = true
+            }
+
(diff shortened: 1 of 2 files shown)


Source/Calling/WireCallCenterV3.swift
```diff
@@ -422,7 +422,11 @@ extension WireCallCenterV3 {
             if previousSnapshot != nil {
                 callSnapshots[conversationId] = previousSnapshot!.update(with: callState)
             }
-            
+
+            if conversation.conversationType == .group {
+                muted = true
```
typfel, line 427: Have you tried this with CallKit enabled? Will it pick up this change automatically?

typfel (approved):

Source/Calling/WireCallCenterV3.swift
David-Henner: Good point! I was concerned CallKit would automatically unmute after the call gets established but testing showed it worked as expected

marcomaddy1982 (approved):

PR #1420, opened 2021-03-11
[Fix] Tests crashing due to not properly tearing down the user session

typfel:
### Issues

When running the tests with the Thread Sanitizer enabled, a few tests were crashing in `SendAndReceiveMessagesTests+Swift.swift`.

### Causes

These tests were tearing down the session manager & user session in an attempt to simulate an interruption to the notificaion stream due to not using the client for and extended period of time. The user session was however not properly torn down which caused a crash.

### Solutions

Re-write the tests in a way which don't require tearing down the session manager and the user session.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1148


marcomaddy1982 (approved):

typfel:
Also pending https://github.com/wireapp/wire-ios-sync-engine/pull/1426

KaterinaWire (approved):

PR #1421, opened 2021-03-12
[Fix] Close the event persistence store when the EventProcessor deinits

typfel:
### Issues

Tests were logging the data store was deleted while Core Data was having an open file descriptor attached to it.

### Causes

We weren't closing the persistence store when shutting down a user session. This can lead to data loss as we experienced in https://github.com/wireapp/wire-ios-data-model/pull/1071

### Solutions

Close the persistence store on deinit.

## Notes

See also https://github.com/wireapp/wire-ios-data-model/pull/1149


David-Henner (approved):

johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1431, opened 2021-04-16
[Fix] Crash after tearing down user session quickly

typfel:
### Issues

App sometimes crash if the user session is shutdown quickly after it started.

### Causes

When the user session is started we asynchronously update the push token. If the user session shut down before this asynchronous call is performed `setPushKitToken()` will be called on the torn down  user session where the managed object contexts have deallocated, which will cause a nil unwrapping crash.

### Solutions

Hold a weak reference to the user session so that we don't call `setPushKitToken()` if it has been shutdown.


David-Henner (approved):

johnxnguyen (approved):

typfel:
@zenkins release patch

typfel:
@zenkins release patch

PR #1433, opened 2021-04-23
[Refactor] Update for use with new core data stack

typfel:
## What's new in this PR?

After the introduction of a new core data stack in https://github.com/wireapp/wire-ios-data-model/pull/1151 we need to:

- Update code paths which loads the core data store
- Update code paths access the managed object contexts

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1151
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/301


Source/SessionManager/SessionManager.swift
```diff
@@ -417,21 +423,8 @@ public final class SessionManager : NSObject, SessionManagerType {
...
-                    guard let strongSelf = self, let userIdentifier = userIdentifier else {
-                        self?.createUnauthenticatedSession()
-                        self?.delegate?.sessionManagerDidFailToLogin(error: nil)
-                        return
-                    }
-                    let account = strongSelf.migrateAccount(with: userIdentifier)
-                    self?.selectInitialAccount(account, launchOptions: launchOptions)
-            })
+            createUnauthenticatedSession()
+            delegate?.sessionManagerDidFailToLogin(error: nil)
```
johnxnguyen, line 427: So nice to see the legacy migration gone 😄

Source/UserSession/ZMUserSession.swift
```diff
@@ -96,19 +96,19 @@ public class ZMUserSession: NSObject, ZMManagedObjectContextProvider {
...
     }
     
     public var syncManagedObjectContext: NSManagedObjectContext { // TODO jacob we don't want this to be public
-        return storeProvider.contextDirectory.syncContext
+        return coreDataStack.syncContext
     }
     
     public var searchManagedObjectContext: NSManagedObjectContext { // TODO jacob we don't want this to be public
-        return storeProvider.contextDirectory.searchContext
+        return coreDataStack.viewContext// TODO jacob
```
johnxnguyen, line 107: Todo?

Tests/Source/Synchronization/StoreUpdateEventTests.swift
```diff
@@ -21,125 +21,119 @@ import WireTesting
...
-        
-        // then
-        XCTAssertEqual(batch.count, 3)
-        let loadedObjectIDs = Set(batch.map { $0.objectID.uriRepresentation() })
-        
-        XCTAssertEqual(objectIDs, loadedObjectIDs)
-        batch.forEach{ XCTAssertFalse($0.isFault) }
-    }
-}
+// TODO jacob disabled for now
```
johnxnguyen, line 24: What's the plan here, move these tests over to data model at some point?

Source/UserSession/ZMUserSession.swift
typfel: Hmm 🤔 

It should return the searchContext and not the viewContext though.

Tests/Source/Synchronization/StoreUpdateEventTests.swift
typfel: No since we don't the migration anymore but we should make sure that we delete the legacy stores files when we detect them. I'll add TODO in https://github.com/wireapp/wire-ios-data-model/pull/1151

johnxnguyen (approved):

KaterinaWire (approved):

PR #1434, opened 2021-04-29
[Fix] call degradation logic

David-Henner:
### Issues

Sometimes an ongoing call will be terminated with the reason `.securityDegraded` although the user did not verify any participants beforehand.

### Causes

`isTrusted` evaluates to `true` when the evaluated user's clients list is empty

### Solutions

use `isVerified` instead of `isTrusted`

### DIFF
--- a/Source/Calling/CallParticipantSnapshot.swift
+++ b/Source/Calling/CallParticipantSnapshot.swift
@@ -54,10 +54,17 @@ class CallParticipantsSnapshot {
             userVerifiedMap[zmuser] = userIsVerified
 
             if userWasVerified && !userIsVerified {
-                callCenter.callDidDegrade(conversationId: conversationId, degradedUser: zmuser)
+                guard let selfUser = selfUser else { return }
+                let degradedUser = selfUser.isTrusted ? zmuser : selfUser
+                callCenter.callDidDegrade(conversationId: conversationId, degradedUser: degradedUser)
                 break
             }
         }
+    }
+
+    private var selfUser: ZMUser? {
+        guard let moc = callCenter.uiMOC else { return nil }
+        return ZMUser.selfUser(in: moc)
     }
 
     /// Worst network quality of all the participants.
--- a/Source/Calling/CallParticipantSnapshot.swift
+++ b/Source/Calling/CallParticipantSnapshot.swift
@@ -38,22 +38,22 @@ class CallParticipantsSnapshot {
 
     private var participants = [CallParticipant]() {
         didSet {
-            updateUserTrustMap()
+            updateUserVerifiedMap()
             notifyChange()
         }
     }
 
-    private var userTrustMap = [ZMUser: Bool]()
+    private var userVerifiedMap = [ZMUser: Bool]()
 
-    private func updateUserTrustMap() {
+    private func updateUserVerifiedMap() {
         for user in participants.map(\.user) {
             let zmuser = user as! ZMUser
-            let userWasTrusted = userTrustMap[zmuser] ?? false
-            let userIsTrusted = zmuser.isTrusted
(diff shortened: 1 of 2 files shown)


Source/Calling/CallParticipantSnapshot.swift
```diff
@@ -38,28 +38,35 @@ class CallParticipantsSnapshot {
...
 
-            userTrustMap[zmuser] = userIsTrusted
+            userVerifiedMap[zmuser] = userIsVerified
 
-            if userWasTrusted && !userIsTrusted {
-                callCenter.callDidDegrade(conversationId: conversationId, degradedUser: zmuser)
+            if userWasVerified && !userIsVerified {
+                guard let selfUser = selfUser else { return }
+                let degradedUser = selfUser.isTrusted ? zmuser : selfUser
+                callCenter.callDidDegrade(conversationId: conversationId, degradedUser: degradedUser)
```
johnxnguyen, line 59: Why do we check if the self user is trusted here?

David-Henner: A connection with a user could become unverified if they, or the self user, adds a new untrusted device. Here we check the self user because we want to know which user (remote or self) became untrusted.

johnxnguyen (approved):

Source/Calling/CallParticipantSnapshot.swift
typfel: Doesn't it make sense to change the `isTrusted` to return false is the user has no clients? Then you wouldn't need this extra logic for determining if it was the selfUser or the user who became untrusted.

David-Henner: I kind of agree with your point that it would be easier to update the implementation of `isTrusted`. 

However, using `isTrusted` wouldn't be right according to the specs of call degradation. The call needs to degrade when a user becomes unverified.
`isTrusted` only indicates if all the clients of the remote user are trusted, but, to check the verification, we need to consider both the remote client & the self client. This is what `isVerified` does, in addition to returning false when the remote user has no clients.

On another note, the fact that the remote users client list can be empty when it shouldn't can be problematic. It's possible that conversations will appear as unverified although the self user verified all his and the remote user's devices.

typfel: >However, using isTrusted wouldn't be right according to the specs of call degradation. The call needs to degrade when a user becomes unverified.
isTrusted only indicates if all the clients of the remote user are trusted, but, to check the verification, we need to consider both the remote client & the self client. This is what isVerified does, in addition to returning false when the remote user has no clients.

So the `participants` in `CallParticipantsSnapshot` doesn't include the self user? Otherwise this rule would be fulfilled since self user would become not trusted.

>On another note, the fact that the remote users client list can be empty when it shouldn't can be problematic. It's possible that conversations will appear as unverified although the self user verified all his and the remote user's devices.

Good point! This would of course need to be taken into account. `isTrusted` is not used in that many places though so it should be feasible.

David-Henner: > So the participants in CallParticipantsSnapshot doesn't include the self user? Otherwise this rule would be fulfilled since self user would become not trusted.

The participants include the self user, and your point makes sense, but there is still a reason for not using `isTrusted`. We need to check that both the remote user and the self user are trusted, and that is what `isVerified` does for us. 
If we use `isTrusted` only, then we might trigger a degradation scenario when unnecessary. Because the self user could have trusted only all of the remote user's devices and not their own. Now imagine the remote user adds a device and becomes untrusted, we will consider this a security degradation even though it was never fully verified in the first place

typfel: Yes makes total sense now I'm on same page :)

typfel (approved):

PR #1437, opened 2021-05-19
[Feature] Notify user to remove from conversation due to legal hold policy conflict

KaterinaWire:
## What's new in this PR?

If users don't want/support LH, they will be kicked-out from group conversations because there is a user under LH in that group, and a push notification should notify about this event.
There is an additional field `”reason”: “legalhold-policy-conflict”` in events `"conversation.member-leave"` data, which should initiate a push notification.

Needs releases with:

- [ wireapp/wire-ios-data-model/pull/1158 ] 

## Notes
https://wearezeta.atlassian.net/browse/SQSERVICES-442


Resources/Base.lproj/Push.strings
```diff
@@ -125,6 +125,7 @@
 "push.notification.member.join.self.nousername" = "You were added";
 
 "push.notification.member.leave.self" = "%1$@ removed you";
+"push.notification.member.leave.self.legalhold-policy-conflict" = "You were removed from a group, because there are users with Legal Hold";
```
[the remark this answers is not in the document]
KaterinaWire: There's an agreement with Eva to show a simplified `"You were removed"` notification

Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Localization.swift
```diff
@@ -238,6 +238,7 @@ extension LocalNotificationType {
         var senderKey = self.senderKey(sender, conversation)
         var conversationTypeKey : String? = (conversation?.conversationType != .oneOnOne) ? GroupKey : OneOnOneKey
         let conversationKey = self.conversationKey(conversation)
+        let reasonKey = reason?.stringValue
```
johnxnguyen, line 241: Is this a localization key? I think you should switch On the `reason` instead because the compiler will complain if the value changes, but it won't complain if the string value changes, which would break the implementation.

johnxnguyen (changes requested):

Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Localization.swift
KaterinaWire: I removed this key

KaterinaWire:
> A snapshot test would be great if you can add one.

@johnxnguyen I don't think we have snapshots in SE

johnxnguyen:
> @johnxnguyen I don't think we have snapshots in SE

@KaterinaWire ahah... yes indeed, wrong project!

johnxnguyen (approved):

PR #1441, opened 2021-05-24
[Feature] Handling error messages when requesting new connections

KaterinaWire:
## What's new in this PR?

BE will add `error code 412` in response to posting a new connection if it's a legal hold policy conflict.
When we receive this error code, we should notify the UI project and display an alert with a message.

### DIFF
--- a/Source/UserSession/NotificationInContext+UserSession.swift
+++ b/Source/UserSession/NotificationInContext+UserSession.swift
@@ -115,7 +115,7 @@ extension ZMConversation {
 }
 
 // MARK: - Connection limit reached
-@objc public protocol ZMConnectionLimitObserver: NSObjectProtocol {
+@objc public protocol ZMConnectionFailureObserver: NSObjectProtocol {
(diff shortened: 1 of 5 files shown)


Source/UserSession/NotificationInContext+UserSession.swift
```diff
@@ -118,23 +118,37 @@ extension ZMConversation {
 @objc public protocol ZMConnectionLimitObserver: NSObjectProtocol {
     
     func connectionLimitReached()
+    func missingConnectionLegalHoldConsent()
```
johnxnguyen, line 121: Perhaps we should rename `ZMConnectionLimitObserver` to something like `ZMConnectionFailureObserver`?

KaterinaWire: Fixed

johnxnguyen (approved):

typfel (approved):

PR #1442, opened 2021-05-25
[Feature] Legal hold simplify version

KaterinaWire:
## What's new in this PR?
This PR includes:
- new and updated clients should send `"capabilities": [ "legalhold-implicit-consent"]` to BE;
- handling error messages when requesting new connections;
- handling new errors when adding participants to a conversation;
- notify user about removing from conversation due to legal hold policy conflict.

Needs releases with:

- [ wireapp/wire-ios-data-model/pull/1162 ]


johnxnguyen (approved):

Source/Synchronization/Strategies/UserClientRequestFactory.swift
```diff
@@ -94,6 +94,16 @@ public final class UserClientRequestFactory {
         })
         return completionHandler
     }
+
+    func storeCapabilities(_ client: UserClient) -> ZMCompletionHandler {
```
David-Henner, line 98: I'd make it more explicit that this method returns a handler by renaming it into something like `storeCapabilitiesHandler`.

Source/UserSession/NotificationInContext+UserSession.swift
```diff
@@ -115,26 +115,32 @@ extension ZMConversation {
...
 // MARK: - Connection limit reached
-@objc public protocol ZMConnectionLimitObserver: NSObjectProtocol {
+@objc public protocol ZMConnectionFailureObserver: NSObjectProtocol {
     
     func connectionLimitReached()
 }
 
 
-@objcMembers public class ZMConnectionLimitNotification : NSObject {
+@objcMembers public class ZMConnectionNotification : NSObject {
```
David-Henner, line 124: should this be name `ZMConnectionFailureNotification` to be consistent with the observer defined above ?

Source/UserSession/NotificationInContext+UserSession.swift
```diff
@@ -115,26 +115,32 @@ extension ZMConversation {
...
 
-    private static let name = Notification.Name(rawValue: "ZMConnectionLimitReachedNotification")
+    public static let limitReached = Notification.Name(rawValue: "ZMConnectionLimitReachedNotification")
+    public static let missingLegalHoldConsent = Notification.Name(rawValue: "ZMConnectionMissingLegalHoldConsentNotification")
     
-    public static func addConnectionLimitObserver(_ observer: ZMConnectionLimitObserver, context: NSManagedObjectContext) -> Any {
-        return NotificationInContext.addObserver(name: self.name, context: context.notificationContext) {
+    public static func addConnectionLimitObserver(_ observer: ZMConnectionFailureObserver, context: NSManagedObjectContext) -> Any {
```
David-Henner, line 129: same here with `addConnectionLimitObserver`

Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Localization.swift
```diff
@@ -274,10 +274,16 @@ extension LocalNotificationType {
...
                 }
                 conversationTypeKey = nil
-            
-            case .participantsAdded, .participantsRemoved:
+
+            case .participantsAdded:
+                conversationTypeKey = nil // System messages don't follow the template and is missing the `group` suffix
+                senderKey = SelfKey
+
+            case .participantsRemoved:
```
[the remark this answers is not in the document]
johnxnguyen: I think the reason type is an objc enum, hence we can't have an associated type. Right @KaterinaWire ?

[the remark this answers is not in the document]
KaterinaWire: Actually @David-Henner is right, it's possible to add an associated value and I fixed that in the last commit.

Source/UserSession/NotificationInContext+UserSession.swift
KaterinaWire: But in the future, it can be used for more than just failure notifications.

Source/Synchronization/Strategies/UserClientRequestFactory.swift
KaterinaWire: Renamed

Source/UserSession/NotificationInContext+UserSession.swift
KaterinaWire: What do you mean here?

David-Henner: I mean that the name `addConnectionLimitObserver` could be a little confusing since the type of observer passed as parameter is named `ZMConnectionFailureObserver`

typfel (approved):

David-Henner (approved):

PR #1444, opened 2021-05-27
[Fix] Update capabilities for new clients

KaterinaWire:
## What's new in this PR?

- Update capabilities `"capabilities": ["legalhold-implicit-consent"]` after registration for all new clients
- Change a HotFixPatch version

### DIFF
--- a/Source/Synchronization/ZMHotFixDirectory.m
+++ b/Source/Synchronization/ZMHotFixDirectory.m
@@ -209,7 +209,7 @@ static NSString* ZMLogTag ZM_UNUSED = @"HotFix";
                      }],
                     /// We need to set implicit legalhold consent capability for the SelfClient
                     [ZMHotFixPatch
-                     patchWithVersion:@"381.0.0"
+                     patchWithVersion:@"381.1.0"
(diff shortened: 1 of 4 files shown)


Source/Synchronization/ZMHotFixDirectory.m
```diff
@@ -209,7 +209,7 @@ - (NSArray *)patches
                      }],
                     /// We need to set implicit legalhold consent capability for the SelfClient
                     [ZMHotFixPatch
-                     patchWithVersion:@"381.0.0"
+                     patchWithVersion:@"381.0.1"
```
johnxnguyen, line 212: Not sure if this is necessary but I also don't think it hurts to have it.

johnxnguyen (approved):

Source/Synchronization/ZMHotFixDirectory.m
typfel: It's not necessary.

typfel (approved):

David-Henner (approved):

PR #1447, opened 2021-06-08
Fix: remove LoginCredentials.dictionaryRepresentation @objc signature

billypchan:
## What's new in this PR?

Remove @objc signature of `LoginCredentials.dictionaryRepresentation` for the requirement of XCFramework.
Convert `ZMClientRegistrationStatus.didFailToRegisterClient` to Swift since it calls the above function.
Convert `testThatItReturns_FetchingClients_WhenReceivingAnErrorWithTooManyClients` to Swift to prove that `ZMClientRegistrationStatus.didFailToRegisterClient` call be called with Swift code.

### DIFF
--- a/Source/UserSession/ZMClientRegistrationStatus.swift
+++ b/Source/UserSession/ZMClientRegistrationStatus.swift
@@ -1,0 +1,54 @@
+//
+//  ZMClientRegistrationStatus.swift
+//  WireSyncEngine-ios
+//
+//  Created by Bill, Yiu Por Chan on 03.06.21.
+//  Copyright © 2021 Zeta Project Gmbh. All rights reserved.
+//
+
+import Foundation
+
+extension ZMClientRegistrationStatus {
+    func didFail(toRegisterClient error: NSError?) {
+        var error = error
+        ///TODO:
+//        ZMLogDebug("%@", NSStringFromSelector(#function))
+        //we should not reset login state for client registration errors
+        
+        
+        
+        if let errorCode = error?.code,
+           errorCode != ZMUserSessionErrorCode.needsPasswordToRegisterClient.rawValue && errorCode != ZMUserSessionErrorCode.needsToRegisterEmailToRegisterClient.rawValue && errorCode != ZMUserSessionErrorCode.canNotRegisterMoreClients.rawValue {
+            emailCredentials = nil
+        }
+
+        if let errorCode = error?.code,
+           errorCode == ZMUserSessionErrorCode.needsPasswordToRegisterClient.rawValue {
+            // help the user by providing the email associated with this account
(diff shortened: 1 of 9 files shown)


KaterinaWire (approved):

Source/UserSession/ZMClientRegistrationStatus.swift
```diff
@@ -0,0 +1,66 @@
...
+
+import Foundation
+
+private let zmLog = ZMSLog(tag: "ZMClientRegistrationStatus")
+
+extension ZMClientRegistrationStatus {
+    @objc(didFailToRegisterClient:)
+    public func didFail(toRegisterClient error: NSError?) {
+        var error = error
+        //we should not reset login state for client registration errors
```
David-Henner, line 27: I think this comment belongs to line 29

Source/UserSession/ZMClientRegistrationStatus.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+
+import Foundation
+
+private let zmLog = ZMSLog(tag: "ZMClientRegistrationStatus")
+
+extension ZMClientRegistrationStatus {
+    @objc(didFailToRegisterClient:)
+    public func didFail(toRegisterClient error: NSError?) {
+        var error = error
```
David-Henner, line 26: is this necessary?

Source/UserSession/ZMClientRegistrationStatus.swift
```diff
@@ -0,0 +1,66 @@
...
+        //we should not reset login state for client registration errors
+        zmLog.debug(#function)
+        
+        if let errorCode = error?.code,
+           errorCode != ZMUserSessionErrorCode.needsPasswordToRegisterClient.rawValue && errorCode != ZMUserSessionErrorCode.needsToRegisterEmailToRegisterClient.rawValue && errorCode != ZMUserSessionErrorCode.canNotRegisterMoreClients.rawValue {
```
David-Henner, line 31: I could be wrong but I think that instead of writing `if let errorCode = error?.code` in each if statement, you could use a single `guard let errorCode = error?.code else { return }` at the beginning of the method.


It would help with readability, and you could use multi-line `if`s for when you need to evaluate the equality of `errorCode` to multiple values

Source/UserSession/ZMClientRegistrationStatus.swift
billypchan: cleared

Source/UserSession/ZMClientRegistrationStatus.swift
billypchan: I fixed the optional type parameter error and unwrapping is no longer needed.

David-Henner (approved):

billypchan:
@zenkins release patch

PR #1448, opened 2021-06-08
Fix: remove @obj class and extension signatures

billypchan:
## What's new in this PR?

Remove @objc signatures which cause errors when building XCFramework.


Source/UserSession/ZMClientRegistrationStatus.m
```diff
@@ -305,39 +301,6 @@ - (void)fetchExistingSelfClientsAfterClientRegistered:(UserClient *)currentSelfC
     }
 }
 
-- (void)didFailToRegisterClient:(NSError *)error
```
typfel, line 308: Why was is necessary to convert this method to Swift?

typfel (commented):
So the reason why you need to remove all `objc` annotation and extension is because  the classes inherits from `AbstractRequestStrategy` which is annotated with `@objcMembers`?

Source/UserSession/ZMClientRegistrationStatus.m
billypchan: This charge is made here:
https://github.com/wireapp/wire-ios-sync-engine/pull/1447

I will change the base branch of the PR.

billypchan:
@typfel yes, `AbstractRequestStrategy` subclassing is not possible if we do not update the iOS version to 13.0.

Wire 2021-06-02 at 11_58 AM

Source/Synchronization/Strategies/ConversationRoleDownstreamRequestStrategy.swift
```diff
@@ -77,10 +76,7 @@ public final class ConversationRoleDownstreamRequestStrategy: AbstractRequestStr
     public var requestGenerators: [ZMRequestGenerator] {
         return [downstreamSync]
     }
-}
-
 
-extension ConversationRoleDownstreamRequestStrategy: ZMDownstreamTranscoder {
```
David-Henner, line 83: Is there a specific reason for removing the extensions?

Source/Synchronization/Strategies/TeamDownloadRequestStrategy.swift
```diff
@@ -91,7 +91,8 @@ fileprivate extension Team {
 
 /// Responsible for downloading the team which the self user belongs to during the slow sync
 /// and for updating it when processing events or when manually requested.
-public final class TeamDownloadRequestStrategy: AbstractRequestStrategy, ZMContextChangeTrackerSource {
+    
+public final class TeamDownloadRequestStrategy: AbstractRequestStrategy, ZMContextChangeTrackerSource, ZMEventConsumer, ZMSingleRequestTranscoder, ZMDownstreamTranscoder {
```
David-Henner, line 95: what's the reason for moving the methods from `TeamDownloadRequestStrategy+Events` to this file?

Source/Synchronization/Strategies/ConversationRoleDownstreamRequestStrategy.swift
billypchan: if we do not set iOS version to iOS 13, this line will give compile error. It is solved by moving the protocol extension to class definition

Source/Synchronization/Strategies/TeamDownloadRequestStrategy.swift
billypchan: same as above.

David-Henner (approved):

typfel (approved):

PR #1452, opened 2021-06-09
[Chore] Annotate tests for BSI

mythsunwind:
### Issues

BSI needs a documentation that contains security-relevant test cases which are mapped to security functions (SF) and interfaces (TSFI). Additionally we need to deliver a document for them which contains all test cases containing purpose, procedure, expected result and actual result.

### Solutions

This PR annotates unit/integration test cases with a comment in front of each test case function. This helps to automatically gather these test cases and create a document out of it.

## Dependencies

There are additional pull requests in wire-ios and wire-ios-data-models like this.

## Notes

See https://github.com/wireapp/bsi-mapping/ for the scripts that do the parsing and creation of the documents.


zenkins:

typfel (approved):

PR #1453, opened 2021-06-13
[Feature] Process the url action to join the conversation

KaterinaWire:
## What's new in this PR?

Make a request and handle the result for the url action to join the conversation.


Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
```diff
@@ -34,11 +34,22 @@ class DeepLinkURLActionProcessor: URLActionProcessor {
...
+                ZMConversation.join(key: key,
+                                    code: code,
+                                    userSession: self.userSession,
+                                    managedObjectContext: self.uiMOC) { (result, conversation) in
+                    switch result {
+                    case .success:
+                        guard let conversation = conversation else {
+                            return
+                        }
+                        delegate?.showConversation(conversation, at: nil)
```
johnxnguyen, line 47: Passing in the conversation as a `Result<ZMConversation>` will make this safer and clearer to read. For instance, it means that we would never have the success case without a reference to the conversation, making the awkward   guard here unnecessary.

Source/Data Model/Conversation+Join.swift
```diff
@@ -76,7 +76,7 @@ extension ZMConversation {
                     return completion(.failure(ConversationJoinError.unknown), nil)
                 }
 
-                syncMOC.performGroupedBlock {
+                syncMOC.performGroupedBlockAndWait {
```
typfel, line 79: Why does this need to be a `blockAndWait`?

typfel (changes requested):

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
KaterinaWire: Fixed

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
```diff
@@ -20,26 +20,37 @@ import Foundation
...
+         eventProcessor: UpdateEventProcessor) {
+        self.contextProvider = contextProvider
+        self.transportSession = transportSession
+        self.eventProcessor = eventProcessor
     }
     
     func process(urlAction: URLAction, delegate: PresentationDelegate?) {
         switch urlAction {
         case let .joinConversation(key: key, code: code):
-            delegate?.shouldPerformAction(urlAction) { shouldJoin in
```
KaterinaWire, line 34: We don't need it, because shouldPerformAction is triggered by the UI project (Also checked on the simulator).

Source/Data Model/Conversation+Join.swift
KaterinaWire: We don't need it, I removed it.

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
```diff
@@ -20,26 +20,37 @@ import Foundation
...
+                case .success(let conversation):
+                    delegate?.showConversation(conversation, at: nil)
+                case .failure(let error) :
+                    delegate?.failedToPerformAction(urlAction, error: error)
+                }
+                delegate?.completedURLAction(urlAction)
             }
             
         case .openConversation(let id):
+            let uiMOC = contextProvider.viewContext
```
johnxnguyen, line 53: It's nice to be consistent with naming, so I'd suggest `viewContext`.

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
```diff
@@ -49,6 +60,7 @@ class DeepLinkURLActionProcessor: URLActionProcessor {
             delegate?.completedURLAction(urlAction)
 
         case .openUserProfile(let id):
+            let uiMOC = contextProvider.viewContext
```
johnxnguyen, line 63: Suggestion: `viewContext`

johnxnguyen (commented):
Are there tests that the delegate methods are called?

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
KaterinaWire: Done

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
KaterinaWire: Done

KaterinaWire:
> Are there tests that the delegate methods are called?

Tests were added

typfel (approved):

Tests/Source/UserSession/DeepLinkURLActionProcessorTests.swift
```diff
@@ -94,5 +102,32 @@ class DeepLinkURLActionProcessorTests: DatabaseTest {
...
+        // given
+        let action: URLAction = .joinConversation(key: "test-key", code: "test-code")
+
+        // when
+        sut.process(urlAction: action, delegate: presentationDelegate)
+        XCTAssertTrue(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
+
+        // then
+        XCTAssertEqual(presentationDelegate.completedURLActionCalls.count, 1)
+        XCTAssertEqual(presentationDelegate.completedURLActionCalls.first, action)
```
johnxnguyen, line 116: To align with the test name, shouldn't we assert with `shouldPerformActionCalls`?

johnxnguyen: Also, do we want to assert that it asks to show the conversation?

KaterinaWire: We can't ask to assert `showConversationCalls`, because there is a `mockUpdateEventProcessor` that doesn't process events and there's no conversation as a result. I just added one assertion `XCTAssertEqual(mockUpdateEventProcessor.processedEvents.count, 1)`  to make sure we are accepting the event.

KaterinaWire: > To align with the test name, shouldn't we assert with `shouldPerformActionCalls`?

I chose this name to be consistent with test names in this file. These test only check the processing url actions. If you have suggestions to rename it, we can apply it.
As for `shouldPerformActionCalls`, we have a test class `SessionManagerTests_URLActions`, that checks `shouldPerformActionCalls`. But this class uses the connectService url action as an example.

johnxnguyen: `testThatItCompletesTheJoinConversationAction_WhenCodeIsValid` and `testThatItReportsTheJoinConversationActionAsFailed_WhenCodeIsInvalid`

KaterinaWire: Renamed

johnxnguyen (approved):

PR #1454, opened 2021-06-15
[Feature] Add a new endpoint GET conversation/join

KaterinaWire:
## What's new in this PR?

Add an endpoint that we can use to fetch the conversation with the code and key. 
We need this endpoint when a user wants to join the conversation via a deep link and is already a participant in the conversation.


Source/Data Model/Conversation+Join.swift
```diff
@@ -87,6 +105,42 @@ extension ZMConversation {
...
+
+        request.add(ZMCompletionHandler(on: managedObjectContext, block: { response in
+            switch response.httpStatus {
+            case 200:
+                guard let payload = response.payload as? [AnyHashable : Any],
+                      let conversationString = payload["conversation"] as? String,
+                      let conversationId = UUID(uuidString: conversationString),
+                      let conversation = ZMConversation(remoteID: conversationId, createIfNeeded: false, in: managedObjectContext) else {
+                    return  completion(.failure(ConversationFetchError.unknown))
+                }
```
johnxnguyen, line 131: What does this endpoint return exactly? The tech specs just say the conversation id and name, but it doesn't say much more than that.

Currently we're just fetching the id and using that to see if we have a local instance of that conversation. But I would only expect that have that instance (and therefore return a success) if the user is or ever was a part the conversation. 

In the case that this is a completely unknown conversation, then we wouldn't be able to fetch it locally and therefore would return the ` unknown` error. Is this intentional?

Source/Data Model/Conversation+Join.swift
```diff
@@ -101,4 +155,14 @@ internal struct ConversationJoinRequestFactory {
         return ZMTransportRequest(path: path, method: .methodPOST, payload: payload as ZMTransportData)
     }
 
+    static func requestForGetConversation(key: String, code: String) -> ZMTransportRequest {
+        let path = "/conversations/join"
+        let payload: [String: Any] = [
+            "key": key,
+            "code": code
+        ]
```
johnxnguyen, line 163: The tech specs state that these should be passed as query parameters.

KaterinaWire: Yes, you are right. 
Fixed it

Source/Data Model/Conversation+Join.swift
```diff
@@ -73,10 +87,14 @@ extension ZMConversation {
...
+            /// The user is already a participant in the conversation
             case 204:
-                /// TODO It's a temporary solution. The new endpoint will be introduced in the next PR, which we should call in this case and get the conversation
-                completion(.failure(ConversationJoinError.unknown))
+                fetch(key: key,
+                      code: code,
+                      transportSession: transportSession,
+                      managedObjectContext: viewContext) { result in
+                    completion(result)
+                }
```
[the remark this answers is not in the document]
KaterinaWire: Updated the logic

Source/Data Model/Conversation+Join.swift
```diff
@@ -87,6 +105,42 @@ extension ZMConversation {
         transportSession.enqueueOneTime(request)
     }
 
+    /// Fetch a conversation using a reusable code
+    /// - Parameters:
+    ///   - key: stable conversation identifier
+    ///   - code: conversation code
+    ///   - transportSession: session to handle requests
+    ///   - managedObjectContext: current managedObjectContext. It should be a viewContext, because we have to handle completion in the viewContext
```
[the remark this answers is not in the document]
KaterinaWire: Fixed

Source/Data Model/Conversation+Join.swift
```diff
@@ -87,6 +105,42 @@ extension ZMConversation {
...
+    ///   - completion: a handler when the network request completes with http status code
```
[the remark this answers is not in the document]
KaterinaWire: Updated the doc

Source/Data Model/Conversation+Join.swift
KaterinaWire: Updated the logic

Source/Data Model/Conversation+Join.swift
```diff
@@ -73,10 +88,11 @@ extension ZMConversation {
                     }
                 }
 
-            /// The user is already a participant in the conversation.
+            /// The user is already a participant in the conversation
             case 204:
-                /// TODO It's a temporary solution. The new endpoint will be introduced in the next PR, which we should call in this case and get the conversation
-                completion(.failure(ConversationJoinError.unknown))
+                /// If we get to this case, then we need to re-sync local conversations
```
johnxnguyen, line 93: Is this todo tracked somewhere?

Source/Data Model/Conversation+Join.swift
```diff
@@ -101,4 +157,13 @@ internal struct ConversationJoinRequestFactory {
         return ZMTransportRequest(path: path, method: .methodPOST, payload: payload as ZMTransportData)
     }
 
+    static func requestForGetConversation(key: String, code: String) -> ZMTransportRequest {
+        var url = URLComponents()
+        url.path = "/conversations/join"
+        url.queryItems = [URLQueryItem(name: "key", value: key), URLQueryItem(name: "code", value: code)]
+        let urlString = url.string ?? ""
```
johnxnguyen, line 164: Since the url construction is fallible, I suggest to reveal this to the caller of the method so they know how to handle the issue rather than receive a request to an empty path. You can do this by returning an optional request.

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
```diff
@@ -35,18 +35,45 @@ class DeepLinkURLActionProcessor: URLActionProcessor {
...
+                switch response {
+                case .success((let conversationId, let conversationName)):
+                    /// First of all, we should try to fetch the conversation with ID from the response.
+                    /// If the conversation doesn't exist, we should initiate a request to join the conversation
+                    if let conversation = ZMConversation(remoteID: conversationId, createIfNeeded: false, in: viewContext),
+                       conversation.isSelfAnActiveMember {
+                        delegate?.showConversation(conversation, at: nil)
+                        delegate?.completedURLAction(urlAction)
+                    } else {
+                        delegate?.shouldPerformActionWithMessage(conversationName, action: urlAction) { shouldJoin in
```
johnxnguyen, line 55: Passing the conversation name as the message is a little misleading because it's not being used as the message, but rather only as part of the message. The intent here is that you need to pass the conversation name to the delegate so that it can present a confirmation alert using the name. To make this intent clearer, you could change the method signature to

```swift
func shouldPerformAction(for conversation: ZMConversation: action: URLAction, decisionHandler: (Bool) -> Void)
```

In this way, you leave the construction of the confirmation message to the delegate. What do you think?

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
```diff
@@ -35,18 +35,45 @@ class DeepLinkURLActionProcessor: URLActionProcessor {
...
+                            guard shouldJoin else { return }
```
johnxnguyen, line 56: Does it make sense to also call the completion here?

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
```diff
@@ -35,18 +35,45 @@ class DeepLinkURLActionProcessor: URLActionProcessor {
...
-                switch result {
-                case .success(let conversation):
-                    delegate?.showConversation(conversation, at: nil)
-                case .failure(let error) :
+            ZMConversation.fetchIdAndName(key: key,
+                                          code: code,
+                                          transportSession: transportSession,
+                                          eventProcessor: eventProcessor,
+                                          contextProvider: contextProvider) { [weak self] (response) in
+                guard let strongSelf = self else { return }
```
johnxnguyen, line 43: Since this method depends on there being a delegate, I suggest to also unwrap it here to avoid so many `?`.

Tests/Source/Integration/ConversationTests+Join.swift
```diff
@@ -71,39 +73,102 @@ class ConversationTests_Join: ConversationTestsBase {
...
+                                      code: "existing-conversation-code",
+                                      transportSession: userSession!.transportSession,
+                                      eventProcessor: userSession!.updateEventProcessor!,
+                                      contextProvider: userSession!.coreDataStack) { (result) in
+            // THEN
+            if case .success((let conversationID, let conversationName)) = result {
+                XCTAssertNotNil(conversationID)
+                XCTAssertNotNil(conversationName)
+                let conversation = ZMConversation.fetch(withRemoteIdentifier: conversationID, in: viewContext)
+                XCTAssertTrue(conversation!.isSelfConversation)
```
johnxnguyen, line 119: What is the relevance of `isSelfConversation`?

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
KaterinaWire: At that moment we have no conversation in the local database (there may be an archived conversation, but this case is rather an exception). We call this method when we receive information that there is no conversation, and we should trigger a `joinConvesation` request. The `joinConvesation` request will receive an update event and create a conversation.

Source/Data Model/Conversation+Join.swift
KaterinaWire: Done

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
KaterinaWire: Yes.
Done

Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
KaterinaWire: Done

Tests/Source/Integration/ConversationTests+Join.swift
KaterinaWire: It's a wrong property.
It should be `isSelfAnActiveMember`.
I wanted to check if the conversation exists and selfUser is participant.

Source/Data Model/Conversation+Join.swift
KaterinaWire: https://wearezeta.atlassian.net/browse/SQSERVICES-533

typfel (approved):

johnxnguyen (approved):

PR #1461, opened 2021-06-22
[Fix] Perform the request from viewContext

KaterinaWire:
### Issues
The crash in the UI when receiving an error message.

### Causes
`completion(.failure(error))` was called from syncContext. The completion handler triggers a delegate method that displays a popup with an error message.

### Solutions
Perform the request from `viewContext`.


johnxnguyen (approved):
Just update the doc comment to:

`completion: called on the main thread when the user joins the conversation or when it fails.`

typfel (approved):

KaterinaWire:
@zenkins release patch

PR #1465, opened 2021-06-30
[Feature] Integrate AVS selective video endpoint

David-Henner:
## What's new in this PR?

AVS selective video allows to specify which video streams should be loaded and made available by AVS.
This is helpful in the context of large conference calls, where we only want to receive video streams for participants that are visible on screen.

### Changes

- Added support for the avs selective video endpoint. 
- Restructured AVS support files. Each struct gets its own file, which helps with searching and declutters the files.
- Updated tests


Source/Calling/AVS/AVSClient.swift
```diff
@@ -0,0 +1,55 @@
...
+
+        self.init(userId: userId, clientId: clientId)
+    }
+
+}
+
+extension AVSClient: Codable {
+
+    enum CodingKeys: String, CodingKey {
+
```
KaterinaWire, line 49: Just a small note, in `CodingKeys` enum you didn't make an empty line before `case`.

Source/Calling/AVS/AVSWrapper.swift
```diff
@@ -186,6 +187,11 @@ public class AVSWrapper: AVSWrapperType {
         wcall_config_update(handle, httpStatusCode == 200 ? 0 : EPROTO, callConfig ?? "")
     }
 
+    /// Requests AVS to load video streams for the given clients list
+    public func requestVideoStreams(_ videoStreams: AVSVideoStreams, conversationId: UUID) {
```
KaterinaWire, line 191: According to the comment, we should provide a clients list, but we don't have such a parameter. Am I wrong?

David-Henner: Good catch! At first the parameter was a clients list but I've changed it and forgot to update the comment. I'll update.

Source/Calling/AVS/JSONStringEncodable.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+import Foundation
+
+protocol JSONStringEncodable: Encodable { }
+
+extension JSONStringEncodable {
+    var json: String? {
+        guard let data = try? JSONEncoder().encode(self) else { return nil }
+        return String(data: data, encoding: .utf8)
+    }
```
typfel, line 26: You are creating `JSONEncoder` instance every time you are encoding to JSON here. We have similar use case here https://github.com/wireapp/wire-ios-request-strategy/blob/f0458055a360b978bc4f00489dc12ad180f0c12f/Sources/Payloads/Payload%2BDecoding.swift#L77

You could adapt the same strategy here and do something like:
```
func jsonString(encoder: JSONEncoder = .init()) -> String? {
    guard let data = try? encoder.encode(self) else { return nil }
    return String(data: data, encoding: .utf8)
}
```

You could then create the JSONEncoder and store it in the `AVSWrapper` and supply it when doing the encoding. In the tests you could rely on the default encoder for convenience.

Tests/Source/Calling/AVSVideoStreamsTest.swift
```diff
@@ -0,0 +1,49 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import XCTest
+@testable import WireSyncEngine
+
+class AVSVideoStreamsTest: XCTestCase {
+    func testThat_JsonStringValue_IsConformTo_avsAPI() {
```
typfel, line 24:
```suggestion
    func testThatJSONStringValue_ConformToAVSAPI() {
```

typfel (changes requested):

Source/Calling/AVS/AVSWrapper.swift
```diff
@@ -187,9 +188,12 @@ public class AVSWrapper: AVSWrapperType {
         wcall_config_update(handle, httpStatusCode == 200 ? 0 : EPROTO, callConfig ?? "")
     }
 
-    /// Requests AVS to load video streams for the given clients list
+    /// Requests AVS to load at list of video streams
```
KaterinaWire, line 191: Requests AVS to load a list of video streams

KaterinaWire (approved):

typfel (approved):

David-Henner:
@zenkins release major

PR #1492, opened 2021-07-23
[Refactor] Feature configs

johnxnguyen:
## What's new in this PR?

See https://github.com/wireapp/wire-ios-data-model/pull/1183 for a more thorough overview of the refactoring.

In this PR we...
- combine separate app lock fields in the `SessionManagerConfiguration` into a single `legacyAppLockConfig`. This is optional, and if present, will override any config fetched from the backend.
- make use of the `FeatureService` to refresh existing configs.
- update tests.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1183
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/331


Tests/Source/Utility/ZMUserSession+DefaultConfiguration.swift
```diff
@@ -20,13 +20,6 @@ import Foundation
...
     static var defaultConfig: ZMUserSession.Configuration {
-        Self.init(
-            appLockConfig: .init(
-                isAvailable: true,
-                isForced: false,
-                timeout: 19,
-                requireCustomPasscode: false
-            )
-        )
+        Self.init(appLockConfig: nil)
```
KaterinaWire, line 23: Do we need it if it's `nil`?

johnxnguyen: Are you referring to the default config, or the user session config?

KaterinaWire: default config

johnxnguyen: If we provide the default value in the initializer, then we don't need this extension.

johnxnguyen: Done

KaterinaWire (approved):

David-Henner (approved):

johnxnguyen:
@zenkins release major

PR #1512, opened 2021-08-27
[Chore] Don't embed the Starscream framework

typfel:
## What's new in this PR?

Don't embed the Starscream framework since it will be embedded at the top level.


johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1516, opened 2021-09-01
[Feature] Pre-call actions

David-Henner:
## What's new in this PR?

This PR updates the behaviour when answering a call & when handling an incoming call.

### Changes
- Remove setting video state when answering a call
- Mute microphone for group calls when handling an incoming call instead of when answering


KaterinaWire (approved):

billypchan (approved):

David-Henner:
@zenkins release minor

PR #1519, opened 2021-09-09
[Feature] Add internal usePackagingFeatureConfig flag

KaterinaWire:
### Issues
We need to cover the upcoming packaging feature with an internal flag so that we can silently release it next time.

### Solutions
Add the `usePackagingFeatureConfig` property to the `WireCallCenterV3` class and set it from the UI project.


johnxnguyen (approved):

johnxnguyen:
You might need to put some of the new tests (for starting calls) behind this flag too.

KaterinaWire:
> You might need to put some of the new tests (for starting calls) behind this flag too.

I added a test

Tests/Source/Calling/WireCallCenterV3Tests.swift
```diff
@@ -648,6 +649,20 @@ class WireCallCenterV3Tests: MessagingTest {
         // then
         XCTAssertNil(mockAVSWrapper.startCallArguments)
     }
+
+    func testThatItStartsAConferenceCall_IfConferenceCallingFeatureIsDisabledByInternalFlag(){
```
David-Henner, line 653: The name is a bit confusing to me, it sounds like the conference calling feature is disabled but we should still be able to start a conference call? 
Should it be something like `testThatItStartsAConferenceCall_IfPackagingFeatureIsDisabledByInternalFlag()` instead?

KaterinaWire: Yes, fixed

David-Henner (approved):

PR #1521, opened 2021-09-10
[Fix] Failed tests that were affected by adding new requests to QuickSync

KaterinaWire:
### Issues
Failed `testThatItFetchesLastNotificationsFromBackendIgnoringTransientNotificationsID` test. 

### Causes
Requests to fetch feature configs have been added to QuickSync.

### Solutions
Exclude the fetch feature config requests from the expected result, as the test checks if there are transient notification in the given result.

### DIFF
--- a/Tests/Source/Integration/PushChannelTests.swift
+++ b/Tests/Source/Integration/PushChannelTests.swift
@@ -116,9 +116,10 @@ class PushChannelTests: IntegrationTest {
         _ = waitForAllGroupsToBeEmpty(withTimeout: 0.5)
         
         // THEN
-        let expectedLastRequest = "/notifications?size=\(ZMMissingUpdateEventsTranscoderListPageSize)&since=\(messageAddLastNotificationID?.transportString() ?? "")&client=\(userSession?.selfUserClient?.remoteIdentifier ?? "")"
+        let expectedRequest = "/notifications?size=\(ZMMissingUpdateEventsTranscoderListPageSize)&since=\(messageAddLastNotificationID?.transportString() ?? "")&client=\(userSession?.selfUserClient?.remoteIdentifier ?? "")"
 
-        XCTAssertTrue(mockTransportSession.receivedRequests().contains(where: { $0.path == expectedLastRequest } ))
+        let targetRequests = mockTransportSession.receivedRequests().filter { $0.path == expectedRequest }
+        XCTAssertEqual(targetRequests.count, 1)
     }
 
 }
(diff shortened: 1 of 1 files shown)


johnxnguyen (commented):
Is this a new issue? If not I'm wondering why it wasn't a problem before with the app lock and file sharing configs.

Tests/Source/Integration/PushChannelTests.swift
```diff
@@ -117,6 +117,18 @@ class PushChannelTests: IntegrationTest {
         
         // THEN
         let expectedLastRequest = "/notifications?size=\(ZMMissingUpdateEventsTranscoderListPageSize)&since=\(messageAddLastNotificationID?.transportString() ?? "")&client=\(userSession?.selfUserClient?.remoteIdentifier ?? "")"
```
johnxnguyen, line 119: Does this expected request necessarily have to be the last request? Perhaps its mere existence is sufficient, in which case you can avoid filtering the requests and simply search for this specific one.

KaterinaWire:
> Is this a new issue? If not I'm wondering why it wasn't a problem before with the app lock and file sharing configs.

Because we used different endpoints with team ID and this test creates personal users.

Tests/Source/Integration/PushChannelTests.swift
KaterinaWire: I guess this test checks that the last request doesn't contain a transient notification ID. If we filter the expected result, we cannot be sure that the last request will not be with a transient notification ID.

typfel: We only expect one request to `/notifications` and in that request we assert that the ID is not a transient ID. So IMO is should be OK and more stable to filter out the `/notifications` request from the list of received requests.

KaterinaWire: Ok, fixed with:
`XCTAssertTrue(mockTransportSession.receivedRequests().contains(where: { $0.path == expectedLastRequest } ))`

johnxnguyen: This would still pass even if there were two requests, so you could filter it and assert that there is only one result.

typfel (approved):

Tests/Source/Integration/PushChannelTests.swift
KaterinaWire: Fixed

johnxnguyen (approved):

PR #1559, opened 2021-11-05
Integrate SwiftLint into iOS Sync Engine Project

agisilaos:
## What's new in this PR?

Integrating SwiftLint into the iOS Sync Engine project. The `.swiftlint.yml` file is based on the file that's been added on the `wire-ios` repository. I've also added a GitHub action for SwiftLint 

## Notes

https://wearezeta.atlassian.net/jira/software/projects/IC/boards/100?selectedIssue=IC-6

### DIFF
--- a/.github/workflows/swiftlint.yml
+++ b/.github/workflows/swiftlint.yml
@@ -1,0 +1,17 @@
+name: SwiftLint
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/swiftlint.yml'
+      - '.swiftlint.yml'
+      - '**/*.swift'
+
+jobs:
+  SwiftLint:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v1
+        - name: GitHub Action for SwiftLint
+          uses: norio-nomura/action-swiftlint@3.2.1
+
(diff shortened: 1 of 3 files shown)


.github/workflows/swiftlint.yml
```diff
@@ -0,0 +1,17 @@
+name: SwiftLint
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/swiftlint.yml'
+      - '.swiftlint.yml'
```
billypchan, line 7: can we just keep one copy of rules?

billypchan:
the branch name should be standardise like `fix/integrate-swiftlint-into-the-sync-engine-project`, and it would be better to create a jira ticket for a record.

billypchan:
I would suggest create a demo PR to check how does the Github action works.

agisilaos:
> the branch name should be standardise like fix/integrate-swiftlint-into-the-sync-engine-project, and it would be better to create a jira ticket for a record.

There's already a Jira ticket for the issue.

.github/workflows/swiftlint.yml
agisilaos: This is how it's been set up in the `wire-ios` repo.

agisilaos:
Closing this one to make the appropriate changes @billypchan made regarding the name of the branch etc and will open a new one.

PR #1575, opened 2021-11-16
[Chore] Revert changes

KaterinaWire:
## What's new in this PR?

Another attempt to fix the crash. We currently have weird behavior after killing the app and I want to try to revert these changes.


Source/UserSession/ZMUserSession.swift
```diff
@@ -538,18 +538,15 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
         }
 
-        syncContext.performGroupedBlock {
-            let featureService = FeatureService(context: self.syncContext)
-            featureService.enqueueBackendRefresh(for: .appLock)
-            featureService.enqueueBackendRefresh(for: .fileSharing)
-            featureService.enqueueBackendRefresh(for: .conferenceCalling)
-            featureService.enqueueBackendRefresh(for: .selfDeletingMessages)
-        }
+        featureService.enqueueBackendRefresh(for: .appLock)
```
johnxnguyen, line 546: Which context is passed to this instance? Doesn't it fail if it's the view content, because this method mutates the feature entity?

KaterinaWire: `public lazy var featureService = FeatureService(context: syncContext)`

johnxnguyen (approved):

sb88k (approved):

KaterinaWire:
release patch

PR #1597, opened 2021-11-30
fix: Request loop when get conversation roles #FS-303

sb88k:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-303" title="FS-303" target="_blank">BugFS-303</a>  iOS: Send debug logs alert shows on receiving a message with mention in a federated group 
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Because there is no endpoint for fetching conversation roles in a federated conversation, the request is being triggered again and again.

By setting `needsToDownloadRoles` to false to prevent the request loop.


Source/Synchronization/Strategies/ConversationRoleDownstreamRequestStrategy.swift
```diff
@@ -90,13 +90,26 @@ public final class ConversationRoleDownstreamRequestStrategy: AbstractRequestStr
...
     public func delete(_ object: ZMManagedObject!, with response: ZMTransportResponse!, downstreamSync: ZMObjectSync!) {
-        // do not delete conversation
+        // Do not delete conversation and set needsToDownloadRoles to false to avoid request loop
+        guard
+            downstreamSync as? ZMDownstreamObjectSync == self.downstreamSync,
+            let conversation = object as? ZMConversation,
+            response.httpStatus == 404
+        else {
+            return
+        }
```
sb88k, line 100: @typfel 
Not sure if all these checks are needed. Any idea?

sb88k: After having this logic, it is still happening. Maybe we need to fix where it sets the flag to true.

sb88k: From my investigation, I found that the flag is being set to true everytime when the user enters the conversation screen.

Not an usual case that the user enters and leaves the conversation screen for 20 times.
so, maybe we need to ask QA to adjust the test flow to not leaving and entering the conversation screen too often?

typfel: I see, then I guess it also makes sense to update `markToDownloadRolesIfNeeded()` to not update the roles if the conversation is federated.

sb88k: Also, during the testing, it tries to fetch the roles when mention is in message, and it triggers the failed requests too.
This fix prevents it triggers the failed requests, so I think this will fix the issue.

And, it contradicts the `isFederating` check in `markToDownloadRolesIfNeeded`, because `isTeamConversation` is checking if it is federated conversation.

typfel (approved):

sb88k:
@typfel 
The playground build 993007 is working fine.

David-Henner (approved):

sb88k:
release patch

sb88k:
release patch

PR #1603, opened 2021-12-01
chore: Fix SwiftLint warnings in Source Part 5

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR I fix all warnings related to the rules below:

- **Computed Accessors Order Violation**: Computed properties should declare first the getter and then the setter. `(computed_accessors_order)`
- **Implicit Getter Violation**: Computed read-only properties should avoid using the get keyword. `(implicit_getter)`
- **Returning Whitespace Violation**: Return arrow and return type should be separated by a single space or on a separate line. `(return_arrow_whitespace)`
- **Operator Function Whitespace Violation**: Operators should be surrounded by a single whitespace when defining them. `(operator_whitespace)`
- **Opening Brace Spacing Violation**: Opening braces should be preceded by a single space and on the same line as the declaration. `(opening_brace)`
- **Trailing Whitespace Violation**: Lines should not have trailing whitespace. `(trailing_whitespace)`
- **Colon Violation**: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. `(colon)`
- **Unneeded Break in Switch Violation**: Avoid using unneeded break statements. `(unneeded_break_in_switch)`
- **Redundant String Enum Value Violation**: String enum values can be omitted when they are equal to the enumcase name. `(redundant_string_enum_value)`
- **For Where Violation**: `where` clauses are preferred over a single `if` inside a `for`. `(for_where)`
- **Switch and Case Statement Alignment Violation**: Case statements should vertically align with their enclosing switch statement. `(switch_case_alignment)`
- **Duplicate Imports Violation**: Imports should be unique. `(duplicate_imports)`
**No Fallthrough Only Violation**: Fallthroughs can only be used if the `case` contains at least one other statement. `(no_fallthrough_only)`
- **Orphaned Doc Comment Violation**: A doc comment should be attached to a declaration. `(orphaned_doc_comment)`
- **Vertical Parameter Alignment Violation**: Function parameters should be aligned vertically if they're in multiple lines in a declaration. `(vertical_parameter_alignment)`
- **Unused Optional Binding Violation:** Prefer `!= nil` over `let _ =` `(unused_optional_binding)`

----
##### References


Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Localization.swift
```diff
@@ -321,7 +321,7 @@ extension String {
         case 4:
             return String.localizedStringWithFormat(format, arguments[0], arguments[1], arguments[2], arguments[3])
         case 0:
-            fallthrough
+            return NSLocalizedString(format, comment: "")
```
agisilaos, line 324: In this case it expected a string that's why I decided to go with return NSLocalizedString(format, comment: "")

Source/Utility/ZMSound.swift
```diff
@@ -72,7 +72,7 @@ public enum ZMSound: String, CustomStringConvertible {
     fileprivate static var playingPreviewURL: URL?
 
     fileprivate static func stopPlayingPreview() {
-        if let _ = self.playingPreviewURL,
+        if self.playingPreviewURL != nil ,
```
agisilaos, line 75: This change is regarding the rule: `Unused Optional Binding Violation: Prefer `!= nil` over `let _ =` (unused_optional_binding)`. Please let me know if that's the right approach.

Tests/Source/Synchronization/StoreUpdateEventTests.swift
```diff
@@ -156,9 +156,9 @@ class StoreUpdateEventTests: MessagingTest {
         let payload = payloadForMessage(in: conversation, type: EventConversationAdd, data: ["foo": "bar"])!
         let event = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: UUID.create())!
 
-        guard let _ = StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 0),
-            let _ = StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 1),
-            let _ = StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 2)
+        guard StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 0),
+            StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 1),
+            StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 2) != nil
```
agisilaos, line 161: This change is regarding the rule: Unused Optional Binding Violation: Prefer != niloverlet _ = (unused_optional_binding). Please let me know if that's the right approach.

Tests/Source/Integration/ConversationTests+Ephemeral.swift
```diff
@@ -182,6 +182,7 @@ extension ConversationTests_Ephemeral {
         XCTAssertEqual(ephemeral.hiddenInConversation, conversation)
         XCTAssertNil(ephemeral.sender)
 
+        // Unused Optional Binding Violation: Prefer `!= nil` over `let _ =` (unused_optional_binding)
```
agisilaos, line 185: I'm not sure how to convert that from let _ to != nil so any suggestions are more than welcome.

Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Localization.swift
billypchan: we can remove case 0 and then default would handle it.

Source/UserSession/CallStateObserver.swift
```diff
@@ -102,7 +102,7 @@ extension CallStateObserver: WireCallCenterCallStateObserver, WireCallCenterMiss
                 switch (systemMessage.systemMessageType, callState, conversation.conversationType) {
                 case (.missedCall, .terminating(reason: .canceled), _ ):
                     // the caller canceled the call
-                    fallthrough
+                    break
```
billypchan, line 105: fallthrough call the code below and it should not be change to break

billypchan (commented):
comment about fall through

Source/UserSession/CallStateObserver.swift
agisilaos: According to the rule: `Fallthroughs can only be used if the `case` contains at least one other statement` Should we disable it at this case? @billypchan

Source/Utility/ZMSound.swift
```diff
@@ -72,7 +72,7 @@ public enum ZMSound: String, CustomStringConvertible {
     fileprivate static var playingPreviewURL: URL?
 
     fileprivate static func stopPlayingPreview() {
-        if let _ = self.playingPreviewURL,
+        if self.playingPreviewURL != nil,
```
agisilaos, line 75: This change is regarding the rule: Unused Optional Binding Violation: Prefer != niloverlet _ = (unused_optional_binding). Please let me know if that's the right approach.

Tests/Source/Synchronization/StoreUpdateEventTests.swift
```diff
@@ -156,9 +156,9 @@ class StoreUpdateEventTests: MessagingTest {
         let payload = payloadForMessage(in: conversation, type: EventConversationAdd, data: ["foo": "bar"])!
         let event = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: UUID.create())!
 
-        guard let _ = StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 0),
-            let _ = StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 1),
-            let _ = StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 2)
+        guard (StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 0) != nil),
+              (StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 1) != nil),
+              (StoredUpdateEvent.encryptAndCreate(event, managedObjectContext: eventMOC, index: 2) != nil)
```
agisilaos, line 161: This change is regarding the rule: Unused Optional Binding Violation: Prefer != niloverlet _ = (unused_optional_binding). Please let me know if that's the right approach.

Tests/Source/Integration/ConversationTests+Ephemeral.swift
agisilaos: I'm thinking It's going to be like this. Open to suggestions.
``` swift
        guard (conversation.hiddenMessages.first(where: {
            if let message = $0 as? ZMClientMessage,
                    let deleteMessage = message.underlyingMessage, deleteMessage.hasDeleted,
                deleteMessage.deleted.messageID == ephemeral.nonce!.transportString() {
                return true
            }
            else {
                return false
            }
        })) != nil
        else {
            return XCTFail()
        }
```

Source/UserSession/CallStateObserver.swift
```diff
@@ -100,13 +100,13 @@ extension CallStateObserver: WireCallCenterCallStateObserver, WireCallCenterMiss
...
-                case (.missedCall, .terminating(reason: .canceled), _ ):
-                    // the caller canceled the call
-                    break
                 case (.missedCall, .terminating(reason: .normal), .group):
                     // group calls we didn't join, end with reason .normal. We should still insert a missed call in this case.
                     // since the systemMessageGenerator keeps track whether we joined or not, we can use it to decide whether we should show a missed call APNS
                     self.localNotificationDispatcher.processMissedCall(in: conversation, caller: caller)
+                case (.missedCall, .terminating(reason: .canceled), _ ):
+                    // the caller canceled the call
+                    break
```
agisilaos: @billypchan Changed the order of the cases. Let me know if that would work so we don't bring back the `fallthrough` and we can use `break` instead.

Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Localization.swift
```diff
@@ -321,7 +321,7 @@ extension String {
         case 4:
             return String.localizedStringWithFormat(format, arguments[0], arguments[1], arguments[2], arguments[3])
         case 0:
```
billypchan: can we remove case 0, since it has has same handling as default

Source/UserSession/CallStateObserver.swift
billypchan: so the fall through here will make this case and below case running same code, it would be clearer to merge 2 cases.

billypchan (commented):
comment about fall through

Source/Notifications/Push notifications/Notification Types/LocalNotificationType+Localization.swift
agisilaos: Already did. You looked at an outdated file.

agisilaos:
Took care of all of your comments @billypchan! Ready for another review.

billypchan (approved):

KaterinaWire (approved):

PR #1611, opened 2021-12-14
fix: Remove reply and like notification actions SQSERVICES-1069

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1069" title="SQSERVICES-1069" target="_blank">TaskSQSERVICES-1069</a>  [iOS] Remove notification actions
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In this PR I remove/ "disable" the reply and like notification actions from the list of available actions. 

### Issues

The "reply" and "like" notification actions (visible when you expand a notification in the Notification Center) do not work: the reply fails to send and will crash the app, the like has no effect.

### Causes (Optional)

The reason is something to do with a past refactoring where we no longer have the message in the database until the main application is opened. There's no simple way to solve this for now, but it will be much easier to fix once we start using the notification service extension, which will be solely responsible for processing push notifications. Until then, we should remove these actions from notifications.

----
##### References


Source/Notifications/Push notifications/Notification Types/PushNotificationCategory.swift
```diff
@@ -47,17 +47,17 @@ enum PushNotificationCategory: String, CaseIterable {
...
-            return [ConversationNotificationAction.reply]
+            return []
         case .conversationWithMute:
-            return [ConversationNotificationAction.reply, ConversationNotificationAction.mute]
+            return [ConversationNotificationAction.mute]
         case .conversationWithLike:
-            return [ConversationNotificationAction.reply, ConversationNotificationAction.like]
+            return []
         case .conversationWithLikeAndMute:
-            return [ConversationNotificationAction.reply, ConversationNotificationAction.like, ConversationNotificationAction.mute]
```
billypchan, line 60: should we remove the `ConversationNotificationAction.like` and `ConversationNotificationAction.reply` enum cases to check they are used in somewhere else?

agisilaos: I think we solved this. 🚢

billypchan (approved):

David-Henner (approved):

zenkins:
Hello, @agisilaos! Merging PR for command **release patch** for PR #**1611**. Will create a **patch** release after merged.
Next version is 404.2.1.

PR #1615, opened 2021-12-17
feat: support federation of 1:1 calls - FS-12

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-12" title="FS-12" target="_blank">TaskFS-12</a>  [iOS] As a user, I can place a call with a user in my contact list who is part of a federated backend
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Support is needed to enable calling between federated backends. The avs library provides and expects a domain as part of conversations and users identifiers. The identifiers are provided as strings composed of the object's identifier and domain. e.g: `E621E1F8-C36C-495A-93FC-0C247A3E6E5F@wire.link`

### Solutions

- Introduced struct `AVSIdentifier` to hold the objects remote identifier & domain (optional)
- Replaced the users and conversations identifiers type (`UUID`) with `AVSIdentifier` in calling code.
- Replaced the users and conversations identifiers type (`UUID`) with `String` in payload structs to support serialised ids when communicating with AVS.
- Converted serialised ids provided in avs callback to `AVSIdentifier`
- Used domain to fetch users and conversations

### Testing

- Used `AVSIdentifier` where needed
- Added tests for `AVSClient` support of `AVSIdentifier`
- Added tests to check that objects are fetched according to remote id and domain

----

----

----
##### References


typfel:
I think you mentioned this previously but the reason why you are not using the  QualifiedID type from data model is that you need to keep the `domain` optional and `QualifiedID` requires it. Once we've fully transitioned to only use qualified IDs there's nothing stopping us from using the `QualifiedID` type, right?

typfel (approved):

David-Henner:
> I think you mentioned this previously but the reason why you are not using the QualifiedID type from data model is that you need to keep the `domain` optional and `QualifiedID` requires it. Once we've fully transitioned to only use qualified IDs there's nothing stopping us from using the `QualifiedID` type, right?

That's right, since the AVS library passes IDs to us as a String which may or may not contain the domain, we need a type that can accommodate both use cases

Source/Calling/AVSIdentifier/AVSIdentifier+Stub.swift
```diff
@@ -0,0 +1,25 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public extension AVSIdentifier {
+    static var stub: AVSIdentifier {
+        AVSIdentifier(identifier: UUID(), domain: "domain.com")
+    }
+}
```
johnxnguyen, line 25: Does this need to be part of sources or can we move it to the test source directory?

johnxnguyen (approved):

Source/Calling/AVSIdentifier/AVSIdentifier+Stub.swift
David-Henner: Originally, I've added it to the test source directory, but the UI project would also need it so I moved it to the sources directory to make it available instead of duplicating it

David-Henner:
release major

zenkins:
Hello, @David-Henner! Merging PR for command **release major** for PR #**1615**. Will create a **major** release after merged.
Next version is 406.0.0.

PR #1629, opened 2022-01-05
feat: support federated client discovery  - FS-140

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-140" title="FS-140" target="_blank">StoryFS-140</a>  [iOS] As the organizer of a conference call I like to place a conference call in a conversation with users who are part of different federated backends
  </td></table>
  <br />
 

----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

In order to place conference calls, we need to pass the list of clients in the conversation to avs.
We typically get this list from the backend. 
For federation, we need to query the list using a new endpoint: `/conversations/{cnv_domain}/{cnv}/proteus/messages`
The response payload will also be different from the legacy endpoint payload.

### Solutions

- Updated `CallingRequestStrategy`
  - to use a request adapted to federation flag
  - to decode the response payload according to federation flag

- Added new Decodable struct `FederatedClientDiscoveryResponsePayload` to decode response from federated endpoint
- Refactored existing `ClientDiscoveryResponsePayload`

### Dependencies (Optional)

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/464

### Testing

- Added tests to cover request creation and payload decoding

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [x] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1641, opened 2022-01-11
fix: call events processing - FS-331

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-331" title="FS-331" target="_blank">BugFS-331</a>  [iOS] Call failures in clients with federation disabled
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

**Given** federation is disabled **when** user A starts a calls with user B (either 1:1 or conference), **then** the call appears as ringing for user A but no incoming call is received on user B side.

### Causes (Optional)

The `ZMUpdateEvent` received by user B contains information about the domain of the conversation and the sender.
When processing the event, we create `AVSIdentifiers` with the object's (conversation / user) identifier and domain.
However, when federation is disabled, the client doesn't have any information about a conversation or a user's domain.
Which results in a mismatch between the identifiers provided to avs when forwarding the processed call events, and the identifiers used by the client. 

### Solutions

ignore the domain when processing call events if federation is disabled

### Dependencies (Optional)

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1629

### Testing

I tried to add unit tests to verify that the domain was ignored, however, I lack the `data` parameter necessary to create a valid `ZMUpdateEvent`.

----

----

----
##### References


johnxnguyen (approved):

johnxnguyen:
@David-Henner do we need to compatibility with old clients? How does the domain get set in the update event? Is it from the sending side when A initiates a call, regardless of whether federation is enabled?

David-Henner:
> How does the domain get set in the update event? Is it from the sending side when A initiates a call, regardless of whether federation is enabled?

I think this is related to the backend. Federation aware backends include qualified identifiers in the payload of update events. This would explain why, regardless of federation being disabled, we receive information about the domain. 

In old clients without **federation support**, we don't have any notion of domain. Even if the backend includes it in the payload, it will be ignored. 

The risk is with clients that **support federation** but have it **disabled**. Since they will decode the domain from update events, we need to ignore it when processing the `ZMUpdateEvent` objects.

We might make this more secure and avoid further issues by ignoring the domain in the [data model layer](https://github.com/wireapp/wire-ios-data-model/blob/4eef5433cd6146a02b130d8b29cde4ee962a956b/Source/Utilis/ZMUpdateEvent%2BWireDataModel.m#L70). However, I don't know if this layer is federation aware. Note that this would happen when builds with federation disabled meet federation aware backends.

johnxnguyen (approved):

sb88k (approved):

David-Henner:
release patch

PR #1644, opened 2022-01-13
feat: Separate access controls for Guests and Services SQSERVICES-1052

agisilaos:
----

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

With this PR I separate the access controls for Services and Guests. Not sure if a new test needs to be added since the one `testThatItSendsRequestToChangeAccessMode` is being covered by `setAllowGuests`.

##### References


Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -167,6 +171,32 @@ extension ZMConversation {
                     userSession.updateEventProcessor?.storeAndProcessUpdateEvents([event], ignoreBuffer: true)
                 }
                 completion(.success)
+            } else {
+                zmLog.debug("Error creating wireless link: \(response)")
+                completion(.failure(SetAllowServicesError.unknown))
```
KaterinaWire, line 176: Should be `completion(.failure(SetAllowGuestsError.unknown))` ?

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -205,12 +235,22 @@ internal struct WirelessRequestFactory {
         return .init(path: "/conversations/\(identifier)/code", method: .methodDELETE, payload: nil)
     }
 
-    static func set(allowGuests: Bool, for conversation: ZMConversation) -> ZMTransportRequest {
+    static func setGuests(allowGuests: Bool, for conversation: ZMConversation) -> ZMTransportRequest {
```
KaterinaWire, line 238: I would rename the method to setAllowGuests(_allowGuests: Bool, Conversation: ZMConversation) since we are setting the allowGuests property, not the guests

agisilaos: We have a couple of methods with the same name in UI project that's why I decided to go with that naming convention.

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -176,6 +180,32 @@ extension ZMConversation {
...
+                let event = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil) {
+                //TODO: Change self.allowGuests to allowServices when you make the appropriate change in Data Model
+                self.allowGuests = allowServices
+                // Process `conversation.access-update` event
+                userSession.syncManagedObjectContext.performGroupedBlock {
+                    userSession.updateEventProcessor?.storeAndProcessUpdateEvents([event], ignoreBuffer: true)
+                }
+                completion(.success)
+            } else {
+                zmLog.debug("Error creating wireless link: \(response)")
```
KaterinaWire, line 201: I think this error message is incorrect because we don't create any links to the services.

Source/Data Model/Conversation+AccessMode.swift
KaterinaWire: I think `setGuests` or `setServices` sounds like we want to add specific guests or services to the conversation. In our case, we want to get a request that sets the access role. Also, according to the tech spec, the new `access_role_v2` will contain a set of values. What if we create a setAccessRoleRequest(...) method with a set of roles as a parameter?

Tests/Source/Integration/ConversationTests+Guests.swift
```diff
@@ -87,10 +94,14 @@ class ConversationTests_Guests: IntegrationTest {
         guard let request = mockTransportSession.receivedRequests().first else { return }
         XCTAssertEqual(request.path, "/conversations/\(conversation.remoteIdentifier!.transportString())/code")
         XCTAssertEqual(request.method, .methodPOST)
+#endif
     }
 
-    func testThatItSendsRequestToSetModeIfLegacyWhenFetchingTheLink() {
+    func testThatItSendsRequestToSetModeIfLegacyWhenFetchingTheLink() throws {
         // given
+#if arch(x86_64)
```
KaterinaWire, line 102: Why should we skip only ConversationTests_Guests tests?

Tests/Source/Integration/ConversationTests+Guests.swift
```diff
@@ -54,10 +57,14 @@ class ConversationTests_Guests: IntegrationTest {
         XCTAssertEqual(mockTransportSession.receivedRequests().count, 1)
         guard let request = mockTransportSession.receivedRequests().first else { return }
         XCTAssertEqual(request.path, "/conversations/\(conversation.remoteIdentifier!.transportString())/access")
+        #endif
```
KaterinaWire, line 60: There is a whitespace issue

Tests/Source/Integration/ConversationTests+Guests.swift
agisilaos: Because that's where we get the failures/crashes when testing with X86 simulator on M1 Mac.

agisilaos: Also the `#if statement` in which we use to skip tests is also under evaluation and it may change in the future.

Cartfile.resolved
```diff
@@ -7,12 +7,12 @@ github "wireapp/libPhoneNumber-iOS" "0.9.3_xcframework"
 github "wireapp/ocmock" "v3.4.3_xcframework"
 github "wireapp/swift-protobuf" "1.15.0_xcframework"
 github "wireapp/wire-ios-cryptobox" "31.0.0"
-github "wireapp/wire-ios-data-model" "282.0.0"
+github "wireapp/wire-ios-data-model" "feature/seperate-access-controls-for-guest-links-and-services"
```
billypchan, line 10: the frameworks need release first

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -176,6 +180,32 @@ extension ZMConversation {
...
+    public func setAllowServices(_ allowServices: Bool, in userSession: ZMUserSession, _ completion: @escaping (VoidResult) -> Void) {
+        guard canManageAccess else {
+            return completion(.failure(WirelessLinkError.invalidOperation))
+        }
+
+        let request = WirelessRequestFactory.setServices(allowServices: allowServices, for: self)
+        request.add(ZMCompletionHandler(on: managedObjectContext!) { response in
+            if let payload = response.payload,
+                let event = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil) {
+                //TODO: Change self.allowGuests to allowServices when you make the appropriate change in Data Model
```
billypchan, line 193: todo leave

Tests/Source/Integration/ConversationTests+Guests.swift
```diff
@@ -28,8 +28,11 @@ class ConversationTests_Guests: IntegrationTest {
         createTeamAndConversations()
     }
 
-    func testThatItSendsRequestToChangeAccessMode() {
+    func testThatItSendsRequestToChangeAccessMode() throws {
         // given
+#if arch(x86_64)
```
billypchan, line 33: this check is always pass since this framework is alway build for `x86_64`. I have not find a way to check the test is running on M1 or X86 machine.

Tests/Source/Integration/ConversationTests+Guests.swift
```diff
@@ -54,10 +57,14 @@ class ConversationTests_Guests: IntegrationTest {
         XCTAssertEqual(mockTransportSession.receivedRequests().count, 1)
         guard let request = mockTransportSession.receivedRequests().first else { return }
         XCTAssertEqual(request.path, "/conversations/\(conversation.remoteIdentifier!.transportString())/access")
+        #endif
     }
 
-    func testThatItSendsRequestToCreateTheLink() {
+    func testThatItSendsRequestToCreateTheLink() throws {
         // given
+#if arch(x86_64)
```
billypchan, line 65: also here, please undo this first.

billypchan (changes requested):
undo M1 Mac check

Source/Data Model/Conversation+AccessMode.swift
agisilaos: After thinking it more during the weekend I came to the conclusion your initial suggestion is a good one. Will go with that one for now.

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -176,6 +181,32 @@ extension ZMConversation {
...
+                self.allowGuests = allowServices
+                // Process `conversation.access-update` event
+                userSession.syncManagedObjectContext.performGroupedBlock {
+                    userSession.updateEventProcessor?.storeAndProcessUpdateEvents([event], ignoreBuffer: true)
+                }
+                completion(.success)
+            } else {
+                zmLog.debug("Error \(response)")
```
KaterinaWire, line 202:
```suggestion
                zmLog.debug("Error setting access role:  \(response)")
```

KaterinaWire:
@agisilaos Looks good, but there are some TODO, SwiftLint issues, you need to fix and add tests. So I would wait until everything is ready for the second review

Tests/Source/Integration/ConversationTests+Guests.swift
```diff
@@ -126,6 +126,7 @@ class ConversationTests_Guests: IntegrationTest {
 
     func testThatItSendsRequestToFetchTheLink_NoLink() {
         // given
+        XCTExpectFailure("Test is expected to fail on an m1 machine because of the login method for integration tests")
```
billypchan, line 129:
```suggestion
#if targetEnvironment(simulator) && swift(>=5.4)
        XCTExpectFailure("Test is expected to fail on an m1 machine because of the login method for integration tests")
#endif        
```

billypchan (changes requested):
fix build error for Xcode 12

Tests/Source/Integration/ConversationTests+Guests.swift
agisilaos: I'm removing the XCTExpectFailure for now.

Tests/Source/Synchronization/ZMConversationAccessModeTests.swift
```diff
@@ -51,16 +51,16 @@ public class ZMConversationAccessModeTests: MessagingTest {
...
         XCTAssertEqual(request.method, .methodPUT)
         XCTAssertEqual(request.path, "/conversations/\(conversation.remoteIdentifier!.transportString())/access")
         let payload = request.payload as! [String: AnyHashable]
         XCTAssertNotNil(payload)
         XCTAssertNotNil(payload["access"])
         XCTAssertEqual(Set(payload["access"] as! [String]), Set(["invite", "code"]))
-        XCTAssertNotNil(payload["access_role"])
-        XCTAssertEqual(payload["access_role"], "non_activated")
+        XCTAssertNotNil(payload["access_role_v2"])
+        XCTAssertEqual(payload["access_role_v2"], "non_activated")
```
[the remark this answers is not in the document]
agisilaos: > We should add tests to check the new values for this new property?

@KaterinaWire Can you expand a bit on this? How that would look like?

I want to tackle data model changes first and then come back here and make any appropriate changes.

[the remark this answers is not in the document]
KaterinaWire: "access_role_v2" can have a different set of values and it would be good to verify it.

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -157,19 +162,36 @@ extension ZMConversation {
             return completion(.failure(WirelessLinkError.invalidOperation))
         }
 
-        let request = WirelessRequestFactory.set(allowGuests: allowGuests, for: self)
+        setAllowGuestsAndServices(allowGuests: allowGuests, allowServices: self.allowServices, in: userSession, completion)
+    }
+
+    /// Changes the conversation access mode to allow services.
```
agisilaos, line 168: Do we need a better doc for this method here? Explaining better what it does?

Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -157,19 +162,36 @@ extension ZMConversation {
...
+    public func setAllowServices(_ allowServices: Bool, in userSession: ZMUserSession, _ completion: @escaping (VoidResult) -> Void) {
+        guard canManageAccess else {
+            return completion(.failure(SetAllowServicesError.invalidOperation))
+        }
+
+        setAllowGuestsAndServices(allowGuests: self.allowGuests, allowServices: allowServices, in: userSession, completion)
+
+    }
+
+    /// Changes the conversation access mode to allow services.
```
agisilaos, line 178: same here

johnxnguyen (approved):

PR #1668, opened 2022-02-09
fix: muted by incoming call while in call SQCALL-475

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQCALL-475" title="SQCALL-475" target="_blank">BugSQCALL-475</a>  iOS: User gets muted in a group call automatically on receiving a new call and hanging up the new call 
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

when receiving an incoming call while already in a call, the user is automatically muted in the ongoing call.

### Causes (Optional)

we don't check that there is no active call when muting 

### Solutions

check that there is no active call before muting

----

----

----
##### References


Tests/Source/Calling/WireCallCenterV3Tests.swift
```diff
@@ -849,15 +849,47 @@ class WireCallCenterV3Tests: MessagingTest {
...
+        let incomingCall = CallSnapshotTestFixture.callSnapshot(
+            conversationId: incomingCallConversationId,
+            callCenter: sut,
+            clients: [],
+            state: incomingState
+        )
+
+        sut.callSnapshots = [activeCallConversationId: activeCall,
+                           incomingCallConversationId: incomingCall]
+        sut.muted = false
```
KaterinaWire, line 885: If this test checks that it does not mute the microphone, so the initial state must be true. Am I wrong?

David-Henner: I don't think so, we want to verify that `muted` stays `false`

sb88k: Then, maybe we can have different test case name like
`testThatTheMicrophoneStaysMuted_WhenHandlingIncomingGroupCall_WhileAlreadyInCall`

sb88k: Ah, you are right.  Now I fully got what this test is trying to check. 👍

sb88k (approved):

Tests/Source/Calling/WireCallCenterV3Tests.swift
KaterinaWire: @David-Henner Thanks for the explanation! Now it's clear to me, you're right.

KaterinaWire (approved):

David-Henner:
release patch

zenkins:
Hello, @David-Henner! Merging PR for command **release patch** for PR #**1668**. Will create a **patch** release after merged.
Next version is 412.3.6.

PR #1676, opened 2022-02-25
feat: Set up Sync Engine for 2 Factor Authentication SQSERVICES-1164

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1164" title="SQSERVICES-1164" target="_blank">TaskSQSERVICES-1164</a>  [iOS] 2nd auth factor for "login"
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This is a PR where we set up Sync Engine to support 2FA. This PR:

- Adds 2FA error code response handler in `ZMLoginTranscoder`.
- Adds an additional optional parameter to the body: `verification_code` for the login endpoint
- Adds new error to `ZMUserSessionErrorCode`: `ZMUserSessionAccountIsPendingVerification`
- Adds new method in `ZMAuthenticationStatus`: `didFailLoginBecause2FAIsRequired()`

In addition to that as you can see a lot of changes have been made in Tests. Objc doesn't support default argument values. You need to provide a value, or nil. ~~That's why we've updated methods to have twoVerificationCode as a parameter and its value is nil.~~

I followed @johnxnguyen suggestion: To provide two variants of this method, once that accepts a verfication code, and one that uses a default.

So now we don't have to pass a verificationCode in every test.

----
##### References


Source/UserSession/ZMCredentials.m
```diff
@@ -48,11 +49,12 @@ + (nonnull ZMPhoneCredentials *)credentialsWithPhoneNumber:(nonnull NSString *)p
 
 @implementation ZMEmailCredentials
 
-+ (nonnull ZMEmailCredentials *)credentialsWithEmail:(nonnull NSString *)email password:(nonnull NSString *)password
++ (nonnull ZMEmailCredentials *)credentialsWithEmail:(nonnull NSString *)email password:(nonnull NSString *)password twoFactorVerificationCode:(nullable NSString *)code
```
johnxnguyen, line 52: If you want to avoid having to change all the tests, you can simply provide two variants of this method, once that accepts a verfication code, and one that uses a default. For instance:

```objc
(nonnull ZMEmailCredentials *)credentialsWithEmail:(nonnull NSString *)email password:(nonnull NSString *)password {
    [ZMEMailCredentials credentialsWithEmail:email password:password twoFactorVerificationCode:nil];
}


(nonnull ZMEmailCredentials *)credentialsWithEmail:(nonnull NSString *)email password:(nonnull NSString *)password twoFactorVerificationCode:(nullable NSString *)code {
  ZMEmailCredentials *credentials = [[ZMEmailCredentials alloc] init];
     credentials.email = email;
     credentials.password = password;
     credentials.twoFactorVerificationCode = code;
     return credentials;
}
```

Source/Public/ZMCredentials.h
```diff
@@ -26,6 +26,7 @@
 @property (nonatomic, copy, readonly, nullable) NSString *password;
 @property (nonatomic, copy, readonly, nullable) NSString *phoneNumber;
 @property (nonatomic, copy, readonly, nullable) NSString *phoneNumberVerificationCode;
+@property (nonatomic, copy, readonly, nullable) NSString *twoFactorVerificationCode;
```
johnxnguyen, line 29: For consistency, I suggest to call this `emailVerificationCode`

Source/Synchronization/Transcoders/ZMLoginTranscoder.m
```diff
@@ -175,6 +175,9 @@ - (void)didReceiveResponse:(ZMTransportResponse *)response forSingleRequest:(ZMS
         
         if ([self isResponseForSuspendedAccount:response]) {
             [authenticationStatus didFailLoginBecauseAccountSuspended];
+        } else if ([self isResponseFor2FACode:response]){
+            [authenticationStatus didFailLoginBecause2FAIsRequired];
```
johnxnguyen, line 179: We should have a test that covers this new code path.

Source/UnauthenticatedSession/ZMAuthenticationStatus.h
```diff
@@ -102,6 +102,7 @@ typedef NS_ENUM(NSUInteger, ZMAuthenticationPhase) {
 - (void)didFailLoginWithEmailBecausePendingValidation;
 - (void)didFailLoginWithEmail:(BOOL)invalidCredentials;
 - (void)didFailLoginBecauseAccountSuspended;
+- (void)didFailLoginBecause2FAIsRequired;
```
johnxnguyen, line 105: The naming of the verification code is a bit inconsistent. Here it's referred to as 2FA, else where as two factor verification. Let's stick with one and use it consistently. My suggestion would be would be `didFailLoginWithEmailBecauseVerificationCodeIsRequired`

Source/UnauthenticatedSession/ZMAuthenticationStatus.m
```diff
@@ -374,8 +374,7 @@ - (ZMEmailCredentials *)emailCredentials
 {
     if (self.loginCredentials.credentialWithEmail) {
         return [ZMEmailCredentials credentialsWithEmail:self.loginCredentials.email
-                                               password:self.loginCredentials.password
-                                               twoFactorVerificationCode:self.loginCredentials.emailVerificationCode];
```
johnxnguyen: Why do we not include the verification code?

agisilaos: Because it is optional, I thought because it is optional we shouldn't have it here. @johnxnguyen What's your thoughts on this?

johnxnguyen: But here we want to return email credentials from some other credentials, so it would make sense to me to pass along the verification code if it's there. Or am I missing something?

agisilaos: You don't miss something, that is correct and that's why I brought it back with my last commit.

Tests/Source/Synchronization/Transcoders/ZMLoginTranscoderTests.m
```diff
@@ -395,6 +395,23 @@ - (void)testThatItCallsAuthenticationFailOnEmailLoginWithSuspendedAccount
     }];
 }
 
+- (void) testThatItCallsAccountPendingVerification
```
KaterinaWire, line 398: I didn't find `THEN` block in this test

Source/Public/ZMCredentials.h
```diff
@@ -44,4 +45,6 @@
 
 + (nonnull ZMEmailCredentials *)credentialsWithEmail:(nonnull NSString *)email password:(nonnull NSString *)password;
 
++ (nonnull ZMEmailCredentials *)credentialsWithEmail:(nonnull NSString *)email password:(nonnull NSString *)password twoFactorVerificationCode:(nullable NSString *)code;
```
KaterinaWire, line 48:
```suggestion
+ (nonnull ZMEmailCredentials *)credentialsWithEmail:(nonnull NSString *)email password:(nonnull NSString *)password emailVerificationCode:(nullable NSString *)code;
```

Tests/Source/Synchronization/Transcoders/ZMLoginTranscoderTests.m
agisilaos: I'm following the same pattern as the test above. You can check `testThatItCallsAuthenticationFailOnEmailLoginWithSuspendedAccount`

johnxnguyen:
You have commits from David here because you merged in develop. But probably your base branch is something else?

KaterinaWire (approved):

PR #1683, opened 2022-03-07
feat: handle call events from notification extension SQSERVICES-1111

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1111" title="SQSERVICES-1111" target="_blank">TaskSQSERVICES-1111</a>  [iOS] Notification extension: Handle call notifications
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR adds support for handling call events received in the notification service extension. Specifically, this means that when notification service extension (a separate process) receives a calling event (either call start or end), it will then attempt to pass the call message to the main app so that it can be processed. It does this by sending a "fake voip" push to the main app which is received in the PushKit method 

```swift
pushRegistry(_ registry: PKPushRegistry, didReceiveIncomingPushWith payload: PKPushPayload, for type: PKPushType, completion: @escaping () -> Void)
```

The payload is set from the notification service extension, and can contain anything that we want. The data we put in the payload is everything that that is needed in the sync engine to process the call. This is encapsulated in `VoIPPushPayload`. Processing the call event means to pass it to AVS, which will typically result in reporting the call to CallKit.

**Reporting the call to CallKit is a hard requirement by iOS, since voip pushes must only be used for calling purposes.** We must in fact report the call to CallKit before the end of the push kit delegate method mentioned previously, otherwise iOS will terminate the app.

This presents some challenges:
1. Typically we only reported calls to CallKit after AVS has processed the call event. But this happens asynchronously and we need to do it synchronously. For this reason, we preemptively report the call to CallKit, then pass the message to AVS.2. 
2. The path from push payload to reporting the call to CallKit is long and has **many** points of early return, all of which mean that we would not report the call to CallKit and the app would be terminated. I've made these points of failure very transparent so that we can understand why it failed, if it does.
3. Allowing the app to be terminated several times (3 or 4) results in our app being punished: iOS will stop to deliver voip pushes altogether. For this reason, we must take extra care to avoid hitting a point of failure.
4. The approach I've taken to ensure this is make sure that we don't even send the call event to the main app from the notification service extension if we think that it might fail. Instead, we'll just show a user notification.

## Dependencies

- https://github.com/wireapp/wire-ios-request-strategy/pull/514

----

----

----
##### References


Source/SessionManager/SessionManager+Push.swift
```diff
@@ -22,6 +22,31 @@ import UserNotifications
...
+public struct PushFromNotificationExtension: Codable {
+    enum CodingKeys: String, CodingKey {
+        case conversationId = "conversationId"
+        case senderId = "senderId"
+        case senderClientID = "senderClientID"
+        case conversationDomain = "conversationDomain"
+        case senderDomain = "senderDomain"
+        case payloadData = "payloadData"
+        case timestamp = "timestamp"
+    }
```
johnxnguyen, line 39: Is there a particular reason to explicitly define the raw values? Since they don't differ from the case names, the compiler can generate them for you.

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -75,6 +100,41 @@ extension SessionManager: PKPushRegistryDelegate {
...
+                                                                               senderUUID: pushPayload.senderId,
+                                                                               clientId: pushPayload.senderClientID,
+                                                                               conversationDomain: pushPayload.conversationDomain,
+                                                                               senderDomain: pushPayload.senderDomain,
+                                                                               payload: pushPayload.payloadData,
+                                                                               eventTimestamp: pushPayload.timestamp)
+        })
+    }
+
+    private func handleIncomingNonCallPushNotification(_ payload: PKPushPayload, for type: PKPushType, completion: @escaping () -> Void) {
```
johnxnguyen, line 137: I think it would be more correct to say `handleLegacyPushPayload`.

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -75,6 +100,41 @@ extension SessionManager: PKPushRegistryDelegate {
...
+              let account = self.accountManager.account(with: accountId) else {
+                  return completion()
+              }
+
+        withSession(for: account, perform: { userSession in
+            guard let dictionaryPayload = payload.dictionaryPayload as? [String : Any],
+                let pushPayload = PushFromNotificationExtension(dictionaryPayload) else {
+                Logging.push.safePublic("Aborted processing of payload: \(payload)")
+                return
+            }
```
johnxnguyen, line 124: Maybe we can get this payload before loading the session.

Source/SessionManager/SessionManager+Push.swift
KaterinaWire: No, there is no reason.
Fixed.

Source/SessionManager/SessionManager+Push.swift
KaterinaWire: Changed

Source/SessionManager/SessionManager+Push.swift
KaterinaWire: Yes, it makes sense. 
Fixed

johnxnguyen (commented):
Looks good to me. One question though that I didn't yet consider... typically the calling request strategy will process events while in the background, which is how it usually processes call events. In such cases, is the call event not persisted in the event db?

Do we also need to take care now that we detect call events from the notification extension? As in, are we still persisting it, and if so, should we?

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -22,6 +22,50 @@ import UserNotifications
...
+public struct PushFromNotificationExtension: Codable {
+    enum CodingKeys: String, CodingKey {
+        case conversationId
+        case senderId
+        case senderClientID
+        case conversationDomain
+        case senderDomain
+        case payloadData
+        case timestamp
+    }
```
johnxnguyen, line 39: I don't think the coding keys are necessary, they can be inferred by the compiler.

johnxnguyen: You also only need to conform to `Decodable`.

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -22,6 +22,50 @@ import UserNotifications
...
+
+    let conversationId: UUID
+    let senderId: UUID
+    let senderClientID: String
+    let conversationDomain: String?
+    let senderDomain: String?
+    let payloadData: Data
+    let timestamp: Date
+
+    public init? (event: ZMUpdateEvent) {
```
johnxnguyen, line 49: Where is this being called?

KaterinaWire: From the UI to convert the event to a payload for the CallKit.

johnxnguyen: I see now, added some comments in the other PR.

Source/SessionManager/SessionManager+Push.swift
KaterinaWire: Fixed

David-Henner (approved):

Source/Calling/CallKitManager.swift
```diff
@@ -73,7 +73,14 @@ public class CallKitManager: NSObject {
     fileprivate var callStateObserverToken: Any?
     fileprivate var missedCallObserverToken: Any?
     fileprivate var connectedCallConversation: ZMConversation?
-    fileprivate var calls: [UUID: CallKitCall]
+
+    fileprivate var calls: [UUID: CallKitCall] {
+        didSet {
+            VoIPPushHelper.setOngoingCalls(
+                conversationIDs: calls.values.map { $0.conversation.remoteIdentifier! }
+            )
```
johnxnguyen, line 81: This info gets stored in user defaults that is accessible from the notification service extension. We use it to determine if we can report the call to CallKit.

Source/Calling/CallKitManager.swift
```diff
@@ -305,30 +331,40 @@ extension CallKitManager {
...
         log("provider.reportNewIncomingCall")
 
-        provider.reportNewIncomingCall(with: callUUID, update: update) { [weak self] (error) in
+        provider.reportNewIncomingCall(with: callUUID, update: update) { [weak self] error in
             if let error = error {
                 self?.log("Cannot report incoming call: \(error)")
                 self?.calls.removeValue(forKey: callUUID)
                 conversation.voiceChannel?.leave()
-            } else {
-                self?.mediaManager?.setupAudioDevice()
```
johnxnguyen, line 314: We now set up the audio device when the user answers the call.

Source/Calling/CallKitManager.swift
```diff
@@ -472,13 +510,13 @@ extension CallKitManager: WireCallCenterCallStateObserver, WireCallCenterMissedC
         case .incoming(video: let video, shouldRing: let shouldRing, degraded: _):
             if shouldRing, let caller = caller as? ZMUser {
                 if conversation.mutedMessageTypesIncludingAvailability == .none {
-                    reportIncomingCall(from: caller, in: conversation, video: video)
+                    try? reportIncomingCall(from: caller, in: conversation, video: video)
```
johnxnguyen, line 513: Because AVS will report the incoming call to us, it will change the call state. Here we listen for the incoming call state and try to report the incoming call. This happens in both the normal flow and the flow with the NSE involved. The difference is that in flow with the NSE, we've already reported the call incoming.

Source/Calling/WireCallCenterV3.swift
```diff
@@ -90,6 +90,8 @@ public class WireCallCenterV3: NSObject {
     /// Set to true once AVS calls the ReadyHandler. Setting it to `true` forwards all previously buffered events to AVS.
     var isReady: Bool = false {
         didSet {
+            VoIPPushHelper.isAVSReady = isReady
```
johnxnguyen, line 93: AVS won't process events if it's not ready yet. The NSE can use this info to know whether to pass the call event to the main app or just show a user notification instead.

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -83,6 +84,83 @@ extension SessionManager: PKPushRegistryDelegate {
     }
 
     public func pushRegistry(_ registry: PKPushRegistry, didReceiveIncomingPushWith payload: PKPushPayload, for type: PKPushType, completion: @escaping () -> Void) {
+        if let voipPushPayload = VoIPPushPayload(payload: payload) {
+            do {
+                try handleCallPushPayload(voipPushPayload, completion: completion)
+            } catch let error as VOIPPushError {
+                Logging.push.safePublic("Failed to handle voip push payload: \(error)")
+            } catch {
+                Logging.push.safePublic("Failed to handle voip push payload for unknown reason")
```
johnxnguyen, line 93: If we had a failure then we'll immediately get a crash when this method returns. We can't do anything to prevent that, the least we can do is report why it failed and try to fix that the next time.

Source/SessionManager/SessionManager.swift
```diff
@@ -205,7 +206,14 @@ public final class SessionManager: NSObject, SessionManagerType {
         }
     }
 
-    public fileprivate(set) var backgroundUserSessions: [UUID: ZMUserSession] = [:]
+    public private(set) var backgroundUserSessions = [UUID: ZMUserSession]() {
+        didSet {
+            VoIPPushHelper.setLoadedUserSessions(
+                accountIDs: Array(backgroundUserSessions.keys)
+            )
```
johnxnguyen, line 213: We can't process call events without a user session, so the NSE won't pass the event if the user session isn't loaded yet.

Source/SessionManager/SessionManager.swift
```diff
@@ -258,7 +266,11 @@ public final class SessionManager: NSObject, SessionManagerType {
     fileprivate var memoryWarningObserver: NSObjectProtocol?
     fileprivate var isSelectingAccount: Bool = false
 
-    public var callKitManager: CallKitManager?
+    public var callKitManager: CallKitManager? {
+        didSet {
+            VoIPPushHelper.isCallKitAvailable = callKitManager != nil
```
johnxnguyen, line 271: You can't report a call to CallKit if there is no CallKitManager, so the NSE needs to know if one is available.

johnxnguyen (approved):

Tests/Source/Synchronization/Strategies/CallingRequestStrategyTests.swift
```diff
@@ -418,7 +418,9 @@ class CallingRequestStrategyTests: MessagingTest {
 
     func testThatItAsksCallCenterToMute_WhenReceivingRemoteMuteEvent() {
         // GIVEN
-        let json = ["type": "REMOTEMUTE"]
+        let json = ["src_userid": UUID.create().uuidString,
+                    "resp": false,
+                    "type": "REMOTEMUTE"] as [String : Any]
```
David-Henner, line 423: swift lint warning here :)

KaterinaWire: Fixed

KaterinaWire:
release major

zenkins:
Hello, @KaterinaWire! Merging PR for command **release major** for PR #**1683**. Will create a **major** release after merged.
Next version is 423.0.0.

PR #1684, opened 2022-03-09
feat: (2FA) Prepare for requesting email verification code for login SQSERVICES-1377

agisilaos:
----

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

In this PR we introduce a couple of new things and preparing stuff so 2FA can work during Login:

1. Introducing `requestEmailVerificationCodeForLogin` as a public method so we can use it in the UI Project
2. Added a new` ZMAuthenticationPhase` to the list -> `ZMAuthenticationPhaseRequestEmailVerificationCodeForLogin`
3. `prepareForRequestingEmailVerificationCodeForLogin` has been added to ZMAuthenticationStatus

Open Question: Not exactly sure what we can test atm since we didn't make any changes to the the ZMLoginCodeRequestTranscoder.
----
##### References

### DIFF
--- a/Source/UnauthenticatedSession/UnauthenticatedSession+Login.swift
+++ b/Source/UnauthenticatedSession/UnauthenticatedSession+Login.swift
@@ -69,4 +69,22 @@ extension UnauthenticatedSession {
         }
         return true
     }
+
+    /// Requires an email verification code for login. Returns NO if the email address was invalid
+    @objc(requestEmailVerificationCodeForLogin:)
+    @discardableResult public func requestEmailVerificationCodeForLogin(email: String) -> Bool {
+        do {
+            var email: String? = email
+            _ = try ZMUser.validate(emailAddress: &email)
+        } catch {
+            return false
+        }
+
+        authenticationErrorIfNotReachable {
+            self.authenticationStatus.prepareForRequestingEmailVerificationCode(forLogin: email)
+            RequestAvailableNotification.notifyNewRequestsAvailable(nil)
+        }
+        return true
+
+    }
 }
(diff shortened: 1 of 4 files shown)


Tests/Source/UserSession/ZMAuthenticationStatusTests.swift
```diff
@@ -18,7 +18,7 @@
 
 import XCTest
 
-class ZMAuthenticationStatusTests_PhoneVerification: XCTestCase {
+class ZMAuthenticationStatusTests_PhoneAndEmailVerification: XCTestCase {
```
agisilaos, line 21: Renamed `ZMAuthenticationStatusTests_PhoneVerification` to `ZMAuthenticationStatusTests_PhoneAndEmailVerification` because I don't think it's worth it to create a separate class conforming to XCTestCase for just one test. If you @KaterinaWire or @johnxnguyen think that it would be better to create a separate class for just that test that I wrote here please let me know.

KaterinaWire (approved):

Tests/Source/UserSession/ZMAuthenticationStatusTests.swift
KaterinaWire: I think we can have both tests in the same class. 
The test name might be `ZMAuthenticationStatusTests_SecondFactorVerification`, but I'm also fine with the current name.

Source/UnauthenticatedSession/UnauthenticatedSession+Login.swift
```diff
@@ -69,4 +69,22 @@ extension UnauthenticatedSession {
         }
         return true
     }
+
+    /// Requires an email verification code for login. Returns NO if the email address was invalid
```
johnxnguyen, line 73:
```suggestion
    /// Triggers a request for an email verification code for login. 
    ///
    /// Returns: false if the email address was invalid.
```

johnxnguyen (approved):

PR #1700, opened 2022-03-21
fix: Unable to create a conversation with a service SQSERVICES-1394

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1394" title="SQSERVICES-1394" target="_blank">BugSQSERVICES-1394</a>  [iOS] Unable to start a 1:1 conversation with a service
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

if we want to create a conversation with bot. We need to set the `conversation.allowServices` to true.  So `service` would be added to the accessRoles and the payload is going to be correct. The reason wasn't working was because we didn't add "service" to the accessRoles so the payload was incorrect.

The error we got from backend was. `403 "invalid-conversation"` That response appears when adding or removing services. Also that meant that maybe the set of accessRoles was incorrect when trying to create a conversation with a service. 

So by setting `conversation.allowServices = true` we're fixing the issue. We know that accessRoles and the payload is going to be updated correctly.

----
##### References


KaterinaWire (approved):

Source/Data Model/ServiceUser.swift
```diff
@@ -209,6 +209,10 @@ public extension ServiceUser {
         conversation.creator = selfUser
         conversation.team = selfUser.team
 
+        if conversation.allowServices == false {
+            conversation.allowServices = true
+        }
```
johnxnguyen, line 214: I don't think the `if` is necessary here, you can just set it to `true`.

johnxnguyen (commented):
I thought that when creating a conversation, the default value for the access roles already includes services. Is that not true?

Source/Data Model/ServiceUser.swift
agisilaos: Done!

agisilaos:
> I thought that when creating a conversation, the default value for the access roles already includes services. Is that not true?

When creating a (group) conversation allow Guests and Services toggle is turned on by default. But user Can turn them off before creating that conversation. We didn't take into account creating a convo with a service though and the default value for accessRoles wasn't including service.

johnxnguyen (approved):

agisilaos:
release patch

zenkins:
Hello, @agisilaos! Merging PR for command **release patch** for PR #**1700**. Will create a **patch** release after merged.
Next version is 416.0.1.

PR #1710, opened 2022-04-06
fix: crash when searching federated users FS-520

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-520" title="FS-520" target="_blank">BugFS-520</a>  [iOS] App crash when searching for another user including backend domain, adding a space, and adding another character
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The app would crash when searching for a federated user with spaces in the query, such as `john@exa mple.com`.

### Causes

The cause of the crash stems from the usage of the domain (and the handle) as part of the endpoint path. The endpoint to search for federated users is `"/users/by-handle/\(domain)/\(handle)"`, and if either the domain of the handle have illegal characters in them, then we'd fail to create a `URL` instance of the path. Somewhere in the network code we assume we have a valid path and crash when it fails.

### Solutions

The endpoint `"/users/by-handle/\(domain)/\(handle)"` is actually deprecated, and we should instead be using `/search/contacts` which takes the domain as a query parameter. We actually already use this endpoint for normal (non-federated) searches, so I decided to adjust it to accept a possible domain if one was found in the search query.

The search logic was quite complex, so I aimed to make it a bit more manageable by doing some refactoring. The search query string is the user input string, which may contain be a full text search query (eg. "bob"), an exact handle query (e.g. "@bob"), and additionally may contain a search domain (eg. "@bob@example.com").

Previously, the original user input query was being passed around and parsed when both creating a search request and processing the search result. I found this error prone since a mismatch in how the query was interpreted led to unexpected results. Therefore I introduce `Query` that requires a single parse of the user search query, as well as making a clear distinction between exact handle search and full text search. 

Finally, all code related to the deprecated federation search has been removed.

### Testing

Updated existing tests.

----

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----

----
##### References


agisilaos (approved):

David-Henner (approved):

johnxnguyen:
release major

zenkins:
Hello, @johnxnguyen! Merging PR for command **release major** for PR #**1710**. Will create a **major** release after merged.
Next version is 419.0.0.

PR #1726, opened 2022-04-26
feat: load session using async SQSERVICES-1111

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1111" title="SQSERVICES-1111" target="_blank">TaskSQSERVICES-1111</a>  [iOS] Notification extension: Handle call notifications
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In order to successfully process call events fetched from the notification service extension, we must report incoming calls to CallKit by the end of the PushKit method [pushRegistry(_:didReceiveIncomingPushWith:for:completion:)](https://developer.apple.com/documentation/pushkit/pkpushregistrydelegate/2875784-pushregistry), otherwise iOS will terminate the app. See [Apple docs](https://developer.apple.com/documentation/callkit/sending_end-to-end_encrypted_voip_calls) for more detail.

This is not currently possible because our normal flow to process call events from push payloads is inherently asynchronous:. For instance, we must load the user session for a given account, which we will only receive via a completion handler.

Once we have the user session however, we can then gain access to the `WireCallCenter` and proceed to report an incoming call to CallKit.

This PR will help facilitate "synchronous" loading of the user session by leveraging the async / await language features of Swift. We do this by creating an "async" version of `withSession` which uses a continuation to wait for the completion handlers of the existing `withSession` method to be called.

For this to work, the completion handler of `withSession` must be called in every code path out of the method, which I found was not the case when there was a database error. For this reason I converted the argument of the completion handler to contain a result of either the user session or an error.

I updated the code for this change, but didn't focus on handling the new error.

#### How to Test

I added a test to ensure that loading the session works.

----

----

----
##### References

### DIFF
--- a/Source/SessionManager/SessionManager+Push.swift
+++ b/Source/SessionManager/SessionManager+Push.swift
@@ -102,7 +102,9 @@ extension SessionManager: PKPushRegistryDelegate {
                 return completion()
         }
 
-        withSession(for: account, perform: { userSession in
+        withSession(for: account, perform: { result in
+            guard case let .success(userSession) = result else { return }
+            
             Logging.push.safePublic("Forwarding push payload to user session with account \(account.userIdentifier)")
 
             userSession.receivedPushNotification(with: payload.dictionaryPayload, completion: { [weak self] in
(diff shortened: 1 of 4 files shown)


David-Henner (commented):
There are some SwiftLint warnings to take care of

Source/SessionManager/SessionManager+Push.swift
```diff
@@ -102,7 +102,9 @@ extension SessionManager: PKPushRegistryDelegate {
                 return completion()
         }
 
-        withSession(for: account, perform: { userSession in
+        withSession(for: account, perform: { result in
+            guard case let .success(userSession) = result else { return }
+
```
KaterinaWire, line 107: Swift lint issue

johnxnguyen:
This is no longer needed.

PR #1763, opened 2022-07-05
fix: can't login to deep-linked backend FS-799

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-799" title="FS-799" target="_blank">BugFS-799</a>  Cannot login with deeplinked backend
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When switching a backend via deep link, then logging in with valid credentials an alert is shown stating that is issues with the network.

### Causes 

When logging in, there is a reachability check to see if the server is reachable. This was failing because the reachability object was still configured to the previous backend environment, which in this case was unreachable.

### Solutions

When the backend environment changes, then the reachability helper must be recreated and configured with the new server names. 

### Testing

Tested manually

### Notes 

This is a somewhat simple solution to unblock a delivery. I think though that we need to prevent this problem occurring by coupling the reachability logic closely to the backend environment, so that the environment owns the reachability object and we don't need to keep two separate objects in sync. I'll address this in another PR.

----

----

----
##### References

### DIFF
--- a/Source/SessionManager/SessionFactories.swift
+++ b/Source/SessionManager/SessionFactories.swift
@@ -93,25 +93,34 @@ open class UnauthenticatedSessionFactory {
         }
     }
 
-    var reachability: ReachabilityProvider
+    var reachability: ReachabilityProvider & TearDownCapable
     let appVersion: String
 
-    init(appVersion: String,
-         environment: BackendEnvironmentProvider,
-         reachability: ReachabilityProvider) {
+    init(
+      appVersion: String,
+      environment: BackendEnvironmentProvider,
+      reachability: ReachabilityProvider & TearDownCapable
+    ) {
         self.environment = environment
         self.reachability = reachability
         self.appVersion = appVersion
     }
 
-    func session(delegate: UnauthenticatedSessionDelegate,
-                 authenticationStatusDelegate: ZMAuthenticationStatusDelegate) -> UnauthenticatedSession {
-        let transportSession = UnauthenticatedTransportSession(environment: environment,
-                                                               reachability: reachability,
-                                                               applicationVersion: appVersion)
-        return UnauthenticatedSession(transportSession: transportSession,
-                                      reachability: reachability,
-                                      delegate: delegate,
-                                      authenticationStatusDelegate: authenticationStatusDelegate)
+    func session(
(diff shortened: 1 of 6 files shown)


Source/SessionManager/SessionFactories.swift
```diff
@@ -81,8 +87,13 @@ open class AuthenticatedSessionFactory {
...
 
-    var environment: BackendEnvironmentProvider
-    let reachability: ReachabilityProvider
+    var environment: BackendEnvironmentProvider {
+        didSet {
+            reachability = environment.reachability
+        }
+    }
+
+    var reachability: ReachabilityProvider
```
ghuMustafa, line 96: I'm not sure here but just to understand, do we also need TearDownCapable for `UnauthenticatedSessionFactory` as its above in `AuthenticatedSessionFactory` class?

Source/SessionManager/SessionManager+ServerConnection.swift
```diff
@@ -53,13 +44,5 @@ extension SessionManager: ServerConnection {
...
     }
 
-    /// Add observer of server connection. Returns a token for de-registering the observer.
-    public func addServerConnectionObserver(_ observer: ServerConnectionObserver) -> Any {
-
-        return reachability.addReachabilityObserver(on: .main) { [weak self, weak observer] _ in
-            guard let `self` = self else { return }
-            observer?.serverConnection(didChange: self)
-        }
-    }
```
ghuMustafa, line 47: Vertical Whitespace warning

ghuMustafa (approved):

Source/SessionManager/SessionManager+ServerConnection.swift
johnxnguyen: Fixed

Source/SessionManager/SessionFactories.swift
johnxnguyen: Not sure but I added it just in case.

zenkins:
Hello, @johnxnguyen! Merging PR for command **release major** for PR #**1763**. Will create a **major** release after merged.
Next version is 429.0.0.

PR #1764, opened 2022-07-06
chore: Drop forceVoipSession IC - 59 

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we drop the forceVoipSession because it has been deprecated.

----
##### References

### DIFF
--- a/Source/Synchronization/Strategies/CallingRequestStrategy.swift
+++ b/Source/Synchronization/Strategies/CallingRequestStrategy.swift
@@ -86,7 +86,6 @@ public final class CallingRequestStrategy: AbstractRequestStrategy, ZMSingleRequ
                       clientDiscoverySync.nextRequest(for: apiVersion) ??
                       messageSync.nextRequest(for: apiVersion)
 
-        request?.forceToVoipSession()
         return request
     }
 
--- a/Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
+++ b/Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
@@ -264,7 +264,7 @@ NSUInteger const ZMMissingUpdateEventsTranscoderListPageSize = 500;
         ZMTransportRequest *request = [self.listPaginator nextRequestForAPIVersion:apiVersion];
 
         if (self.isFetchingStreamForAPNS && nil != request) {
-            [request forceToVoipSession];
+            
             [self.notificationsTracker registerStartStreamFetching];
             [request addCompletionHandler:[ZMCompletionHandler handlerOnGroupQueue:self.managedObjectContext block:^(__unused ZMTransportResponse * _Nonnull response) {
                 [self.notificationsTracker registerFinishStreamFetching];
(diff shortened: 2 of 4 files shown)


Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -86,7 +86,6 @@ public final class CallingRequestStrategy: AbstractRequestStrategy, ZMSingleRequ
                       clientDiscoverySync.nextRequest(for: apiVersion) ??
                       messageSync.nextRequest(for: apiVersion)
 
-        request?.forceToVoipSession()
```
johnxnguyen, line 89: @David-Henner or @typfel do you know why we were trying to use a voip url session for this request? My guess is to have possibly more reliable network service? I guess it's fine to use a normal url session.

Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
```diff
@@ -264,7 +264,7 @@ - (ZMTransportRequest *)nextRequestIfAllowedForAPIVersion:(APIVersion)apiVersion
         ZMTransportRequest *request = [self.listPaginator nextRequestForAPIVersion:apiVersion];
 
         if (self.isFetchingStreamForAPNS && nil != request) {
-            [request forceToVoipSession];
+
```
johnxnguyen, line 267: This could possibly run in the background, so it might make sense to use a background session for this, but I'm not sure. @typfel any thoughts on this?

Source/Synchronization/Strategies/CallingRequestStrategy.swift
typfel: The voip url session is not that thing anymore https://developer.apple.com/documentation/foundation/nsurlrequestnetworkservicetype/nsurlnetworkservicetypevoip

In practice is just a regular url session but it still have some side effects since we have multiple url sessions and `forceToVoipSession` would make it pick this particular url session.

typfel: A background session is something very different. You would use a background session if you want to perform long running tasks like downloading a large file which should continue even if the app is suspended or killed.

agisilaos: So how do you think we should move forward? @typfel

agisilaos: @johnxnguyen Talking with @typfel in person, he suggested that we go ahead and remove them for now. Let's not replace it with something else and do some testing and see if everything works.

Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
agisilaos: @johnxnguyen Talking with @typfel in person, he suggested that we go ahead and remove them for now. Let's not replace it with something else and do some testing and see if everything works.

johnxnguyen (approved):

typfel (approved):

PR #1777, opened 2022-08-19
refactor(mls): core crypto setup flow FS-541

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-541" title="FS-541" target="_blank">TaskFS-541</a>  [iOS] Find your MLS conversations on login
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When discovering MLS conversations after login with a new client, we may find MLS conversations that the self user is a part of, but the self client is not. In such case, we need to request to join the MLS conversation.

In order to do this, we need to have access to an instance of `MLSController`. However, the `MLSController` is `nil` during the slow sync. 

### Causes

Currently, the `MLSController` is initialised in the UI, when the application transitions to the `authenticated` state.
But this happens too late in the flow, resulting in the `MLSController` being nil when performing the slow sync.

### Solutions

Move the initialisation of the `MLSController` in two places:
1. In the completion handler of client registration. To cover the case of login on a new client, since we need the client ID to initialise `CoreCrypto`, this is the earliest we can do it.
2. In the init of `ZMUserSession`. To cover the case of an app relaunch where the client has already been registered. 

Since we don't have access to `CoreCrypto` in the Sync Engine, we need to pass a closure from the UI to the user session, that would take care of initialising and returning an instance of `CoreCrypto`

And, since the UI won't be responsible for setting up the `MLSController` we can remove some of the public helper protocols
 
### Testing

- [x] added a dummy `MockCoreCrypto` 
- [x] added a `MockCoreCryptoSetup` with a default instance providing an empty core crypto mock for convenience
- [x] wrote tests for `ZMUserSession+MLS`
- [x] injected the mocked CoreCrypto setup method where necessary

----

----

----
##### References


johnxnguyen (commented):
Looks good! Are you planning to add tests for this PR?

ghuMustafa (approved):

David-Henner:
> Looks good! Are you planning to add tests for this PR?

Yes! Done

johnxnguyen (approved):

PR #1806, opened 2022-12-08
fix: registering multiple push tokens SQCORE-1267

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQCORE-1267" title="SQCORE-1267" target="_blank">BugSQCORE-1267</a>  [iOS] Can't receive push notifications for second account
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Only the first account will receive push notifications, all other accounts will not.

### Causes

This is caused by two bugs:
1. The backend actually rejects registration of the second push token, this is not solved here but in the backend.
2. In some cases, we don't try to register the existing push token for the second account, this is addressed here.

### Solutions

The existing push token management is convoluted and hard to read. The solution is therefore refactor the flow of generating and synchronizing tokens with the backend. It works like this:

1. When a user session is being configured2. 
2. Generate a local push token if needed, either because
  - it doesn't exist yet (fresh install), or
  - it does exist, but it's not the right one (we switched from voip to apns)
3. Register the local token with the backend.4. 
4. Fetch all registered tokens from the backend and deleted any that don't match the local token. This ensures that the backend doesn't send pushes to the device when it shouldn't, e.g for an old voip token.

### Dependencies

Needs releases with:

- https://github.com/wireapp/wire-ios-request-strategy/pull/584

#### Test Coverage

- Test standard token generation  when no local token exists
- Test voip token generation  when no local token exists
- Test migration from voip to standard
- Test syncing with remote (register local token, unregister all others)

----

----

----
##### References


Source/SessionManager/SessionManager+PushToken.swift
```diff
@@ -0,0 +1,151 @@
...
+
+extension SessionManager {
+
+    /* Which pushes do we use?
+     *
+     * iOS 15 and later:
+     *    - Standard APNS pushes -> delivered to Notification Service Extension (NSE).
+     *    - Still register for voIP pushes (BUT DON'T REGISTER WITH BACKEND), so that
+     *      the NSE can wake up the main app to notify calls to CallKit (via PushKit).
+     *    - Why? VoIP pushes are restricted for calling only since iOS 13, out exemption
```
tmspzz, line 30: typo? 

```suggestion
     *    - Why? VoIP pushes are restricted for calling only since iOS 13, our exemption
```

Source/SessionManager/SessionManager+PushToken.swift
```diff
@@ -0,0 +1,107 @@
...
+            return
+        }
+
+        guard localToken.tokenType == requiredPushTokenType else {
+            Logging.push.safePublic("local token is of type \(localToken.tokenType) but should be \(requiredPushTokenType), will generate a new token")
+            generateLocalToken(session: session)
+            return
+        }
+
+        syncLocalTokenWithRemote(session: session)
```
netbe, line 62: don't really get why we pass the session as sessionManager owns all sessions? Or am I missing something?

netbe (commented):
so `SQCORE-1267 [iOS] Can't receive push notifications for second account` is not needed now ?

johnxnguyen:
> so [wireapp/wire-ios-request-strategy#584](https://github.com/wireapp/wire-ios-request-strategy/pull/584) is not needed now ?

@netbe correct, it's no longer needed.

Source/SessionManager/SessionManager+PushToken.swift
johnxnguyen: Yes the session manager owns all the sessions. It also configures sessions when they're created, which is what is happening here:

1. Create a session
2. Configure the session (including pushes)
3. Store the session.

It may be nicer to move the logic to set up pushes to the sessions themselves, but that would take a bit more refactoring.

tmspzz (approved):

KaterinaWire (approved):

johnxnguyen:
release major

zenkins:
Hello, @johnxnguyen! Merging PR for command **release major** for PR #**1806**. Will create a **major** release after merged.
Next version is 434.0.0.

PR #1807, opened 2022-12-12
fix: CallKit integration FS-1142

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1142" title="FS-1142" target="_blank">TaskFS-1142</a>  [iOS] CallKit not working
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR contains substantial refactoring to allow CallKit to work with the NSE. The flow for displaying a CallKit call is as follows:

1. iOS delivers a push to the Notification Service Extension (NSE)
2. The NSE fetches events from the backend, one of which may be a calling events.
3. If the event should represents a new incoming call, or the termination of an existing incoming call, then the NSE wakes up the main app and passes some information about the call event.
4. The main app receives this info and informs CallKit (either start ringing, or stop ringing). 5. 

#### Issues
The existing implementation had a number of issues that fall under two categories:
1. Not satisfying CallKit requirements (i.e if the NSE wakes up the main app, it must report the call to CallKit).
2. Not satisfying AVS requirements such that AVS can actually connect calls via CallKit. 

#### Solution
The details for these issues are numerous and complicated. Instead of describing the problems in detail,  I'll do my best to be thorough in describing the solution I have arrived at:

In order to satisfy CallKit, we must report calls to CallKit when being woken by the NSE. Through some experimentation, we have learned that we can't just report calls willy nilly: what we report needs to make sense, anything else may result in being terminated by CallKit. 

What we found is that we must:
- only report a _new_ incoming call once,
- only report a call ended if you have previously reported the call incoming.

The NSE is responsible for deciding whether or not to wake the main app, but CallKit integration exists only in the main app. Therefore we need to share some information between the sync engine and the notification engine. This is done via the `VoIPPushHelper`, in particular via the `knownCalls` property. It is crucial that the known calls is an accurate reflection of what CallKit sees, i.e if we report an incoming call, it becomes known. If an existing call is terminated in any way, it becomes unknown.

Having satisfied CallKit, the next challenge is to satisfy AVS. Being able to see CallKit calls come and go -- without the app being killed -- is great, but it's not much use at all if the call can't be answered. The first major problem is that since the NSE can only wake up the main app to inform CallKit to start or stop ringing, it is forced to ignore all other call events that may be received. In conference calls, there are a large number of call events that don't correspond to these start and stop events, a simple example is CONFCHECK messages.

AVS requires that **all call messages be processed** so that it can get in a state to provide successful telephony. Before the extension and VoIP restrictions, we had no problem to forward all call messages to AVS because it all happened in the main app. With the extension however it is not possible to process all events because AVS doesn't exist in the extension (it would need to be the exact same one in the main app). Forwarding the events from the extension to the main app when reporting to CallKit doesn't work either, because then we'd need to wake up the main app for events that don't correspond to a CallKit event (leading to app termination). The solution we've arrived at is to not forward any events to AVS at all... until the user actually wants to answer the call. When the user tries to answer, the main app is woken in the background. We then perform a sync, which will pass all received events to AVS, and only when we're done will we attempt to join the call.

One final problem is that there are cases where the app is woken up from a terminated state and can not set up the PushKit delegate (SessionManager) in time to process the push payload, resulting in app termination. We must therefore register the delegate as early as possible in the app life cycle (directly after launch). In order to do this, we need to take this PushKit responsibilities away from the SessionManager.

#### Changes

The main changes are this:
- Create `VoIPPushManager`, a class responsible for talking to PushKit and receiving the push payloads from the extension. It gets created in the `AppDelegate` (UI project) just after launch. 
- Factor out the `CallKitManager` from the `SessionManager`, move it to the `VoIPPushManager` so that it can report calls to CallKit without waiting for the `SessionManager` to load.
- Refactor `CallKitManager` and pieces that it uses to be able report calls to CallKit that originate from the NSE. 
  - We've added some "preemptive" reporting methods, used only by the `VoIPPushManager`.
  - Refactored `CallHandle`s to not be coupled with `ZMConversation` since they now need to work before we may even have a `CoreDataStack` available
  - Refactoring performing CallKit actions to ensure that the app syncs before we try to interact with calls (via AVS)

### Dependencies

Needs releases with:

- https://github.com/wireapp/wire-ios-request-strategy/pull/582

#### Test Coverage

TODO

----

----

----
##### References


Source/Calling/CallKitManager.swift
```diff
@@ -20,42 +20,21 @@ import Foundation
...
-
-    init?(customIdentifier value: String) {
-        let identifiers = value.split(separator: identifierSeparator).compactMap({ UUID(uuidString: String($0)) })
-
-        guard identifiers.count == 2 else { return nil }
-
-        self.accountId = identifiers[0]
-        self.conversationId = identifiers[1]
-    }
-}
```
johnxnguyen, line 52: These were moved to new files and received some changes.

Source/Calling/CallKitManager.swift
```diff
@@ -64,41 +43,94 @@ protocol CallKitManagerDelegate: AnyObject {
...
-         callController: CXCallController,
-         delegate: CallKitManagerDelegate,
-         mediaManager: MediaManagerType?) {
+    // MARK: - Properties
 
+    public var isEnabled: Bool {
+        didSet {
+            VoIPPushHelper.isCallKitAvailable = isEnabled
+        }
+    }
```
johnxnguyen, line 68: One major change:
- **previously** the `CallKitManager` was created and destroyed by the `SessionManager` (its former owner) when the user enabled / disabled CallKit in the options.
- **now** that the `VoIPPushManager` is the owner of the `CallKitManager`, we allow the `SessionManager` to still enabled / disable CallKit via this property.

Source/Calling/CallKitManager.swift
```diff
@@ -134,205 +178,329 @@ public class CallKitManager: NSObject {
...
-        guard let interaction = userActivity.interaction
-        else { return false }
+        Self.logger.trace("continue user activity")
+        guard let interaction = userActivity.interaction else { return false }
 
         let intent = interaction.intent
         var contacts: [INPerson]?
         var video = false
 
+        // TODO: handle INStartVideoCallIntent for when CallKit video is toggled.
```
johnxnguyen, line 235: We have a ticket for this, it'll be addressed separately.

Source/Calling/CallKitManager.swift
```diff
@@ -134,205 +178,329 @@ public class CallKitManager: NSObject {
...
         else {
+            Self.logger.warning("fail: request start call: context or handle missing")
             self.log("Ignore request to start call since remoteIdentifier or handle is nil")
             return
         }
 
-        let callUUID = UUID()
-        calls[callUUID] = CallKitCall(conversation: conversation)
+        // TODO: do we need to check there doesn't already exist a call?
+        let call = callRegister.registerNewCall(with: handle)
```
johnxnguyen, line 320: I need to check if this is still a valid concern.

Source/Calling/CallKitManager.swift
```diff
@@ -134,205 +178,329 @@ public class CallKitManager: NSObject {
...
         }
     }
 
+    // MARK: - Reporting calls
+
+    func reportIncomingCallPreemptively(
+        handle: CallHandle,
+        callerName: String,
+        hasVideo: Bool
+    ) {
```
johnxnguyen, line 397: This is the method to call when reporting an incoming call detected from the NSE (as opposed to when the app is running in the foreground and processing events).

Source/Calling/CallKitManager.swift
```diff
@@ -134,205 +178,329 @@ public class CallKitManager: NSObject {
...
+                self?.log("Cannot preemptively report incoming call: \(error)")
+                self?.callRegister.unregisterCall(call)
+            }
+        }
+    }
+
+    func reportCallEndedPreemptively(
+        handle: CallHandle,
+        reason: CXCallEndedReason
+    ) {
```
johnxnguyen, line 431: This is the method to call when reporting a terminated call (detected via the NSE)

Source/Calling/CallKitManager.swift
```diff
@@ -134,205 +178,329 @@ public class CallKitManager: NSObject {
...
 
-    func reportIncomingCall(from user: ZMUser, in conversation: ZMConversation, video: Bool) throws {
-        guard !callExists(for: conversation) else {
-            log("Cannot report incoming call: call already exists, probably b/c it was reported earlier for a push notification")
-            throw ReportIncomingCallError.callAlreadyExists
+    func reportIncomingCall(
+        from user: ZMUser,
+        in conversation: ZMConversation,
+        hasVideo: Bool
+    ) {
```
johnxnguyen, line 459: This is an existing method to report an incoming call, it should only be used when the app is already in the foreground and processing events.

Source/Calling/CallKitManager.swift
```diff
@@ -343,199 +511,424 @@ extension CallKitManager {
...
 
-    func reportCallEnded(in conversation: ZMConversation, atTime timestamp: Date?, reason: CXCallEndedReason) throws {
-        let associatedCallUUIDs = calls
-            .filter { $0.value.conversation == conversation }
-            .keys
+    func reportCallEnded(
+        in conversation: ZMConversation,
+        atTime timestamp: Date?,
+        reason: CXCallEndedReason
+    ) {
```
johnxnguyen, line 519: Likewise, this an existing method and should only be used when the app is in the foreground.

Source/Calling/CallKitManager.swift
```diff
@@ -343,199 +511,424 @@ extension CallKitManager {
...
             return
         }
 
-        call.observer.onAnswered = {
-            provider.reportOutgoingCall(with: action.callUUID, startedConnectingAt: Date())
+        guard let delegate = delegate else {
+            Self.logger.warning("fail: perform start call action: delegate doesn't exist")
+            log("fail CXStartCallAction because can't fetch conversation")
+            action.fail()
+            return
```
johnxnguyen, line 573: The delegate is needed in order to get the `ZMConversation`. Previously this was store as part of the `CallKitCall`, but it was removed since calls can now exist before a CoreData stack exists. Hence, we ask on demand when needed.

Source/Calling/CallKitManager.swift
```diff
@@ -343,199 +511,424 @@ extension CallKitManager {
...
+        delegate.lookupConversation(by: call.handle) { [weak self] result in
+            guard let `self` = self else {
+                action.fail()
+                return
+            }
 
-        mediaManager?.setupAudioDevice()
+            switch result {
+            case .success(let conversation):
+                call.observer.startObservingChanges(in: conversation)
```
johnxnguyen, line 584: Before we started observing as soon as we created the `CallKitCall`. Now we do it on demand when needed.

Source/Calling/CallKitManager.swift
```diff
@@ -343,199 +511,424 @@ extension CallKitManager {
...
-        call.observer.onFailedToJoin = {
+        guard let delegate = delegate else {
+            Self.logger.warning("fail: perform answer call action: delegate doesn't exist")
+            log("fail CXAnswerCallAction because can't fetch conversation")
             action.fail()
+            return
         }
 
-        mediaManager?.setupAudioDevice()
+        delegate.lookupConversationAndSync(by: call.handle) { [weak self] result in
```
johnxnguyen, line 635: It's important to perform a sync first so that AVS can process all pending call events, so that we can eventually ask it to join the call.

Source/Calling/CallKitManager.swift
```diff
@@ -343,199 +511,424 @@ extension CallKitManager {
...
+        switch (requirePushTokenType, application.applicationState) {
+        case (.standard, .background):
+            // Processing call states in the background wiLl interfere with
+            // CallKit calls triggered from the NSE.
+            return false
+
+        default:
+            return true
+        }
+    }
```
johnxnguyen, line 820: The side effect of processing call events (via AVS) is that AVS will invoke callbacks to the `WireCallCenter`, which will in turn generate and broadcast `CallState`s. The `CallKitManager` listens for callstate changes to inform report calls. We still want to do this, but only in some cases, so we don't conflict with reporting done by the NSE via `VoIPPushManager`.

If we have standard push tokens (and therefore the NSE is active) and we're in the background, we don't handle call states because we're reporting calls via another path. Otherwise, we open this path and let the main app report calls as usual.

KaterinaWire:
release major

zenkins:
Hello, @KaterinaWire! Merging PR for command **release major** for PR #**1807**. Will create a **major** release after merged.
Next version is 435.0.0.

PR #1814, opened 2022-12-21
feat(api-v3): add api migration mechanism + migrate access token FS-1237

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1237" title="FS-1237" target="_blank">Sub-taskFS-1237</a>  [iOS] Implement API version migrations
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR introduces the `APIMigrationManager` that enables us to run migrations when we detect that the api version has changed. 
It also adds a migration for v3, `AccessTokenMigration`, which associates the session's client id to the access token.

### Dependencies (Optional)

https://github.com/wireapp/wire-ios-transport/pull/261

#### Test Coverage (Optional)

- [ ] I have added automated test to this contribution

### Note

I've added 3 TODOs in the code. I plan on removing them before merging and I'll either implement changes in this PR or create tickets for them if needed.

----

----

----
##### References

### DIFF
--- a/Source/SessionManager/APIMigration/APIMigrationManager.swift
+++ b/Source/SessionManager/APIMigration/APIMigrationManager.swift
@@ -25,6 +25,7 @@ protocol APIMigration {
 
 class APIMigrationManager {
     let migrations: [APIMigration]
+    var previousAPIVersion: APIVersion?
 
     private let logger = Logging.apiMigration
 
@@ -64,18 +65,11 @@ class APIMigrationManager {
                 return
             }
 
-            // TODO: David
-            // What if there is no last used API version, but we still need the migrations to be done?
-            // e.g: client with active session installs latest version of the app,
-            // has no last used version saved.
-            // previous version was v2, and current version is v3.
-            // we need the client to migrate the access token,
-            // but, since no last used version is recorded, we won't perform the migration.
-            //
-            // solution: maybe we can persist the last used version before resolving the api version?
-            // -> note that we may need to do that for all sessions
-            guard let lastUsedVersion = lastUsedAPIVersion(for: clientID) else {
+            defer {
                 persistLastUsedAPIVersion(for: clientID, apiVersion: apiVersion)
+            }
+
+            guard let lastUsedVersion = lastUsedAPIVersion(for: clientID) ?? previousAPIVersion else {
                 return
             }
 
@@ -85,7 +79,6 @@ class APIMigrationManager {
                 from: lastUsedVersion,
                 to: apiVersion
             )
-            persistLastUsedAPIVersion(for: clientID, apiVersion: apiVersion)
         }
     }
 
--- a/Source/SessionManager/APIMigration/APIMigrationManager.swift
+++ b/Source/SessionManager/APIMigration/APIMigrationManager.swift
@@ -1,0 +1,135 @@
+//
+// Wire
+// Copyright (C) 2022 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+protocol APIMigration {
+    func perform(with session: ZMUserSession, clientID: String) async throws
+    var version: APIVersion { get }
+}
+
+class APIMigrationManager {
+    let migrations: [APIMigration]
+
+    private let logger = Logging.apiMigration
+
+    init(migrations: [APIMigration]) {
+        self.migrations = migrations
+    }
+
+    func migrate(
+        session: ZMUserSession,
+        clientID: String,
+        from lastVersion: APIVersion,
+        to currentVersion: APIVersion
+    ) async {
+        guard lastVersion.rawValue < currentVersion.rawValue else {
+            return
+        }
+
+        logger.info("starting API migrations from api v\(lastVersion.rawValue) to v\(currentVersion.rawValue) for session with clientID \(String(describing: clientID))")
+
+        for migration in migrations.filter({
+            (lastVersion.rawValue+1..<currentVersion.rawValue+1).contains($0.version.rawValue)
+        }) {
+            do {
+                logger.info("starting migration (\(String(describing: migration))) for api v\(migration.version.rawValue)")
+                try await migration.perform(with: session, clientID: clientID)
+            } catch {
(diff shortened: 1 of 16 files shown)


Source/SessionManager/APIMigration/APIMigrationManager.swift
```diff
@@ -0,0 +1,141 @@
...
+            // TODO: David
+            // What if there is no last used API version, but we still need the migrations to be done?
+            // e.g: client with active session installs latest version of the app,
+            // has no last used version saved.
+            // previous version was v2, and current version is v3.
+            // we need the client to migrate the access token,
+            // but, since no last used version is recorded, we won't perform the migration.
+            //
+            // solution: maybe we can persist the last used version before resolving the api version?
+            // -> note that we may need to do that for all sessions
```
David-Henner, line 76: As this corner case would only affect clients updating from an app version without the migration manager to an app with it, it would be acceptable to perform all migrations from v2 to the current api version

Source/SessionManager/APIMigration/AccessTokenMigration.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    func perform(with tokenRenewer: AccessTokenRenewing, clientID: String) async throws {
+        logger.info("performing access token migration for clientID \(clientID)")
+
+        tokenRenewer.setAccessTokenRenewalObserver(self)
+
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Swift.Error>) in
+            self.continuation = continuation
+            tokenRenewer.renewAccessToken(with: clientID)
```
johnxnguyen, line 57: I'm not sure if this could be a retain cycle, maybe safest to have a weak self.

Source/SessionManager/APIMigration/AccessTokenMigration.swift
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    // TODO: David
+    // We may want to tear down the access token renewal observer when it succeeds / fails
+    // to avoid getting callbacks when the token is renewed from elsewhere.
+    // Note that since we tear down the continuation, the only consequence to that is that we will log.
```
johnxnguyen, line 64: This would be done on the `ZMUserSession` side though, right?

Source/SessionManager/SessionManager+APIVersionResolver.swift
```diff
@@ -46,6 +46,27 @@ extension SessionManager: APIVersionResolverDelegate {
...
     }
 
+    func apiVersionResolverDidResolve(apiVersion: APIVersion) {
+
+        // TODO: David
+        // calling `sessionManagerWillMigrateAccount` will transition the app in the
+        // migrating state and show a loading screen. it will happen each time the version
+        // is resolved i.e each time the app goes in the foreground. This may be too much
+        // and could be avoided if we delayed this until after we've checked if migrations
+        // are needed. However, there'd be a risk that requests are fired in the meantime.
```
johnxnguyen, line 56: Why don't we first ask the api migration manager if a migration is need, and only then transition to the migration phase?

Source/SessionManager/APIMigration/AccessTokenMigration.swift
```diff
@@ -0,0 +1,81 @@
...
+    let version: APIVersion
+
+    private var continuation: CheckedContinuation<Void, Swift.Error>?
+    private let logger = Logging.apiMigration
+
+    enum Error: Swift.Error {
+        case failedToRenewAccessToken
+    }
+
+    init(version: APIVersion) {
```
johnxnguyen, line 42: Is it necessary to inject a version? I would expect this migration to only apply to a single (hardcoded) version.

Source/UserSession/ZMUserSession+AccessToken.swift
```diff
@@ -0,0 +1,45 @@
...
+            let selfUser = ZMUser.selfUser(in: self.managedObjectContext)
+            let error = NSError.userSessionErrorWith(.accessTokenExpired, userInfo: selfUser.loginCredentials.dictionaryRepresentation)
+            self.notifyAuthenticationInvalidated(error)
+        }
+
+        accessTokenRenewalObserver?.accessTokenRenewalDidFail()
+    }
+
+    func transportSessionAccessTokenDidSucceed() {
+        accessTokenRenewalObserver?.accessTokenRenewalDidSucceed()
```
johnxnguyen, line 43: As you mention, it would be good to clear the observer in both success and failures.

Tests/Source/SessionManager/APIMigrationManagerTests.swift
```diff
@@ -0,0 +1,188 @@
...
+        sut.resetLastUsedAPIVersion(for: clientID)
+        XCTAssertNil(sut.lastUsedAPIVersion(for: clientID))
+
+        // When
+        await sut.migrateIfNeeded(sessions: [userSession], to: APIVersion.v3)
+
+        // Then
+        XCTAssertEqual(sut.lastUsedAPIVersion(for: clientID), APIVersion.v3)
+
+        userSession.tearDown()
```
johnxnguyen, line 58: Perhaps we should also delete the user defaults suite that was created for this client.

Source/SessionManager/APIMigration/APIMigrationManager.swift
```diff
@@ -25,28 +25,49 @@ protocol APIMigration {
...
 
+    func isMigration(to apiVersion: APIVersion, neededForSessions sessions: [ZMUserSession]) -> Bool {
+
+        for session in sessions {
+            if isMigration(to: apiVersion, neededForSession: session) {
+                return true
+            }
+        }
+
+        return false
```
johnxnguyen: I believe this is equivalent to:

```swift
return sessions.contains { 
    isMigration(to: apiVersion, neededForSession: $0) 
}
```

Source/SessionManager/APIMigration/APIMigrationManager.swift
```diff
@@ -61,39 +82,37 @@ class APIMigrationManager {
...
+            persistLastUsedAPIVersion(for: $0, apiVersion: apiVersion)
         }
     }
 
-    func lastUsedAPIVersion(for clientID: String) -> APIVersion? {
-        return userDefaults(for: clientID).lastUsedAPIVersion
+    // MARK: - Helpers
+
+    func lastUsedAPIVersion(for clientID: String) -> APIVersion {
+        return userDefaults(for: clientID).lastUsedAPIVersion ?? previousAPIVersion ?? .v2
```
johnxnguyen, line 103: The fallback to `.v2` is because that's would likely be the latest version at the time of release?

Source/SessionManager/APIMigration/APIMigrationManager.swift
David-Henner: good catch

Source/SessionManager/APIMigration/APIMigrationManager.swift
David-Henner: More-so because that would be the minimum version needed to trigger migrations (with the access token migration being the only one for now and starting at v3). This means that the fallback to having no previously saved global api version would be running all migrations (note that this is the same fallback that android has, since they do not persist the resolved api version anywhere. we on iOS would use it only as a second fallback)

johnxnguyen (approved):

